PR #281, opened 2025-12-25
Update hooks API to comply with Keystone specification

borisno2:
This commit updates the hooks system to be fully compliant with Keystone's
hooks specification. All hook arguments now match the documented API.

Changes:
- Added listKey parameter to all list-level and field-level hooks
- Added inputData parameter to track original input before transformations
- Renamed validateInput to validate (kept validateInput as deprecated alias)
- Updated field-level hooks to use fieldKey parameter consistently
- Added resolvedData parameter to beforeOperation and afterOperation hooks
- Updated all hook execution to pass correct arguments

The hooks API now matches Keystone's specification exactly, making it
easier for developers familiar with Keystone to use the stack.

Backward compatibility is maintained where possible through deprecated
aliases and parameter additions (not removals).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -165,7 +165,7 @@ The hooks system provides data transformation and side effects during database o
 All hooks receive these common arguments:
 
 - `listKey` - The name of the list being operated on
-- `operation` - The operation type ('create', 'update', 'delete', or 'query')
+- `operation` - The operation type ('create', 'update', or 'delete'). For `resolveOutput` hooks, this is 'query'
 - `context` - The AccessContext object
 
 **List-level hooks** additionally receive:
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -68,6 +68,76 @@ async function executeFieldResolveInputHooks(
 }
 
 /**
+ * Execute field-level validate hooks
+ * Allows fields to perform custom validation after resolveInput but before database write
+ */
+async function executeFieldValidateHooks(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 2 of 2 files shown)


borisno2:
## Addressing Review Comments

I've addressed all 5 review comments from the Copilot review:

### Comment 1: beforeOperation type mismatch (types.ts:154)
**Fixed.** The `executeFieldBeforeOperationHooks` function now has separate branches for create, update, and delete operations. The create branch no longer passes `item` at all, matching Keystone's API exactly and our type definition.

### Comment 2: Field-level validate hook not implemented (types.ts:98-128)
**Fixed.** Added `executeFieldValidateHooks` function that:
- Executes field-level validate hooks for each field
- Supports create, update, and delete operations
- Collects validation errors via `addValidationError`
- Throws `ValidationError` if any errors are found

The function is now called in create, update, and delete operations after list-level validate and before field validation rules.

### Comment 3: Documentation error about operation types (CLAUDE.md:168)
**Fixed.** Updated to clarify that most hooks use 'create', 'update', or 'delete', and 'query' is only for `resolveOutput` hooks.

### Comment 4: Field-level validate documentation (CLAUDE.md:193)
**Now valid.** With the implementation of field-level validate hooks, this documentation is now accurate.

### Comment 5: Changeset claim about field-level validate (.changeset/purple-hooks-comply.md:22)
**Now valid.** With the implementation complete, the changeset claim is now accurate.

### Additional Improvement
I also noticed that delete operations weren't calling list-level `validate` hooks. Per Keystone's docs, delete operations support validate hooks, so I added the call to `executeValidate` for delete operations as well.

PR #397, opened 2026-05-29
Move relationship Prisma schema generation into the relationship field builder

### LINKED ISSUE #395: Move relationship Prisma schema generation into the relationship field builder
## What to build

`packages/cli/src/generator/prisma.ts` contains ~60 lines of custom handling for relationship fields: parsing the `ref` string, determining one-to-one vs. one-to-many topology, deciding which side owns the foreign key, applying join table naming strategies (`keystone` vs. default Prisma), and handling per-field `db.relationName` overrides.

This logic lives in the generator, not in the relationship field builder — which means it is the one field type that breaks the stack's "delegate to field methods, no generator switches" principle. Every other field type generates its Prisma representation via `field.getPrismaType()` and the generator is a neutral coordinator. Relationship fields are a special-cased exception.

The fix is to add a `getPrismaRelation(fieldName, allFields, listKey, config)` method (or equivalent — name is up to the implementer) to the relationship field builder in `packages/core/src/fields/index.ts`. The generator calls this method and uses the result, removing the custom relationship handling from `prisma.ts`.

**Edge cases the method must cover (currently handled inline in the generator):**

- Bidirectional refs (`ref: 'Post.author'`) vs. list-only refs (`ref: 'Post'`)
- Synthetic back-relation field name for list-only refs (`from_<SourceList>_<field>`)
- One-to-one vs. one-to-many (which side owns the FK)
- Many-to-many (no FK, implicit join table)
- `db.foreignKey: true/false`
- `db.relationName` override (per-field)
- Global `db.joinTableNaming: 'keystone'` producing `_<List>_<field>` join table names
- `db.extendPrismaSchema` callback on the relationship field

## Acceptance criteria

- [ ] The relationship field builder exposes a method that returns its complete Prisma schema contribution (field line, optional FK line, optional back-relation line)
- [ ] `packages/cli/src/generator/prisma.ts` removes its per-relationship branching and delegates to the new method
- [ ] All edge cases listed above are handled by the method, not by the generator
- [ ] Generated Prisma schemas for all examples (`examples/blog`, `examples/auth-demo`, `examples/mcp-demo`) are byte-for-byte identical before and after the refactor (run `pnpm generate` in each and diff)
- [ ] All existing tests pass

## Blocked by

None — can start immediately.

borisno2:
## Summary

`packages/cli/src/generator/prisma.ts` previously contained ~300 lines of relationship-specific logic — `ref` parsing, one-to-one vs. one-to-many FK ownership, many-to-many detection, join-table naming, per-field `db.relationName`, and `extendPrismaSchema` handling. Relationships were the one field type that broke the stack's "delegate to field methods, no generator switches" principle.

This PR moves all of that logic into the relationship field builder via a new `getPrismaRelation()` method, so the generator becomes a neutral coordinator that just places the lines each field returns.

- Add `getPrismaRelation(fieldName, allFields, listKey, config)` to the relationship field builder (`packages/core/src/fields/index.ts`), returning a `PrismaRelationResult` with the model line(s), optional FK index, and optional synthetic back-relation.
- Add the `PrismaRelationResult` type and `getPrismaRelation` method to `RelationshipField` (`packages/core/src/config/types.ts`), exported from core.
- Refactor `generatePrismaSchema` to pre-compute each relationship's contribution and delegate placement — removing all per-relationship branching (~471 net lines deleted).

### Edge cases covered by the field method (previously inline in the generator)
- Bidirectional refs (`ref: 'Post.author'`) vs. list-only refs (`ref: 'Post'`)
- Synthetic back-relation field (`from_<List>_<field>`) for list-only refs
- One-to-one vs. one-to-many FK ownership (incl. `db.foreignKey: true/false` validation)
- Many-to-many (implicit join table, no FK)
- `db.relationName` per-field override (with mismatch validation)
- Global `db.joinTableNaming: 'keystone'` deterministic naming
- Field-level `db.extendPrismaSchema` callback

## Test plan
- [x] All existing tests pass (`stack-core`: 570, `stack-cli`: 192, incl. the 58 prisma generator tests + snapshots)
- [x] Generated Prisma schemas for `examples/blog`, `examples/auth-demo`, `examples/mcp-demo` regenerated and verified **byte-for-byte identical** to before the refactor
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` clean
- [x] Changeset added

Closes #395

https://claude.ai/code/session_01QDoQeZU4cgRVULSnMeEkCq

---
_Generated by [Claude Code](https://claude.ai/code/session_01QDoQeZU4cgRVULSnMeEkCq)_

### DIFF
--- a/.changeset/quiet-otters-delegate.md
+++ b/.changeset/quiet-otters-delegate.md
@@ -1,0 +1,8 @@
+---
+'@opensaas/stack-core': patch
+'@opensaas/stack-cli': patch
+---
+
+Move relationship Prisma schema generation into the relationship field builder
+
+The relationship field now exposes a `getPrismaRelation()` method that returns its complete Prisma schema contribution (FK line, relation line, synthetic back-relation). The Prisma generator delegates to this method instead of special-casing relationships, keeping it a neutral coordinator. Generated schemas are unchanged.
--- a/packages/cli/src/generator/prisma.ts
+++ b/packages/cli/src/generator/prisma.ts
@@ -98,6 +98,8 @@ export function generatePrismaSchema(config: OpenSaasConfig): string {
 
   for (const [listName, listConfig] of Object.entries(config.lists)) {
     for (const [fieldName, fieldConfig] of Object.entries(listConfig.fields)) {
+      // Skip non-relationship fields and virtual relationships (which contribute
+      // no database columns and therefore no Prisma schema lines).
       if (fieldConfig.type !== 'relationship' || fieldConfig.virtual) continue
 
       const relField = fieldConfig as RelationshipField
@@ -161,7 +163,7 @@ export function generatePrismaSchema(config: OpenSaasConfig): string {
     }
 
     // Add relationship fields (lines + foreign key indexes) from the precomputed results
-    const foreignKeyIndexes: PrismaRelationResult['foreignKeyIndex'][] = []
+    const foreignKeyIndexes: NonNullable<PrismaRelationResult['foreignKeyIndex']>[] = []
     for (const fieldName of relationshipFieldNames) {
       const result = relationResults.get(`${listName}.${fieldName}`)!
       lines.push(...result.modelLines)
@@ -184,7 +186,6 @@ export function generatePrismaSchema(config: OpenSaasConfig): string {
 
     // Add indexes for foreign key fields
     for (const index of foreignKeyIndexes) {
-      if (!index) continue
       if (index.indexType === 'unique') {
         lines.push(`  @@unique([${index.foreignKeyField}])`)
       } else if (index.indexType === true) {
--- a/packages/cli/src/generator/prisma.ts
+++ b/packages/cli/src/generator/prisma.ts
@@ -1,4 +1,9 @@
-import type { OpenSaasConfig, FieldConfig, RelationshipField } from '@opensaas/stack-core'
+import type {
+  OpenSaasConfig,
+  FieldConfig,
+  RelationshipField,
+  PrismaRelationResult,
+} from '@opensaas/stack-core'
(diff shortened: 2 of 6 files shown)


borisno2:
## Code review summary

A full review of the refactor was completed. **No blockers, no should-fix items** — the change faithfully relocates relationship topology logic into `getPrismaRelation()` with byte-for-byte identical output, verified against the deleted inline generator code path-by-path:

- **One-to-one FK ownership** — identical precedence (explicit `db.foreignKey` → dual-true error → alphabetical `localeCompare` tiebreak → self-referential field-name tiebreak).
- **Keystone bidirectional naming** — old code deduped to one deterministic name; new per-side computation is provably symmetric (`min(sourceKey, targetKey)`), yielding the same name on both sides.
- **Per-field `relationName`** — one-sided resolution via the target lookup and the mismatch-error trigger side are preserved (iteration order unchanged).
- **Synthetic back-relation ordering**, **model line ordering** (regular → relationship → synthetic → timestamps → indexes), and **exact string formatting** (padding, double-space before `@relation`, `@map` defaults) all confirmed identical.
- No dead code or unused imports left in `prisma.ts`; no new `any`/unsafe casts.

### Nits addressed in a63444f
1. Added a comment documenting the intentional skip of virtual relationship fields (the old code could emit a dangling back-relation for that theoretical case; the new behavior is more consistent).
2. Tightened the FK-index array typing to `NonNullable<...>[]` and removed the now-redundant `undefined` guard.

Nit left as-is: the `allFields` parameter on `getPrismaRelation` is currently unused by the implementation, but it's kept (and documented) to match the method signature proposed in #395 and to give third-party relationship-like fields access to sibling field config.

---

PR #403, opened 2026-05-29
refactor(core): name and bound the access-control two-phase read

### LINKED ISSUE #394: Name and extract the access control dual-pass (design checkpoint)
## What to build

`packages/core/src/access/engine.ts` performs access control filtering in two distinct phases, but neither phase is named or bounded as a concept:

**Phase 1 — pre-query:** `buildIncludeWithAccessControl` builds the Prisma `include`/`where` clause that gates which rows and relationships come back from the database. This happens before the query.

**Phase 2 — post-query:** `filterReadableFields` walks the returned objects, removes fields the session cannot read, and invokes `resolveOutput` hooks. This happens after the query on the result set.

The split is architecturally necessary: virtual fields are computed post-query and cannot be expressed in SQL, so they must be handled in Phase 2. However, because neither phase is named, field-level access evaluation logic is duplicated across both, and developers reading `engine.ts` must reconstruct the two-phase invariant from scratch.

The goal is to extract a shared `evaluateFieldAccess(fieldConfig, session, context)` function and name the two phases as distinct, clearly bounded modules so the invariant is explicit in the code structure.

## This is a design checkpoint

Before implementation, this issue needs human decisions on:

1. **Module names** — What should the two phases be called? (Current working names: `QueryBuilder` for pre-query, `ResultFilter` for post-query — but these are open for debate.)
2. **Shared evaluator interface** — What exactly should `evaluateFieldAccess` accept and return? Should it live in `engine.ts`, in a new `access/field-evaluator.ts`, or elsewhere?
3. **Scope** — Is the goal structural naming only (rename/reorganise, no behaviour change), or should this also consolidate the duplicated access logic?

Once those decisions are made, update this issue with the agreed names and interface, then change the label to `ready-for-agent` so an AFK agent can implement it.

## Acceptance criteria (to be confirmed after design checkpoint)

- [ ] The two access control phases are named as distinct modules with clear, agreed names
- [ ] Field-level access evaluation lives in a single shared function, not duplicated across both phases
- [ ] All existing tests pass
- [ ] The two-phase invariant is documented (inline or in a CLAUDE.md update)
- [ ] No behaviour change — this is structural only

## Blocked by

None — can start immediately (but requires human design input before AFK implementation).

borisno2:
## Summary

Splits `packages/core/src/access/engine.ts` into clearly named modules so the two-phase read invariant described in issue #394 is explicit in the code structure. This is **structural only — no behaviour change and no public API change**.

- **`access-filter.ts`** — *Access Filter* (phase 1, pre-query): row/relation scoping. Retains `buildIncludeWithAccessControl`.
- **`field-visibility.ts`** — *Field Visibility* (phase 2, post-query): field stripping + `resolveOutput` + virtual-field computation. Retains `filterReadableFields`.
- **`field-access.ts`** — the canonical field-level evaluator `checkFieldAccess` (unchanged signature) plus `filterWritableFields`, shared by the read and write paths.
- **`engine.ts`** — operation-level helpers (`checkAccess`, `mergeFilters`, `isBoolean`, `isPrismaFilter`) and the shared `getRelatedListConfig` parser.

The repeated "check read access → skip if denied → `resolveOutput`" sequence in the post-query phase is extracted into a single unexported helper (`resolveReadableFieldValue`), removing the duplication between the regular-field and virtual-field branches.

Per the design checkpoint, **no new `evaluateFieldAccess` function was introduced** — `checkFieldAccess` remains the single field-access evaluator with its existing signature (it needs `operation`, `item`, and `inputData`). Each module carries an inline pointer to the two-phase model documented in `CONTEXT.md` and `docs/adr/0001-access-control-is-a-two-phase-read.md`.

The access barrel (`access/index.ts`) re-exports every previously-exported name, so no importer breaks.

## Test plan

- [x] All 570 core tests pass (no assertion changes; only import paths repointed at the barrel)
- [x] `pnpm build` (core) passes
- [x] `pnpm lint` passes (0 errors)
- [x] `pnpm format` / `pnpm manypkg fix` clean

## Acceptance criteria

- [x] Pre-query and post-query phases live in distinct, clearly named modules (Access Filter / Field Visibility) with the shared field-access evaluator in its own module
- [x] The "check read access → skip → resolveOutput" sequence exists in exactly one helper
- [x] No new field-access evaluator added; `checkFieldAccess` keeps its signature
- [x] Every name currently exported from the access barrel is still exported
- [x] All existing tests pass with no modification to assertions
- [x] The two-phase invariant is documented (inline pointers to the ADR/glossary)

Closes #394

https://claude.ai/code/session_01TFshzqx6rpfLDWyygdaZ2o

---
_Generated by [Claude Code](https://claude.ai/code/session_01TFshzqx6rpfLDWyygdaZ2o)_


borisno2:
## Review summary

Ran a correctness-focused review of this refactor (subagent traced the new `field-visibility.ts` branch-by-branch against the old `engine.ts`). **Behaviour is preserved — no blockers, no should-fixes.**

- **Read-access call count unchanged:** old code checked `checkFieldAccess` once per field at the top of the loop; new code checks exactly once per field — either in the relationship branch or inside `resolveReadableFieldValue`. Identical deny→skip semantics.
- **Relationship branch** (array / single / related-config-not-found / `config` undefined / null relationship value) — all paths produce identical results; the new `continue` prevents double-processing.
- **Virtual fields without a `resolveOutput` hook** (the subtlest case) — still call `checkFieldAccess` to preserve access-fn side effects, then skip without adding the field, matching the old behaviour.
- **`item` plumbing** — access checks use the raw fetched row; scalar hooks get `item`, virtual hooks get `filtered`; `_resolveOutputCounter.depth` increment/decrement preserved.
- **Barrel** re-exports all 9 runtime + 11 type names from `main`; `context/index.ts` imports through the barrel, so the move is transparent.
- **No circular imports:** `engine.ts` and `field-access.ts` are leaves; `access-filter.ts → engine.ts`; `field-visibility.ts → {engine.ts, field-access.ts}`.

One NIT: `matchesFilter` in `field-access.ts` is dead code — but it was already dead on `main` and only moved files, so leaving it is consistent with the "structural only" mandate.

All 570 core tests pass; `pnpm build`/`lint`/`format` clean.

---

Issue #450, opened 2026-06-01
Deploy 1: production migration workflow (migrate scripts + pooled/direct prisma.config; SQLite default unchanged)


borisno2:
## Parent

#442 (verified production deployment path). Rationale: ADR-0003.

## What to build

Enable the production migration workflow in the templates and generated config **without changing the zero-setup SQLite local default**. Add `migrate` (`prisma migrate dev`) and `migrate:deploy` (`prisma migrate deploy`) scripts to the `starter` and `starter-auth` templates. Change the generated `prisma.config.ts` datasource so migrations can use a direct (non-pooled) connection on serverless Postgres while local SQLite is untouched: `datasource.url = env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')`. Fix the stale "no migration support — use `db push`" guidance in `CLAUDE.md` (and any CLI generator output that asserts it).

## Acceptance criteria

- [ ] `starter` and `starter-auth` have `migrate` and `migrate:deploy` scripts (alongside the existing `db:push`).
- [ ] The generated `prisma.config.ts` datasource is `env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')`; local SQLite `generate` / `db:push` are unaffected (with `DIRECT_DATABASE_URL` unset it falls back to `DATABASE_URL`).
- [ ] The "no migration support" line is removed/corrected in `CLAUDE.md` (and any generator message that repeats it).
- [ ] The existing scaffold smoke test + core/CLI suites stay green; the `prisma.config.ts` generator change is covered by a test.
- [ ] `pnpm lint` / `pnpm format` / `pnpm manypkg check` clean; changeset per repo convention for any `packages/*` changes (the `prisma.config` generator lives in `@opensaas/stack-cli`; template script changes feed `create-opensaas-app`).

## Blocked by

None - can start immediately.

Issue #442, opened 2026-06-01
PRD: Verified production deployment path for scaffolded apps (HITL)


borisno2:
> **HITL / `ready-for-human`.** This PRD is intentionally not AFK-ready: it needs maintainer decisions (target platform, migration strategy) and credentials (hosting + database accounts, secrets) before it can be sliced. Treat the "Open decisions" section as the agenda for a `grill-with-docs` session. Follow-up to the onboarding initiative (PRD #418); deployment was explicitly deferred there.

## Problem Statement

A developer who has built an app with OpenSaaS Stack (scaffolded via `npm create opensaas-app`, running locally on SQLite) currently has **no verified, friction-free path to production**. A `docs/content/guides/deployment.md` guide exists (Vercel + Neon), but it hasn't been validated end-to-end against the current scaffolder, and several real gaps sit between "runs locally" and "runs in production":

- The scaffolder defaults to **SQLite** with the `PrismaBetterSqlite3` adapter — perfect for zero-setup local dev, but production needs **PostgreSQL** (provider switch + a different Prisma driver adapter + a real `DATABASE_URL`).
- The documented local workflow uses **`prisma db push`** (no migration history). Production needs a **migration** workflow, and the framework's generators currently advertise "no migration support — use `db push`".
- Secrets/env differ in prod (`DATABASE_URL`, and for auth `BETTER_AUTH_SECRET` + the canonical app URL), and the build step (`pnpm build` = `generate` + `next build`) and platform build config need to be correct.

The result: the "third step" of a great onboarding (build it, then ship it) is missing, and a novice can't confidently get their Claude-Code-built app live.

## Solution

A **verified, documented, low-friction production deployment path** for a scaffolded OpenSaaS Stack app — proven by actually deploying the `starter` and `starter-auth` templates to a real environment. Concretely:

- A painless **SQLite (dev) → PostgreSQL (prod)** switch: change the `provider` and driver adapter and set `DATABASE_URL`, with the scaffolder/docs making this a small, well-signposted change (ideally one documented step, optionally a flag).
- A first-class **migration workflow** for production (adopt `prisma migrate` via the generated `prisma.config.ts`), distinct from the local `db push` loop, documented and reflected in template scripts.
- A canonical, **validated deployment guide** (a primary target platform, plus the env/secret checklist, build command, and post-deploy migration step), discoverable from the README and the docs site.
- Correct production build configuration (e.g. not bundling the dev-only SQLite driver; Better-auth prod URLs).

[...]

PR #456, opened 2026-06-02
feat(cli): enable production migrate workflow (migrate scripts + pooled/direct prisma.config)

borisno2:
Implements #450. Part of #442. Rationale: [ADR-0003](https://github.com/OpenSaasAU/stack/blob/main/docs/adr/0003-deployment-uses-postgres-and-prisma-migrate.md).

Enables the production migration workflow **without** changing the zero-setup SQLite local default.

### Generated `prisma.config.ts` datasource (`@opensaas/stack-cli`)

The generated datasource URL now prefers `DIRECT_DATABASE_URL` and falls back to `DATABASE_URL`, so migrations can use a direct (non-pooled) connection on serverless Postgres (e.g. Neon) while the running app connects through the pooled `DATABASE_URL`. Local SQLite is untouched — with `DIRECT_DATABASE_URL` unset, the expression resolves to `DATABASE_URL`.

```typescript
import 'dotenv/config'
import { defineConfig } from 'prisma/config'

// Returns undefined for missing vars so the `??` fallback can take effect.
const env = (name: string): string | undefined => process.env[name]

export default defineConfig({
  schema: 'prisma',
  datasource: {
    url: env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL'),
  },
})
```

**Why a local `env` helper?** The `env` helper exported by `prisma/config` *throws* when a variable is unset (it does not return `undefined`). A literal `env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')` using the upstream helper would throw on the first call before `??` ever evaluated, breaking the SQLite/local fallback. The emitted local helper returns `undefined` for missing vars so the `??` fallback works as intended. Verified: the generated file type-checks (`tsc --noEmit`) and `db:push` against SQLite succeeds with `DIRECT_DATABASE_URL` unset.

### Template scripts (`starter` / `starter-auth`)

Added alongside the existing `db:push`, matching the same prisma-config invocation convention (bare `prisma ...`, which auto-discovers `prisma.config.ts`):

```json
"migrate": "prisma migrate dev",
"migrate:deploy": "prisma migrate deploy"
```

These examples feed `create-opensaas-app`'s `basic` / `with-auth` templates at build time. The committed generated `prisma.config.ts` in both examples was regenerated to the new format.

### Stale "no migration support" guidance corrected

- `CLAUDE.md` "Generator Limitations" — was `No migration support (use prisma db push)`.
- `docs/content/core-concepts/generators.md` — same line.
- `packages/cli/CLAUDE.md` — updated the `prisma.config.ts` example snippet to the new datasource form.

## Tests

- New `packages/cli/src/generator/prisma-config.test.ts` asserts the emitted datasource contains the `DIRECT_DATABASE_URL ?? DATABASE_URL` fallback and the local `env` helper (plus a full snapshot).
- CLI suite: 204 passed.
- Core suite: 651 passed.
- Scaffold smoke test: ran and passed (executed `generate` + `db:push` against SQLite — local default unaffected).
- `pnpm lint` (0 errors), `pnpm format`, `pnpm manypkg check` clean.

Changeset: `@opensaas/stack-cli` minor (new migrate capability). No `create-opensaas-app` changeset — its source/templates are unchanged; only the example `package.json` scripts changed (examples are not published).

Do not merge.

---

### DIFF
--- a/.changeset/wise-otters-migrate.md
+++ b/.changeset/wise-otters-migrate.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-cli': minor
+---
+
+Generate a `prisma.config.ts` datasource that supports the production `prisma migrate` workflow.
+
+The generated datasource URL now prefers `DIRECT_DATABASE_URL` and falls back to `DATABASE_URL`, so migrations can use a direct (non-pooled) connection on serverless Postgres (e.g. Neon) while the running app connects through the pooled `DATABASE_URL`. Local SQLite is unaffected: with `DIRECT_DATABASE_URL` unset, the expression resolves to `DATABASE_URL`.
+
+```typescript
+// generated prisma.config.ts
+import 'dotenv/config'
+import { defineConfig } from 'prisma/config'
+
+// Returns undefined for missing vars so the `??` fallback can take effect.
+const env = (name: string): string | undefined => process.env[name]
+
+export default defineConfig({
+  schema: 'prisma',
+  datasource: {
+    url: env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL'),
+  },
+})
+```
+
+To use a direct connection for migrations on serverless Postgres, set `DIRECT_DATABASE_URL` in your environment; `prisma migrate dev` / `prisma migrate deploy` will use it. See ADR-0003.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1266,7 +1266,7 @@ The `FieldRenderer` extracts `component` and `fieldType` from `ui` options, then
 
 Current generators are basic:
 
-- No migration support (use `prisma db push`)
+- Migrations: the generated `prisma.config.ts` supports `prisma migrate dev` / `prisma migrate deploy`. Local SQLite dev still uses `prisma db push` for a zero-setup loop; production uses `prisma migrate` (see ADR-0003).
 - No introspection support
 - Limited Prisma features (no raw queries, transactions, etc.)
 
--- a/docs/content/core-concepts/generators.md
+++ b/docs/content/core-concepts/generators.md
@@ -194,7 +194,7 @@ Field-level `extendPrismaSchema` is applied before the global `db.extendPrismaSc
 
 Current generators are basic:
 
-- ❌ No migration support (use `prisma db push`)
+- ✅ Migration support: the generated `prisma.config.ts` supports `prisma migrate dev` / `prisma migrate deploy`. Local SQLite dev still uses `prisma db push` for a zero-setup loop; production uses `prisma migrate` (see ADR-0003).
 - ❌ No introspection support
 - ❌ Limited Prisma features (no raw queries, advanced transactions)
 
--- a/examples/starter-auth/package.json
+++ b/examples/starter-auth/package.json
@@ -9,6 +9,8 @@
     "start": "next start",
     "generate": "opensaas generate",
     "db:push": "prisma db push",
+    "migrate": "prisma migrate dev",
+    "migrate:deploy": "prisma migrate deploy",
     "db:studio": "prisma studio"
   },
   "dependencies": {
--- a/examples/starter/package.json
+++ b/examples/starter/package.json
@@ -9,6 +9,8 @@
     "start": "next start",
     "generate": "opensaas generate",
     "db:push": "prisma db push",
+    "migrate": "prisma migrate dev",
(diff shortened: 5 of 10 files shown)


borisno2 (commented):
## Review — Verdict: APPROVE (posted as Comment; GitHub blocks a formal Approve because the review identity authored the PR)

Reviewed strictly from the PR diff plus the code it touches. Core correctness claim verified empirically, not just by reading the description.

### Summary
Generated `prisma.config.ts` now resolves the CLI datasource to `env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')`, emitting a local `env` helper instead of importing the throwing one from `prisma/config`. Adds `migrate` / `migrate:deploy` scripts to both starter examples, corrects the stale "no migration support" docs, and ships a test + changeset. Implements #450 per ADR-0003.

### Verified (the risky bits)

**1. Local `env` helper is necessary and correct — confirmed at the source.**
`@prisma/config@7.5.0` (`dist/index.js`, re-exported by `prisma/config`) defines:
```js
function env(name) {
  const value = process.env[name];
  if (!value) { throw new PrismaConfigEnvError(name); }
  return value;
}
```
So `env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')` with the upstream helper would throw on the first call when `DIRECT_DATABASE_URL` is unset, before `??` ever evaluates. The author's rationale holds. Removing the named import and declaring `const env = (name: string): string | undefined => process.env[name]` means there is no shadowing conflict (the import is gone), and `datasource.url?: string` accepts `string | undefined`, so it type-checks. `tsc --noEmit prisma.config.ts` → exit 0.

**2. Prisma actually reads the datasource via the fallback at CLI time — proven end-to-end.**
Ran real `prisma db push` from `examples/starter`:
- `DIRECT_DATABASE_URL` **unset**, `DATABASE_URL=file:...` → "Loaded Prisma config from prisma.config.ts", SQLite DB created and schema synced. Zero-setup local default is intact.
- `DIRECT_DATABASE_URL` **set** → it takes precedence; the pooled `DATABASE_URL` DB was not created. Precedence matches the documented pooled-app / direct-migration split (ADR-0003).

**3. Test coverage is real.** `prisma-config.test.ts` asserts the `DIRECT_DATABASE_URL ?? DATABASE_URL` expression, asserts the old `import { defineConfig, env }` is gone, asserts the local helper, and snapshots the full output. Ran locally: 4/4 pass.

**4. Scripts, changeset, docs.** `migrate` + `migrate:deploy` present in **both** `examples/starter` and `examples/starter-auth`, using the bare-`prisma` convention matching `db:push`. Changeset `@opensaas/stack-cli` **minor** is appropriate. The stale "no migration support" assertion is corrected in root `CLAUDE.md`, `docs/content/core-concepts/generators.md`, and `packages/cli/CLAUDE.md`. Remaining `db push` mentions elsewhere (ADR-0003, CHANGELOGs, the Keystone join-table guide in `CLAUDE.md:923`) are legitimate context, not the limitation assertion. No new `any`/casts in the diff.

### Findings

**Low / nit — other example `prisma.config.ts` files now drift from the generator output.** Ten git-tracked example configs (`examples/blog`, `auth-demo`, `custom-field`, `composable-dashboard`, `mcp-demo`, `json-demo`, `file-upload-demo`, `tiptap-demo`, `rag-ollama-demo`, `rag-openai-chatbot`) still carry the old `import { defineConfig, env }` + `url: env('DATABASE_URL')` form. Running `pnpm generate` in any of them would now produce a diff. Harmless functionally (single-DB, `db:push` works), and they are not consumed by the scaffolder, so this isn't blocking — but regenerating them in a follow-up would keep committed generated output consistent.

**Non-issue (checked so reviewers don't have to).** The `create-opensaas-app` `templates/basic` and `templates/with-auth` `prisma.config.ts` look stale locally, but `templates/` is gitignored and rebuilt by `copy-templates.ts` (`emptyDir` then copy `examples/starter` → `basic`, `examples/starter-auth` → `with-auth`). So the scaffold output a user receives comes from the starter examples this PR updated — correct.

### CI
`e2e`, GitGuardian, Vercel green; `test` still running at review time (not a blocker).

### Verdict
**APPROVE.** Correctness verified empirically (env throw-behavior, fallback resolution, precedence, type-check, tests). The only follow-up is the optional cosmetic regeneration of the other example configs. Note the PR body says "Do not merge" — respecting that; this is a content approval, not a merge instruction.

---

Issue #446, opened 2026-06-01
CI guards 2: isolated scaffold first-run guard (real CLI, no network install) in the e2e job


borisno2:
## Parent

#443 (testing & CI guards). Rationale: ADR-0002 (PR #444).

## What to build

Replace the in-place scaffold smoke test with an **isolated first-run guard**: in an OS temp dir, invoke the **real `create-opensaas-app` binary** against the local templates, then run `generate` + `db:push` using the workspace toolchain — **without a networked `pnpm install`**. Wire it into the `e2e` CI job (not the fast unit lane). This guards the documented onboarding promise (scaffold → generate → db:push works) end-to-end. Per ADR-0002.

## Acceptance criteria

- [ ] The guard runs the actual built `create-opensaas-app` CLI, producing a project in an OS temp dir (it no longer mutates `examples/starter` in place).
- [ ] It runs `generate` + `db:push` and asserts the expected artifacts exist (`.opensaas/context.ts`, a created `dev.db`) and that a runnable SQLite `.env` was written.
- [ ] No networked `pnpm install` / npm-registry round-trip happens on the PR gate; `generate`/`db:push` use the workspace toolchain.
- [ ] The temp dir is cleaned up and the repo working tree is never mutated (verified by a clean `git status` after the run).
- [ ] Runs in the `e2e` job and is green in CI.
- [ ] Follow repo conventions (`pr-changeset`) for any `packages/*` changes.

## Blocked by

None - can start immediately.

Issue #443, opened 2026-06-01
PRD: Wire tests & CI guards for the onboarding + deep-module work (HITL)


borisno2:
> **HITL / `ready-for-human`.** Not AFK-ready: it changes CI workflows, may need a repo secret (`OPENAI_API_KEY` for the docs build), and requires maintainer policy calls (coverage gating levels, which flows to e2e, CI time/cost budget). The "Open decisions" section is the agenda for a `grill-with-docs` session. Follow-up that hardens the onboarding (PRD #418) and architecture-deepening (#434–#437) work now merged to `main`.

## Problem Statement

The recent onboarding and architecture-deepening work added genuinely good tests — the scaffold smoke test, the adapter-guidance regression guard, and unit suites for the Write Pipeline, Hook Pipeline, nested-op registry, and field-config validator. But several **test-hygiene and CI-guard gaps** surfaced during that work and remain unaddressed, so the guarantees are weaker than they look:

- **`dist` test-duplication:** when a build runs before tests, Vitest also picks up compiled `dist/**/*.test.js`, inflating counts (e.g. core showed 507 clean vs ~607 with `dist`). This makes coverage/counts misleading and risks double-running stale compiled tests.
- **The scaffold smoke test mutates `examples/starter` in place** (writes `.env`, runs `generate`/`db:push`, then cleans up) rather than an isolated temp dir — a hard crash mid-run could dirty the tree, and it's a proxy for, not a real run of, the published `npm create` flow.
- **No true end-to-end guard for the first-run flow** (`npm create` → install → generate → db:push → build) or the **auth sign-in flow**, even though a Playwright harness and a `fix-e2e` skill exist.
- **Coverage is reported per-package but never gated** — there are no thresholds, so coverage can silently erode (the new deep modules are well-covered today, with nothing keeping them that way).
- **A known example build failure** (`examples/rag-ollama-demo`, observed during this initiative) was never ticketed and isn't caught as a CI gap.
- **The docs build requires `OPENAI_API_KEY`** at build time (for the `/api/search` route); without it, `pnpm --filter opensaas-stack-docs build` fails at page-data collection — a fragile, undocumented CI/build dependency.

## Solution

Turn the existing tests into **durable CI guards** and fix the hygiene issues, so the onboarding and deep-module guarantees hold over time:

- Stop Vitest from picking up `dist` test duplicates, so counts/coverage are honest.
[...]

PR #458, opened 2026-06-02
test(create-opensaas-app): isolated scaffold first-run guard in the e2e job

borisno2:
Implements #446. Part of #443. Rationale: ADR-0002 (`docs/adr/0002-testing-and-ci-strategy.md`).

## What changed

Replaces the in-place scaffold smoke test (which mutated `examples/starter` in place — wrote its `.env`, ran `generate`/`db:push`, then restored) with an **isolated first-run guard** that exercises the real published flow without touching the repo tree.

- **New guard** `packages/create-opensaas-app/tests/scaffold-first-run-guard.test.ts`:
  - Runs the **actual built `create-opensaas-app` CLI** (`dist/index.js`) to scaffold a project into an OS temp dir (`os.tmpdir()` via `fs.mkdtemp`).
  - Runs `generate` + `db:push` and asserts the artifacts: `.opensaas/context.ts`, a created `dev.db`, and a runnable SQLite `.env` (`DATABASE_URL="file:./dev.db"`).
  - Cleans up the temp dir and asserts the repo working tree is unchanged (snapshots `git status --porcelain` before vs after, so a clean CI checkout asserts a clean tree while local in-progress edits don't trip it).
- **Removed** the old `tests/scaffold-smoke.test.ts`.
- **CLI**: added `--no-auth` / `--no-ai` flags so the CLI can scaffold fully non-interactively (no prompts, no networked MCP install). Symmetric with the existing `--with-auth` / `--with-ai` and `--no-install`.
- **CI**: the guard runs in the **`e2e` job** (not the fast unit lane). The job now builds `create-opensaas-app` and runs the guard with `RUN_SCAFFOLD_GUARD=1`.

## How the toolchain is resolved offline (no network install)

The key integration challenge: a project scaffolded into `os.tmpdir()` has no `node_modules`, and a real `pnpm install` would hit the npm registry — not allowed on the PR gate.

The guard scaffolds with `--no-install` (skipping the CLI's auto `install → generate → db:push`), then **symlinks the workspace example's already-installed `node_modules`** (`examples/starter/node_modules`, populated by the repo's own `pnpm install` in CI setup) into the temp project. That gives the project the exact toolchain the working example resolves — the `opensaas` CLI, Prisma, the SQLite adapter, `jiti`, `dotenv` — entirely from deps already on disk. `generate`/`db:push` then run via the project's `node_modules/.bin`, so there is no registry round-trip. This is documented in the test's header comment. The symlink is removed before deleting the temp dir so the real workspace `node_modules` is never recursed into or deleted.

## Lane separation

The guard is gated behind `RUN_SCAFFOLD_GUARD=1` so it stays out of the fast unit lane (`turbo run test`, where the built artifacts also happen to exist) and runs only in the `e2e` job. It also self-skips (mirroring the existing `describe.skipIf` pattern) when the built CLI / example `node_modules` aren't present, so a bare `vitest` without a prior `pnpm install && pnpm build` skips rather than fails.

## Verification (local)

- Guard (exact CI command): `RUN_SCAFFOLD_GUARD=1 pnpm --filter create-opensaas-app test run scaffold-first-run-guard` → green (project created in temp dir, artifacts present, `.env` correct, temp dir cleaned, tree unchanged).
- Unit lane (no flag): full `create-opensaas-app` suite green in ~1.3s with the heavy guard skipped.
- `pnpm lint` (0 errors), `pnpm format:check`, `pnpm manypkg check` all clean.

---


borisno2 (commented):
## Review: isolated scaffold first-run guard (e2e)

**Verdict: APPROVE** (posted as a Comment because GitHub blocks a formal Approve on a self-authored PR) — well-constructed, thoroughly documented, and the disk-safety / lane-separation concerns hold up under scrutiny. A couple of minor non-blocking notes below.

Reviewed strictly from the PR diff plus the surrounding files it depends on (`test.yml`, `.github/actions/setup/action.yml`, `src/index.ts`, `src/lib/env.ts`, `examples/starter/{package.json,opensaas.config.ts}`).

### Risk areas — all verified

**1. Disk mutation / symlink safety (the big one): SAFE.**
- `afterAll` ordering is correct: it `fs.unlink`s the `node_modules` symlink *before* `fs.remove(tmpRoot)`, so `fs.remove` can never recurse into the real `examples/starter/node_modules`. (`fs.unlink` removes the link, not its target.)
- Belt-and-suspenders: even without the explicit unlink, `fs-extra`'s `remove` unlinks symlinks rather than following them — but the explicit unlink is the right defensive choice and is correctly placed.
- `beforeAll` failure paths are safe: `projectDir` is only assigned after the scaffold `run` resolves, so a mid-scaffold crash leaves `projectDir === ''`, the unlink block is skipped, and `fs.remove(tmpRoot)` still cleans up.

**2. `git status` cleanliness assertion: PRESENT and well-designed.**
Snapshots `git status --porcelain` in `beforeAll` and compares in a dedicated test. Comparing before-vs-after (rather than asserting an empty tree) means a clean CI checkout still asserts a clean tree, while local in-progress edits don't produce false failures.

**3. e2e gating / lane separation: CONFIRMED.**
- Guard is gated behind `RUN_SCAFFOLD_GUARD=1` **and** `describe.skipIf(!prerequisitesPresent)`, so it stays out of `turbo run test` and self-skips without built artifacts.
- The `e2e` job builds `create-opensaas-app` (added to the `--filter` list) **and** `@opensaas/stack-cli` (already filtered) — the latter provides `packages/cli/dist/index.js` that the scaffolded project's `generate` invokes.
- `.github/actions/setup` runs `pnpm install --frozen-lockfile`, which populates `examples/starter/node_modules` (the borrowed toolchain). All three prerequisites (`createCli`, `opensaasCli`, `toolchainNodeModules`) are satisfied in the e2e job.
- Starter SQLite expectations line up end-to-end: `examples/starter/opensaas.config.ts` uses `provider: 'sqlite'` + better-sqlite3, and `env.ts` emits exactly `DATABASE_URL="file:./dev.db"` — matching the test's `.env` and `dev.db` assertions.

**4. New CLI flags `--no-auth` / `--no-ai`: no behaviour regression, no `any`/casts.**
The prompts are now skipped only when the opt-out flag is set (`!hasAuthFlag && !hasNoAuthFlag`). Default interactive behaviour is unchanged when no flag is passed. Plain `args.includes(...)` booleans — no casts, strongly typed.

**5. Old test removed + changeset present.**
`tests/scaffold-smoke.test.ts` is deleted (not left duplicating), and `.changeset/quiet-foxes-gather.md` is a `create-opensaas-app: patch`. Correct.

### Minor / non-blocking notes (ordered by severity)

1. **Residual write-leak risk the `git status` assertion can't catch.** `prisma db push` runs against the *shared* (symlinked) `node_modules`. The tree-cleanliness check won't notice writes into `node_modules` because it's gitignored. In practice Prisma 7 + driver-adapter `db push` only touches the DB file and the project-local `prisma/schema.prisma` (a real dir, not the symlink), so this shouldn't mutate the borrowed modules — but it's the one place a write *could* leak into the workspace. Worth a mental note if the guard ever starts flaking the example build. (`tests/scaffold-first-run-guard.test.ts`, the `generate`/`db:push` block)

2. **Contradictory flags resolve silently.** Passing both `--with-auth` and `--no-auth` (or `--with-ai`/`--no-ai`) silently lets the `--with-*` side win (`let withAuth = hasAuthFlag`). Not a regression and not exercised here; a future hardening could error on conflicting flags. (`src/index.ts`)

Neither blocks merge. Nice work — the file header comment documents the offline-toolchain trick clearly enough that the next reader won't be surprised.

---

Issue #452, opened 2026-06-01
create-opensaas-app: --db postgres scaffolder flag (deferred follow-up)


borisno2:
Deferred follow-up from the deployment PRD #442 / ADR-0003 (not a sub-issue — a standalone enhancement). The deployment grilling chose a documented one-time manual SQLite→Postgres swap for now; this issue tracks the optional ergonomic improvement.

## What to build

An optional `--db postgres` flag (and matching interactive prompt) for `create-opensaas-app` that scaffolds a Postgres-ready project from the start, instead of the SQLite default:

- emit the `PrismaPg` (or `PrismaNeon`) driver adapter in the generated `opensaas.config.ts` and add the dependency;
- write a Postgres `.env` / `.env.example` with `DATABASE_URL` (pooled) + `DIRECT_DATABASE_URL` placeholders;
- include the `migrate` / `migrate:deploy` scripts;
- skip the SQLite-only steps.

Reuse the existing env/datasource generator (the pure helper from the onboarding work) extended for the `postgresql` provider, so the logic stays unit-tested.

## Acceptance criteria

- [ ] `npm create opensaas-app my-app --db postgres` produces a project whose `opensaas.config.ts` uses the pg/Neon adapter, with a Postgres `.env`/`.env.example` and the migrate scripts.
- [ ] Without the flag, **SQLite remains the default** (unchanged behaviour).
- [ ] Unit tests cover the `postgresql` provider branch of the env/datasource generator and the scaffolder wiring.
- [ ] `create-opensaas-app` changeset (minor); lint/format/manypkg clean.

## Blocked by

None — but best done after the deployment guide (#442 slices) lands, so the emitted config matches the documented adapter pattern.

Issue #451, opened 2026-06-01
Deploy 2: correct & complete the deployment guide (adapter-based config + migrate workflow)


borisno2:
## Parent

#442 (verified production deployment path). Rationale: ADR-0003.

## What to build

Rewrite `docs/content/guides/deployment.md` so it matches the **current adapter-based framework** and the `prisma migrate` workflow. The guide is substantial but its config snippet is stale: it shows the pre-adapter `db: { provider, url, directUrl }` with no `prismaClientConstructor`/driver adapter, which Prisma 7 requires. Replace it with the adapter-based config (`PrismaNeon`/`PrismaPg` constructed with the **pooled** `DATABASE_URL` via `prismaClientConstructor`), document the **pooled-app / direct-CLI** split (app → pooled `DATABASE_URL`; migrations → `DIRECT_DATABASE_URL` via `prisma.config.ts`, per slice #450), and make the `migrate dev`/`migrate deploy` steps, Vercel env setup, build/release step, and Better-auth prod URLs accurate and consistent.

## Acceptance criteria

- [ ] No pre-adapter `db: { url, directUrl }` snippet remains; the guide uses the adapter-based config + the `prisma.config.ts` direct-URL approach from #450.
- [ ] The env section lists `DATABASE_URL` (pooled, app), `DIRECT_DATABASE_URL` (migrations), and the Better-auth vars (`BETTER_AUTH_SECRET`, `BETTER_AUTH_URL` = deployed URL, OAuth) — consistent with the templates.
- [ ] Commands use the canonical set; the build/release step runs `prisma migrate deploy` with the direct connection.
- [ ] Internal links resolve; `pnpm format` clean. README links to the guide.

## Blocked by

- #450

PR #463, opened 2026-06-02
feat(create-opensaas-app): add --db postgres flag for Postgres-ready scaffolds

borisno2:
Implements #452. Part of #442.

## What this adds

An optional `--db postgres` flag (and a matching interactive database prompt) for `create-opensaas-app` that scaffolds a PostgreSQL-ready project instead of the SQLite default. Without the flag, **SQLite remains the zero-setup default** (unchanged behaviour).

```bash
# PostgreSQL-ready: pg driver adapter, Postgres .env, migrate scripts
npm create opensaas-app my-app --db postgres

# Force SQLite and skip the database prompt
npm create opensaas-app my-app --db sqlite
```

The interactive flow now asks "Which database? (SQLite default / PostgreSQL)" using the existing `prompts` select, consistent with the auth prompt. `--db postgres` / `--db sqlite` (and the `--db=...` form / the `postgresql` alias) bypass the prompt, keeping the CLI fully non-interactive for CI.

## How the config + deps transform works

Templates are copied verbatim from the SQLite examples (`examples/starter` / `examples/starter-auth`), so a Postgres project is the SQLite template **rewritten deterministically**. The trickiest integration point lives as two **pure, unit-tested helpers** in `packages/create-opensaas-app/src/lib/postgres.ts` (not inline string-munging in `index.ts`):

- **`toPostgresConfig(source)`** — swaps the `opensaas.config.ts` `db` block from `provider: 'sqlite'` + `PrismaBetterSqlite3` to `provider: 'postgresql'` + the documented `PrismaPg` form: `const pool = new pg.Pool({ connectionString: process.env.DATABASE_URL }); const adapter = new PrismaPg(pool); return new PrismaClient({ adapter })`, and updates the imports (`@prisma/adapter-pg` + `import pg from 'pg'`). Matches the exact form in `examples/rag-openai-chatbot/opensaas.config.ts` and the deployment guide (#451). Throws on template drift so a half-converted config never ships.
- **`toPostgresPackageJson(pkg)`** — removes `better-sqlite3` + `@prisma/adapter-better-sqlite3`; adds `@prisma/adapter-pg` + `pg` (and `@types/pg` dev). Scripts (including `migrate` / `migrate:deploy`, already shipped by the starter post-#450) and `@opensaas/*` deps are preserved untouched.

The Postgres `.env` / `.env.example` (via the extended `generateEnvFiles` `postgresql` branch) emit `DATABASE_URL` (pooled) + `DIRECT_DATABASE_URL` (direct) placeholders — the pooled-app / direct-CLI split documented in the deployment guide. The with-auth path swaps only the database block so the Better-auth variables (secret, URLs, OAuth, rate-limit) survive. For Postgres the auto-run setup omits `db:push` (it needs a live DB the user configures first) and the next-steps list `pnpm migrate` instead.

## Dependency versions (matched to the monorepo)

`@prisma/adapter-pg@^7.8.0`, `pg@^8.20.0`, `@types/pg@^8.18.0` — matched to `examples/rag-openai-chatbot` so the repo carries a single version of each.

## Tests added

- `tests/env.test.ts`: the `postgresql` env branch emits both the pooled `DATABASE_URL` and the direct `DIRECT_DATABASE_URL`, and never an active SQLite file URL.
- `tests/postgres.test.ts` (new): the config transform swaps imports + db block, asserts post-conditions (no leftover `sqlite` / `better-sqlite3`), throws on drift, preserves the rest; the package.json transform drops the SQLite deps, adds the pg deps at the monorepo versions, preserves migrate scripts + `@opensaas` deps, and doesn't mutate its input.
- `tests/setup.test.ts`: Postgres omits `db:push` from the auto-run plan and lists `pnpm migrate` in the manual next steps.

## Verification

- `pnpm --filter create-opensaas-app build` — tsc clean, templates copied.
- `pnpm --filter create-opensaas-app test` — 41 passed, 2 skipped (the network-gated scaffold guard).
- Scaffolded `--db postgres`, `--db postgres --with-auth`, and `--db sqlite` projects into temp dirs and asserted: config uses `PrismaPg`, `.env` has `DATABASE_URL` + `DIRECT_DATABASE_URL`, package.json has the pg deps + migrate scripts, no leftover sqlite refs, and the default (no flag) still produces a SQLite project. Ran `opensaas generate` against the Postgres project — produces a valid `provider = "postgresql"` Prisma schema + context.
- `pnpm lint` (0 errors), `pnpm format`, `pnpm manypkg check` — clean.

Changeset: `create-opensaas-app` minor.

---


borisno2 (commented):
## Code Review — `feat(create-opensaas-app): --db postgres`

**Verdict: APPROVE** (submitted as a Comment because GitHub blocks a self-approve — the review identity authored the PR). No blocking issues found. Findings below are all minor/nits.

I reviewed the diff against the actual template sources, the reference adapter pattern, and the monorepo's existing versions.

### Verified correct

- **`toPostgresConfig` transform & drift-guard** — The structural regex matches the *actual* SQLite `db` block in both `examples/starter/opensaas.config.ts` and `examples/starter-auth/opensaas.config.ts` (identical blocks). The emitted form (`const pool = new pg.Pool({ connectionString: process.env.DATABASE_URL })` → `new PrismaPg(pool)` → `new PrismaClient({ adapter })`, imports `@prisma/adapter-pg` + `import pg from 'pg'`) is byte-for-byte the documented pattern in `examples/rag-openai-chatbot/opensaas.config.ts`. It throws on *both* a missing adapter import and a missing db block, so a half-converted config can't ship. No residual `sqlite`/`better-sqlite3` references — covered by an explicit assertion in `tests/postgres.test.ts`.
- **Version consistency** — `@prisma/adapter-pg@^7.8.0`, `pg@^8.20.0`, `@types/pg@^8.18.0` match `examples/rag-openai-chatbot` exactly, and a repo-wide search confirms `@prisma/adapter-pg` is pinned to `^7.8.0` in the only other place it appears. No duplicate-version risk.
- **`toPostgresPackageJson` purity** — Spreads input + clones dep maps; `tests/postgres.test.ts` asserts the input isn't mutated. Drops the two SQLite deps, adds the three pg deps, and preserves `migrate`/`migrate:deploy` scripts, `@opensaas/*`, and `@prisma/client`.
- **Default path unchanged** — `dbProvider` defaults to `'sqlite'`; both transforms are gated on `=== 'postgresql'`. Without `--db postgres` (or choosing SQLite in the prompt) the scaffold output is identical to before.
- **Auth + Postgres** — `replaceAuthEnvDatabase` swaps only the `DATABASE_URL=...` line (`/^DATABASE_URL=.*$/m`), so the Better-auth secret/URL/OAuth/rate-limit vars in `.env.example` survive. Correct.
- **`planSetupSteps` / `nextStepCommands` provider-awareness** — Postgres omits `db:push` from the auto-run and lists `pnpm migrate` in next steps; SQLite path is byte-identical. This aligns with ADR-0003 (production loop = `prisma migrate`, `db push` is explicitly *not* a production path). The emitted `DIRECT_DATABASE_URL` placeholder is correctly consumed by the already-generated `prisma.config.ts` (`env('DIRECT_DATABASE_URL') ?? env('DATABASE_URL')`).
- **Typing & changeset** — `DbProvider` is a `'sqlite' | 'postgresql'` union; no `any`/casts on exported surfaces. A `create-opensaas-app` **minor** changeset is present. Test coverage is solid (config transform, pkg transform, env branch, setup-plan).

### Minor findings (non-blocking)

1. **Stale `db:push` reference in the generated Postgres `.env`** (`src/lib/env.ts`). Line 2 of the pg `env` reads:
   `# Replace with your provider's pooled connection string before \`pnpm db:push\`.`
   But the Postgres flow intentionally omits `db:push` and uses `pnpm migrate`. This comment will send users to the wrong command. Suggest `before \`pnpm migrate\``.

2. **`DbProvider` is defined twice** — `src/lib/env.ts` and now `src/lib/setup.ts` (with a "Mirrors `DbProvider` in `env.ts`" comment). Re-exporting the `env.ts` type from `setup.ts` instead of redeclaring would prevent the two drifting apart.

3. **`parseDbFlag` trailing-`--db`** (`src/index.ts`). With `--db` as the final arg and no value, `args[index + 1]` is `undefined`, so the function returns `undefined` and silently falls through to the prompt/SQLite default rather than erroring like an unknown value does. Minor UX inconsistency; an explicit "missing value" error would match the unknown-value handling.

4. **ADR-0003 is now stale** — it states the `--db postgres` flag "was considered and deferred." This PR implements it; the ADR should be updated (or a follow-up noted) so the documented decision matches reality. The leftover `# Database` comment line above the swapped `DATABASE_URL` in the auth `.env` is cosmetic.

Nice work — the pure-helper extraction with exhaustive post-condition assertions (no leftover sqlite, throws on drift) is exactly the right shape for this kind of deterministic rewrite.

🤖 Generated with [Claude Code](https://claude.ai/code)

---

borisno2:
Addressed the review nits in `78c0558`:

1. **Stale `db:push` reference in the Postgres `.env`** (`src/lib/env.ts`): the pooled-URL comment now says `before \`pnpm migrate\`` (the Postgres production loop) instead of `db:push`. SQLite branch unchanged.
2. **`DbProvider` declared twice** (`src/lib/env.ts` + `src/lib/setup.ts`): removed the duplicate in `setup.ts` and now `import type` + re-export it from `env.ts` (single source of truth). Exported name/shape unchanged.
3. **`parseDbFlag` trailing `--db` with no value** (`src/index.ts`): a bare trailing `--db` (no following value) now hits the same "Unknown --db value" error + non-zero exit as an invalid value, instead of silently defaulting.
4. **ADR-0003 staleness**: appended a dated `Update:` note recording that the deferred `--db postgres` flag has since been implemented (#452 / #463), without rewriting the original decision.

Verification: `pnpm --filter create-opensaas-app build` and `test` (41 passed, 2 skipped — the network-gated scaffold guard) green; `pnpm lint` (0 errors), `pnpm format`, `pnpm manypkg check` clean. No new changeset added (existing `create-opensaas-app` minor changeset already covers this; these are minor polish on it).

---

Issue #488, opened 2026-06-03
Harden migrate-context-calls skill: where-shape, connect/disconnect, gql.tada, include/select (with fixtures)


borisno2:
## Parent

#486

## What to build

Close the gaps in the `migrate-context-calls` skill (in `claude-plugins/opensaas-migration`) that replaces `context.graphql.run`/`context.query.*` with `context.db.*`. The skill handles CRUD well but omits the hard parts. Add four recipes, each with a compiling **before/after fixture**:

1. **where-shape translation** — a systematic mapping from Keystone relation-filter shapes (`{ author: { id: { equals } } }`) to Prisma scalar-FK / relation filters (`{ authorId: { equals } }`).
2. **connect/disconnect/set nested writes** — map Keystone nested-write mutations to Prisma's relation operations on create/update.
3. **gql.tada typed documents** — recipe for replacing gql.tada typed documents (`TadaDocumentNode`, `ResultOf`/`VariablesOf`) with `defineFragment` + `ResultOf`.
4. **fragment → include/select + null semantics** — clarify how a fragment maps to Prisma `include`/`select` and the null-on-access-denied behaviour for nested relations.

No GraphQL adapter is built (ADR-0005): fragments + this skill are the parity path.

## Acceptance criteria

- [ ] The skill documents all four recipes, each with a before (Keystone/gql.tada) and after (`context.db.*` / `defineFragment`) fixture that compiles conceptually.
- [ ] The where-shape recipe includes a translation table covering relation filters, scalar filters, and common operators.
- [ ] The connect/disconnect/set recipe covers create and update relation operations.
- [ ] The gql.tada recipe shows `TadaDocumentNode` + `ResultOf`/`VariablesOf` → `defineFragment` + `ResultOf`.
- [ ] Plugin version bumped (see `plugin-version` skill) for the `opensaas-migration` change.

## Blocked by

None - can start immediately

Issue #486, opened 2026-06-03
PRD: Migration docs & agent context — canonical guide, fix 404s, harden graphql.run→db skill, CLI discoverability


borisno2:
## Problem Statement

Migration knowledge is scattered and partly unreachable. There are two overlapping guides — `specs/keystone-migration.md` (detailed Keystone recipes) and `docs/content/guides/migration.md` (published, AI-assisted overview) — and the conceptually-referenced `guides/migrating-from-keystone.md` 404s, as do `core-concepts/fields.md` (actual file: `field-types.md`) and `core-concepts/queries.md` (no such page). The `migrate-context-calls` skill that replaces `context.graphql.run` with `context.db.*` is solid for CRUD but omits the hard parts migrators trip on: Keystone-relation-filter → Prisma scalar-FK **where-shape translation**, `connect`/`disconnect`/`set` nested writes, gql.tada typed-document replacement, and fragment → `include`/`select` semantics. Finally, the migration docs/tools are hard to find from where a migrator starts.

This PRD also carries the folded-in GraphQL-parity work (originally SF-02): the stack will NOT ship a GraphQL adapter (ADR-0005); migration is via fragments + the hardened skill + the guide.

## Solution

Make one canonical, published Keystone → stack guide the single source of truth; fix the 404s by creating real pages/aliases at the referenced paths; harden the `migrate-context-calls` skill with the four missing recipes and before/after fixtures; and make the guide + migration plugin discoverable from `opensaas migrate` and the README/docs nav. The comprehensive umbrella guide reflects the new Area A–E behaviour and links to the detailed image (#479) and auth (#483) sections rather than duplicating them.

## User Stories

1. As a migrating developer, I want a single canonical published Keystone→stack guide, so that I'm not reconciling `specs/` against the docs site.
2. As a migrating developer or agent, I want `guides/migrating-from-keystone.md` to resolve instead of 404, so that external and agent references work.
3. As a developer, I want a queries/fragments docs page documenting `defineFragment`/`runQuery`/`ResultOf`, so that I can learn the `context.graphql.run` replacement.
4. As a developer, I want `core-concepts/fields.md` to resolve (alias to `field-types.md`), so that older references don't break.
5. As a migrating developer, I want the `migrate-context-calls` skill to cover where-shape translation (Keystone `{ author: { id: { equals } } }` → Prisma `{ authorId: { equals } }`), so that I rewrite `where` clauses correctly.
[...]

PR #492, opened 2026-06-03
Harden migrate-context-calls skill: where-shape, connect/disconnect, gql.tada, include/select recipes

borisno2:
Implements #488
Part of #486

Adds four migration recipes to the `migrate-context-calls` skill (`claude-plugins/opensaas-migration`), each with a before (Keystone/gql.tada) and after (`context.db.*` / `defineFragment`) fixture that compiles conceptually against the real API in `packages/core/src/query/index.ts`:

1. **`where`-shape translation** — collapses Keystone relation filters (`{ author: { id: { equals } } }`) onto Prisma scalar FKs (`{ authorId: { equals } }`), with a translation table covering relation filters, scalar filters, common operators, and the enum-string-literal gotcha.
2. **`connect` / `disconnect` / `set` nested writes** — maps Keystone nested-write mutations to Prisma relation operations on both create and update, with a nested-write table and the "never write the scalar FK directly" / "`set` replaces, `connect` adds" caveats.
3. **gql.tada typed documents** — replaces `TadaDocumentNode` + `ResultOf`/`VariablesOf` with `defineFragment` + `ResultOf` (and a fragment factory for the `VariablesOf` case), plus a mapping summary.
4. **fragment → `include`/`select` + null semantics** — shows how a fragment maps to a Prisma `include` and documents null-on-access-denied for top-level reads, nested single relations (`T | null`), nested to-many relations (empty array, never null), and field-level access.

The Steps section now references each recipe so the agent applies them during migration.

No GraphQL adapter is built (ADR-0005); fragments + this skill are the parity path.

Plugin version bumped: `opensaas-migration` 0.2.3 → 0.3.0 (new capability) in both `plugin.json` and the `marketplace.json` entry. Marketplace `metadata.version` left unchanged (no marketplace-structure change). `pnpm manypkg fix` and `prettier` run clean.

---

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -29,7 +29,7 @@
       "name": "opensaas-migration",
       "source": "./claude-plugins/opensaas-migration",
       "description": "OpenSaaS Stack migration assistant with AI-guided configuration, schema analysis, and code generation support",
-      "version": "0.2.3",
+      "version": "0.3.0",
       "author": {
         "name": "OpenSaaS Team",
         "url": "https://github.com/OpenSaasAU/stack"
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ pnpm-debug.log*
 # Turbo
 .turbo/
 .playwright-mcp
+
+# Background agent worktrees (transient, auto-cleaned)
+.claude/worktrees/
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -33,3 +33,23 @@ _Avoid_: operation handler, mutation service
 **Hook Pipeline**:
 The module that runs the transform+validate span of a write — list `resolveInput` → field `resolveInput` → list `validate` → field `validate` → built-in field rules — owning that order and the `resolvedData` threading through it. It throws a validation error (never silent) when a validate hook reports via `addValidationError` or a built-in field rule fails, and returns the transformed `resolvedData` on success. The Write Pipeline delegates this span to it; side-effect hooks (`beforeOperation`/`afterOperation`), access, writable-field filtering, persistence and Field Visibility stay in the Write Pipeline.
 _Avoid_: validation service, input resolver
+
+### Migration & schema generation
+
+**Schema parity**:
+The property that the generator's Prisma schema diffs clean — an empty `prisma migrate diff` in both directions — against a pre-existing (typically Keystone-generated) database, so a project can adopt the stack without a destructive migration. It is the go/no-go gate for every migration phase.
+_Avoid_: schema match, clean diff, byte-compatibility
+
+**Keystone-compat mode**:
+An opt-in generator setting that makes schema output follow Keystone 6 conventions a migrating project depends on but a greenfield project would not want by default (e.g. non-null text columns defaulting to `""`). Conventions that every project wants are the plain default, not part of this mode.
+_Avoid_: legacy mode, compatibility flag, migration mode
+
+### Authentication
+
+**Auth lists**:
+The user/session/account/verification lists the auth plugin derives from the better-auth config — their keys, table/column maps, and database schema all follow that config, so they can be modelled to match (adopt) pre-existing better-auth tables. Distinct from any application domain User.
+_Avoid_: auth tables, auth models, auth schema
+
+**Auth identity**:
+The better-auth-owned record of who a session belongs to (the better-auth user). Separate from, and not assumed to be, the application's own domain User; an app links the two itself when it needs to.
+_Avoid_: auth user, principal, account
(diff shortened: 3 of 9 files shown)


borisno2 (commented):
## Review — PR #492 (harden `migrate-context-calls` skill)

**Verdict: APPROVE** (submitted as a Comment review because the reviewer is the PR author and GitHub blocks a formal Approve from the author — the verdict is the intended signal).

I reviewed strictly from the diff and cross-checked every fixture against the real runtime in `packages/core/src/query/index.ts`, `packages/core/src/context/index.ts`, and `packages/core/src/access/field-access.ts`. The four recipes are technically accurate and do not invent API.

### API cross-check (the main risk) — passes
- **`defineFragment` / `ResultOf` / `RelationSelector`** are real exports of `packages/core/src/query/index.ts`. The curried signature `defineFragment<Model>()({ ... } as const)` and `ResultOf<typeof fragment>` match the source exactly.
- **`RelationSelector` shape** `{ query, where?, orderBy?, take?, skip? }` in Recipe 3/4 matches the `RelationSelector` union in `query/index.ts` verbatim (including the to-many nested-args form).
- **The `query` parameter on `context.db.*`** is real: both `createFindMany` and `createFindUnique` in `context/index.ts` accept `query?` and route it through `buildInclude` + `pickFields`. Recipe 4's `findUnique({ where: { id }, query: postFragment })` matches the implemented `{ where: { id: string }, query? }` signature, and the "equivalent `prisma.post.findFirst({ where, include })`" note is correct (`findUnique` internally calls `findFirst` and merges the access filter).
- **Recipe 1 where-shape translation** is correct: Keystone relation filter `{ author: { id: { equals } } }` → scalar FK `{ authorId: { equals } }`; non-id relation fields stay nested; to-many uses `some/every/none`; `{ author: null }` → `{ authorId: null }`; logical operators and `mode: 'insensitive'` are valid Prisma. The enum-string-literal gotcha (quote `'published'`) is the right call-out.
- **Recipe 2 connect/disconnect/set** maps correctly (`set` replaces, `connect` adds, `set: []` clears, `disconnect: true` for nullable single relations). The "never write the scalar FK directly" caveat is backed by `filterWritableFields`, which I verified strips `${field}Id` when a relationship field exists.

### Findings (by severity)

**Major — pre-existing, NOT introduced by this PR (no change required here):**
- The recipes import `defineFragment` / `ResultOf` / `runQuery` / `runQueryOne` from `@opensaas/stack-core` (root), but the root entry `packages/core/src/index.ts` does **not** re-export anything from `./query/index.js`. `extend.ts` doesn't export them either, and `internal.ts` exports only the *types* `Fragment` / `FieldSelection`. So that import currently won't resolve at the documented path. This is pre-existing: the same import already appears in the unchanged top of this SKILL.md, in `packages/core/CHANGELOG.md` (which announces these as shipped), `specs/keystone-migration.md`, and `docs/content/guides/migration.md`. The PR faithfully follows the established (but currently un-wired) convention, so it's not a regression — but it should be tracked as a separate core fix (add `export * from './query/index.js'` to the root, including the `RelationSelector` type used in Recipe 4 prose).

**Minor:**
- Recipe 2 caveat says `filterWritableFields` "strips `<field>Id` keys when a relationship field exists." Accurate for the `authorId` single-relation example shown, but the source only does this for **non-many** relations (`if (!relConfig.many)`). Many-relations have no scalar FK column anyway, so the practical advice is right; consider narrowing the wording to "single-relation `<field>Id`".
- Scope: the PR description and linked issue (#488) are about the four skill recipes, but the diff also adds ADRs 0004/0006/0007 and four CONTEXT.md glossary entries that describe *future/intended* designs (e.g. "userListKey currently a hardcoded `'user'` TODO", image/file multi-column mode, auth-list adoption) not implemented in this branch. ADR-0005 is directly load-bearing for the skill; 0004/0006/0007 are tangential to the stated scope. Harmless as docs, but flagging the scope creep.

**Confirmed correct:**
- Plugin version bump is consistent: `claude-plugins/opensaas-migration/.claude-plugin/plugin.json` and the `opensaas-migration` entry in `.claude-plugin/marketplace.json` both go `0.2.3 → 0.3.0`. Minor bump for a new capability is appropriate, and leaving `marketplace.metadata.version` untouched matches the rule (no marketplace-structure change).
- `.gitignore` addition for `.claude/worktrees/` is reasonable and harmless.

### Bottom line
Approve. The recipes match the real API and are a genuinely useful hardening of the migration skill. The one substantive concern (root export of the fragment helpers) is a pre-existing core gap that this PR neither introduces nor is responsible for fixing, but it should be filed as a follow-up so the import paths these recipes (and the existing docs/CHANGELOG) advertise actually resolve.

_Reviewed by Claude Code._

---

Issue #471, opened 2026-06-03
Field defaultValue → @default: formatPrismaDefault module + text/integer/json


borisno2:
## Parent

#470

## What to build

A shared, pure `formatPrismaDefault` module that serialises a field's `defaultValue` into a Prisma `@default(...)` literal, plus wiring of `text()`, `integer()`, and `json()` field builders to use it so a configured `defaultValue` appears in the generated schema.

Serialisation rules:
- `integer({ defaultValue: 3550 })` → `@default(3550)` (bare numeric literal; nullability `?` handled independently).
- `text({ defaultValue: 'PLEASE_UPDATE' })` → `@default("PLEASE_UPDATE")` (quoted string).
- `json({ defaultValue: [1,2,3,4,5] })` → `@default("[1,2,3,4,5]")` and `json({ defaultValue: [] })` → `@default("[]")` — Keystone's JSON-literal form: `JSON.stringify` the value (canonical, space-free) and wrap in escaped double quotes.

This is the foundational slice — the other field-default work builds on this shared helper.

## Acceptance criteria

- [ ] `formatPrismaDefault(value, fieldType)` is a standalone pure function with table-driven unit tests: string, non-empty string, empty string, integer, JSON array, JSON object, empty array, empty object.
- [ ] `integer()` with `defaultValue` emits `Int @default(N)` (and keeps `?` when nullable).
- [ ] `text()` with `defaultValue` emits `String @default("...")`.
- [ ] `json()` with `defaultValue` emits `Json? @default("<json-literal>")` using canonical space-free JSON serialisation wrapped in escaped quotes.
- [ ] Fields without a `defaultValue` emit no `@default(...)` (no regression in existing snapshots).
- [ ] Generator output tests assert generated schema strings for representative configs.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

Issue #470, opened 2026-06-03
PRD: Keystone-compatible generator defaults (field defaults, timestamps, singleton id, enum names)


borisno2:
## Problem Statement

Teams migrating a Keystone 6 project onto the OpenSaaS Stack need the generator's Prisma schema to diff clean against their existing database (**Schema parity**) so they can adopt the stack without a destructive migration. Today the `0.21.x` generator diverges from Keystone in several concrete ways, forcing every migrating project to reconcile the difference by hand with `db.extendPrismaSchema` string surgery:

- `text()`, `integer()`, and `json()` silently **drop `defaultValue`** — they never emit `@default(...)`, so any column with a live default diffs non-empty.
- The generator **unconditionally appends `createdAt`/`updatedAt`** to every model. Keystone adds neither automatically; many live tables (singletons, `Production`, etc.) have neither. On an existing DB this forces destructive `ADD COLUMN … NOT NULL`, and a list that declares its own `createdAt` ends up with a duplicate column (Prisma `P1012`).
- Singleton lists emit `id Int @id @default(1)`, but Keystone 6 emits a bare `id Int @id`, forcing a `SET DEFAULT 1` diff.
- An **optional** `select()` with a `defaultValue` generates as `NOT NULL` (the nullable `?` is overwritten), so migrating data containing NULLs fails.
- Native-enum `select()` derives the enum type name as `<List><Field>`, but the live Keystone enum often carries a `Type` suffix (e.g. `AccountNoteStatusType`); the enum's DB type name is structurally significant to `prisma migrate diff`.

These were discovered across a real Keystone→stack migration (≈31 lists). Each is reconcilable via `extendPrismaSchema` today, so none is a hard blocker — but every project repeats the same surgery.

## Solution

Bring the generator's default output in line with Keystone 6 conventions so a faithfully-translated config reaches **Schema parity** from config alone. Most changes are unconditional new defaults (acceptable to change directly in beta); the one convention a greenfield project would not want — empty-string text defaults — is gated behind **Keystone-compat mode**. See ADR-0004.

From the developer's perspective: translate your Keystone lists into `opensaas.config.ts`, optionally set `db: { keystoneCompat: true }`, run the generator, and `schema:diff` is clean — no per-field `extendPrismaSchema` workarounds.

## User Stories

1. As a migrating developer, I want `integer({ defaultValue: 3550 })` to emit `Int @default(3550)`, so that a column with a live integer default diffs clean.
[...]

PR #493, opened 2026-06-03
feat(core,cli): honour field defaultValue via formatPrismaDefault module

borisno2:
## Summary

Implements #471 (Part of #470). Adds a shared, pure `formatPrismaDefault(value, fieldType)` module that serialises a field's `defaultValue` into a Prisma `@default(...)` literal, and wires `text()`, `integer()`, and `json()` `getPrismaType` to use it. These three builders previously dropped `defaultValue` and emitted no `@default(...)`, forcing migrators to inject defaults via `extendPrismaSchema`. See ADR-0004 for the Keystone-compatibility rationale.

## Serialisation rules

- `integer({ defaultValue: 3550 })` → `@default(3550)` (bare numeric literal)
- `text({ defaultValue: 'PLEASE_UPDATE' })` → `@default("PLEASE_UPDATE")` (quoted string literal, escaped via `JSON.stringify`)
- `json({ defaultValue: [1,2,3,4,5] })` → `@default("[1,2,3,4,5]")` and `json({ defaultValue: [] })` → `@default("[]")` — Keystone's JSON-literal form

### How JSON serialisation is handled

The `json` case applies `JSON.stringify` twice: first to produce canonical, space-free JSON of the value (`[1,2,3,4,5]`), then a second `JSON.stringify` over that string to wrap it in escaped double quotes (`"[1,2,3,4,5]"`). This yields exactly Keystone's JSON-literal form and stays space-free regardless of value shape. Unserialisable values (e.g. a function, where the inner `JSON.stringify` returns `undefined`) emit no `@default(...)`.

### How nullability is preserved

`formatPrismaDefault` only ever returns the inner `@default(...)` literal (or `undefined`) — it never touches the `?` modifier. Each field builder appends the default to its `modifiers` string after the nullability/native-type modifiers, so the `?` is assembled independently of the default and is never overwritten. A required field still emits `Int @default(1)` (no `?`); an optional one emits `Int? @default(2)`. Fields with no `defaultValue` get `undefined` back and emit no `@default(...)`, so existing snapshots do not regress.

## Module design

`packages/core/src/fields/format-prisma-default.ts` is a standalone pure function (no I/O, no field-builder coupling). It is deliberately scoped to the three field types whose default literal maps through this shared helper; `checkbox`/`decimal`/`timestamp` keep formatting their own divergent literals inline. No field-type switch was added to the generator or UI — behaviour stays inside the field builders.

## Tests

- `packages/core/src/fields/format-prisma-default.test.ts` — table-driven unit tests: string, non-empty string, empty string, integer (incl. zero/negative), JSON array, JSON object, empty array, empty object, plus `undefined` → no default, canonical space-free JSON, and embedded-quote escaping.
- `packages/cli/src/generator/prisma.test.ts` — generator output tests asserting the produced schema strings: integer/text/json defaults, nullable `?` preserved independently of the default, no `@default` without a `defaultValue`, and a representative mixed-default model snapshot.

## Verification

- `pnpm build` (tsc typecheck) — pass for all packages
- `packages/core` tests — 547 passed
- `packages/cli` tests — 212 passed (no snapshot regressions)
- `pnpm manypkg fix` and `pnpm format` — clean

## Changeset

Minor bump for `@opensaas/stack-core` and `@opensaas/stack-cli` (new feature, beta — no major bump).

---

### DIFF
--- a/.changeset/silver-falcons-gather.md
+++ b/.changeset/silver-falcons-gather.md
@@ -1,0 +1,23 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Honour `defaultValue` for `text()`, `integer()`, and `json()` fields in the generated Prisma schema
+
+These three field builders previously dropped `defaultValue` and emitted no `@default(...)`. They now serialise the configured default into a Prisma `@default(...)` literal via a new shared, pure `formatPrismaDefault` module, matching Keystone 6 conventions. The nullable `?` modifier is preserved independently of the default, and fields without a `defaultValue` still emit no `@default(...)`.
+
+```typescript
+fields: {
+  // Int @default(3550)
+  quota: integer({ defaultValue: 3550 }),
+  // String @default("PLEASE_UPDATE")
+  status: text({ defaultValue: 'PLEASE_UPDATE' }),
+  // Json? @default("[1,2,3,4,5]") — Keystone's space-free JSON literal
+  limits: json({ defaultValue: [1, 2, 3, 4, 5] }),
+  // Json? @default("[]")
+  tags: json({ defaultValue: [] }),
+}
+```
+
+See ADR-0004 for the Keystone-compatibility rationale.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -33,3 +33,23 @@ _Avoid_: operation handler, mutation service
 **Hook Pipeline**:
 The module that runs the transform+validate span of a write — list `resolveInput` → field `resolveInput` → list `validate` → field `validate` → built-in field rules — owning that order and the `resolvedData` threading through it. It throws a validation error (never silent) when a validate hook reports via `addValidationError` or a built-in field rule fails, and returns the transformed `resolvedData` on success. The Write Pipeline delegates this span to it; side-effect hooks (`beforeOperation`/`afterOperation`), access, writable-field filtering, persistence and Field Visibility stay in the Write Pipeline.
 _Avoid_: validation service, input resolver
+
+### Migration & schema generation
+
+**Schema parity**:
+The property that the generator's Prisma schema diffs clean — an empty `prisma migrate diff` in both directions — against a pre-existing (typically Keystone-generated) database, so a project can adopt the stack without a destructive migration. It is the go/no-go gate for every migration phase.
+_Avoid_: schema match, clean diff, byte-compatibility
+
+**Keystone-compat mode**:
+An opt-in generator setting that makes schema output follow Keystone 6 conventions a migrating project depends on but a greenfield project would not want by default (e.g. non-null text columns defaulting to `""`). Conventions that every project wants are the plain default, not part of this mode.
+_Avoid_: legacy mode, compatibility flag, migration mode
+
+### Authentication
+
+**Auth lists**:
+The user/session/account/verification lists the auth plugin derives from the better-auth config — their keys, table/column maps, and database schema all follow that config, so they can be modelled to match (adopt) pre-existing better-auth tables. Distinct from any application domain User.
+_Avoid_: auth tables, auth models, auth schema
+
+**Auth identity**:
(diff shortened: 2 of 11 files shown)


borisno2 (commented):
## Code Review — PR #493 (feat: honour field `defaultValue` via `formatPrismaDefault`)

**Verdict: APPROVE** (submitting as a Comment review because the reviewing identity is the same account that authored the PR, so GitHub blocks a formal Approve.)

This is a clean, well-scoped, well-tested change. I verified the implementation end-to-end against the field builders and the generator's spacing logic; all the load-bearing claims in the description hold.

### What it does
Adds a pure `formatPrismaDefault(value, fieldType)` helper (`packages/core/src/fields/format-prisma-default.ts`) and wires `text()`, `integer()`, and `json()` `getPrismaType` to emit `@default(...)` from a configured `defaultValue`. Previously these three silently dropped the default, forcing migrators to inject it via `extendPrismaSchema`. Records the rationale in ADR-0004.

### Scrutiny against the high-risk areas — all pass

**1. JSON default serialisation (Keystone space-free escaped literal) ✅**
The double-`JSON.stringify` is correct. Inner stringify yields canonical space-free JSON; outer wraps it in an escaped double-quoted literal.
- `[1,2,3,4,5]` → `@default("[1,2,3,4,5]")` ✓
- `[]` → `@default("[]")`, `{}` → `@default("{}")` ✓ (covered by tests)
- Nested objects stay space-free (`{"nested":{"x":[1]}}`) ✓ (test guards against pretty-printing)
- Unserialisable values (e.g. a function) → inner `JSON.stringify` returns `undefined` → the explicit `if (serialised === undefined) return undefined` guard emits no `@default(...)` ✓
- Embedded quotes escape correctly (`say "hi"` → `"say \"hi\""`) ✓

**2. Nullable `?` preserved independently of the default ✅**
In all three builders the `?` is appended from `isNullable` first, then `@default(...)` is appended afterwards; the helper never touches the modifier string. Verified end-to-end against the generator (`prisma.ts` L153–158), which splits `?` off the front of the modifier string and re-attaches it to the type, routing `@default(...)` into the attribute part. Required → `Int @default(1)` (no `?`); optional → `Int? @default(2)`. Generator tests assert both shapes.

**3. Fields without a `defaultValue` emit no `@default(...)` ✅**
`formatPrismaDefault(undefined, …)` returns `undefined` before the switch, so the `if (defaultLiteral !== undefined)` guard skips the append. The "no @default" generator test and the unchanged existing snapshots confirm no regression.

**4. No `any` / no type-casting in public types; field self-containment ✅**
`value: unknown`, return `string | undefined`, exported `PrismaDefaultFieldType` is a plain string union — no `any`, no casts. The `switch` lives inside the pure helper called from each builder; no field-type switch was added to the generator or UI. Self-containment is preserved.

### Minor notes (non-blocking, no action required)
- `integer` uses `String(value)`. `IntegerField.defaultValue` is typed `number`, so this is fine for valid inputs; a non-integer number (e.g. `3.5`) would pass through verbatim into an `Int @default(...)`. Pre-existing type contract, not introduced here.
- `defaultValue: null` on a `json` field would emit `@default("null")` (passes the top `=== undefined` guard, `JSON.stringify(null)` → `"null"`). Defensible and an unusual config; worth a line in docs if you want to be explicit, but not worth gating on.
- Snapshot alignment of user fields vs. the hardcoded `id`/`createdAt`/`updatedAt` lines looks slightly ragged, but that's the pre-existing `padEnd(12)` + trim behaviour, not a regression from this PR. Prisma format would normalise it on write anyway.

### Process / conventions
- Changeset present (`minor` for both `@opensaas/stack-core` and `@opensaas/stack-cli`) — correct for a new feature in beta.
- Tests are thorough: table-driven unit tests for the helper plus generator-output assertions covering integer/text/json, nullable independence, the no-default case, and a mixed-default snapshot.
- ESM `.js` import extension used correctly.

Nice work — ship it once CI is green (PR currently shows `mergeable_state: blocked`, presumably pending checks/branch protection rather than anything in this diff).


---

Issue #481, opened 2026-06-03
authPlugin mirrors better-auth config: derive auth-list keys + field maps, resolve runtime key, don't overwrite app User


borisno2:
## Parent

#480

## What to build

Make `authPlugin` derive its **Auth lists** from the better-auth config the developer already writes, instead of hardcoding `User`/`Session`/`Account`/`Verification`. End-to-end slice:

- New pure **`better-auth config → Auth lists` derivation** module: maps better-auth per-model `modelName` → list key + table `@@map`, and `fields` (+ additional fields) → field names/`@map`. `getAuthLists`/`convertBetterAuthSchema` consume it.
- `init()` add-vs-extend logic uses the derived keys: when the derived user key (e.g. `AuthUser`) differs from an app's own existing list (e.g. `public.User`), the plugin **adds** its own list and does NOT extend/overwrite the app list.
- Runtime `getUser`/`getCurrentUser` resolve the user list key from the configured user model (replacing the hardcoded `userListKey = 'user'` TODO).

## Acceptance criteria

- [ ] `better-auth config → Auth lists` derivation is a standalone pure module with unit tests: custom `modelName`/`fields` → expected list keys, `@@map`, and field `@map`.
- [ ] Configuring better-auth with `modelName: 'AuthUser'` (etc.) generates auth lists keyed `AuthUser`/`AuthSession`/`AuthAccount`/`AuthVerification` with matching table/column maps.
- [ ] Default behaviour (no `modelName` overrides) is unchanged (`User`/`Session`/`Account`/`Verification`).
- [ ] Adding the plugin with custom keys does NOT extend or overwrite an app's own separate `User` list (no-overwrite test).
- [ ] `getUser`/`getCurrentUser` use the configured user model key, not a hardcoded `'user'`.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

Issue #480, opened 2026-06-03
PRD: authPlugin adopts existing better-auth tables (mirror better-auth config + relocatable schema)


borisno2:
## Problem Statement

A project migrating to the stack already has a working, hand-wired better-auth installation: its auth tables are `AuthUser`/`AuthSession`/`AuthAccount`/`AuthVerification`, mapped into a separate `auth` Postgres schema (`@@schema("auth")`), and its application `User` (`public.User`, keyed by `subjectId`) is a *different* model from the better-auth user. `authPlugin().init()` hardcodes the four list keys `User`/`Session`/`Account`/`Verification` in the `public` schema and, for each, either `extendList` (if the key exists) or `addList`. Wiring it into the generated config would therefore (a) `extendList('User', …)` — overwriting the app's own `public.User` — and (b) add three new `public` tables. Both diverge from the live database: `schema:diff` flips to NO-GO and a **destructive auth migration** is required. There is no knob to use different list keys, target a non-`public` schema, or tell the plugin the tables already exist. This is the key blocker that prevents adding the plugin to the generated config.

## Solution

Make `authPlugin` a thin wrapper over better-auth (see ADR-0007): the auth lists are **derived from the better-auth config the developer already writes** — list keys from better-auth's per-model `modelName`, field/column maps from its `fields` — and can be placed in a non-`public` database schema via a plugin-level `schema` option. With keys, schema, and field shapes matching the live tables, the generated **Auth lists** diff clean against the existing database — they are modelled for runtime/types without producing a migration. The app's own `User` is left untouched; the application links its domain User to the **Auth identity** itself.

## User Stories

1. As a migrating developer with existing `AuthUser`/`AuthSession`/`AuthAccount`/`AuthVerification` tables, I want `authPlugin` to derive its list keys from better-auth's `modelName` config, so that the generated lists carry the same keys/table names as my live tables.
2. As a migrating developer, I want the auth lists' field and column names derived from better-auth's `fields` config, so that columns match my live schema without manual remapping.
3. As a migrating developer, I want a plugin-level `schema` option that places all generated auth lists in a non-`public` Postgres schema (e.g. `auth`) with the right `@@schema`/`@@map`, so that they match my separate-schema layout.
4. As a migrating developer, I want a per-list schema override, so that an individual auth list can live in a different schema if needed.
[...]

PR #497, opened 2026-06-03
feat(auth): derive Auth lists from better-auth config

borisno2:
Implements #481

Part of #480

## What this does

Makes `authPlugin` derive its Auth lists (user/session/account/verification) from the better-auth config the developer writes, instead of hardcoding the keys `User`/`Session`/`Account`/`Verification`.

### New pure derivation module
`packages/auth/src/config/derive-auth-lists.ts` (`deriveAuthLists`) maps the resolved better-auth model config to the four OpenSaaS Auth lists:
- per-model `modelName` → list key + table `@@map`
- per-model `fields` (better-auth field → column) → field-level `@map`
- the `userId` column override → the `user` relationship foreign-key `@map`
- relationship refs between the Auth lists follow the derived keys (e.g. `Session.user → AuthUser.sessions`)

`getAuthLists`/`createUserList`/... and the plugin's add-vs-extend logic all consume this single module.

### Add-vs-extend uses derived keys (no-overwrite guarantee)
The plugin only ever adds/extends its **own derived keys**. When the user model is renamed (e.g. `user.modelName: 'AuthUser'`), the plugin adds `AuthUser` and leaves an app's separate domain `User` completely untouched — it never extends a list it didn't derive. Extending only kicks in when an existing list shares the derived key (the default `User`), which is the intended "merge auth fields into my User" behaviour.

### Runtime key resolution
`getUser`/`getCurrentUser` resolve the user list's `context.db` key from the configured user model (`getDbKey(modelName)` → `authUser`), replacing the hardcoded `userListKey = 'user' // TODO`.

### Supporting changes
- `createAuth()` mirrors the per-model config (`modelName` + `fields`) back to the running better-auth instance so it reads/writes the same tables/columns the Auth lists were derived from.
- core `ListConfig` gains a model-level `db.map`; the Prisma generator emits `@@map("...")` for it.

## How default behaviour is preserved

When no `modelName`/`fields` overrides are given, the derivation produces the historical default set keyed `User`/`Session`/`Account`/`Verification`, original field shapes, no `@@map`, and the existing FK `@map("user")` convention. Verified end-to-end: regenerating the `examples/auth-demo` Prisma schema produces **zero diff**.

## Usage (adoption)

```typescript
authPlugin({
  user: { modelName: 'AuthUser', fields: { name: 'full_name' } },
  session: { modelName: 'AuthSession', fields: { userId: 'user_id' } },
  account: { modelName: 'AuthAccount' },
  verification: { modelName: 'AuthVerification' },
})
// -> lists keyed AuthUser/AuthSession/AuthAccount/AuthVerification
//    with @@map + column @map matching the live tables; app's own User untouched
```

## Tests

- **Derivation unit tests** (`tests/derive-auth-lists.test.ts`): custom `modelName`/`fields` → expected keys, `@@map`, field `@map`, FK `@map`, relationship refs; default-unchanged assertions.
- **No-overwrite + add-vs-extend** (`tests/plugin-derived-keys.test.ts`): custom keys do NOT extend an app's own `User`; default keys still merge auth fields.
- **Runtime key** (`tests/plugin-derived-keys.test.ts`): `getUser`/`getCurrentUser` use the configured user model db key.
- **Generator `@@map`** (`packages/cli/src/generator/prisma.test.ts`): list-level `db.map` emits `@@map`.

## Verification

- `pnpm build` — all 11 packages pass (tsc)
- `packages/auth`: 117 tests pass (was 100)
- `packages/cli`: 206 tests pass
- `packages/core`: 532 tests pass
- `pnpm lint` — 0 errors; `pnpm manypkg fix` / `pnpm format` clean

Changeset added (minor; auth/core/cli).

---

### DIFF
--- a/.changeset/auth-derive-from-better-auth.md
+++ b/.changeset/auth-derive-from-better-auth.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-auth': minor
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Derive the auth plugin's Auth lists from the better-auth config
+
+`authPlugin` now mirrors the better-auth config a developer writes instead of hardcoding the keys `User`/`Session`/`Account`/`Verification`. Per-model `modelName` becomes the OpenSaaS list key (and a table `@@map`), and the `fields` column map becomes per-field `@map`s. The plugin only ever adds/extends its own derived keys, so an app's separate domain `User` is never overwritten. The runtime `getUser`/`getCurrentUser` helpers now resolve the user list key from the configured user model instead of a hardcoded `'user'`.
+
+Default behaviour (no overrides) is unchanged: the lists are still keyed `User`/`Session`/`Account`/`Verification` with the original field shapes and no `@@map`.
+
+```typescript
+// Adopt existing better-auth tables without a destructive migration
+authPlugin({
+  user: { modelName: 'AuthUser', fields: { name: 'full_name' } },
+  session: { modelName: 'AuthSession', fields: { userId: 'user_id' } },
+  account: { modelName: 'AuthAccount' },
+  verification: { modelName: 'AuthVerification' },
+})
+// -> lists keyed AuthUser/AuthSession/AuthAccount/AuthVerification
+//    with @@map + column @map matching the live tables
+```
+
+Lists also gain a model-level `db.map` option, which emits a `@@map("...")` on the generated Prisma model so a list key can differ from its physical table name.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -33,3 +33,23 @@ _Avoid_: operation handler, mutation service
 **Hook Pipeline**:
 The module that runs the transform+validate span of a write — list `resolveInput` → field `resolveInput` → list `validate` → field `validate` → built-in field rules — owning that order and the `resolvedData` threading through it. It throws a validation error (never silent) when a validate hook reports via `addValidationError` or a built-in field rule fails, and returns the transformed `resolvedData` on success. The Write Pipeline delegates this span to it; side-effect hooks (`beforeOperation`/`afterOperation`), access, writable-field filtering, persistence and Field Visibility stay in the Write Pipeline.
 _Avoid_: validation service, input resolver
+
+### Migration & schema generation
+
+**Schema parity**:
+The property that the generator's Prisma schema diffs clean — an empty `prisma migrate diff` in both directions — against a pre-existing (typically Keystone-generated) database, so a project can adopt the stack without a destructive migration. It is the go/no-go gate for every migration phase.
+_Avoid_: schema match, clean diff, byte-compatibility
+
+**Keystone-compat mode**:
+An opt-in generator setting that makes schema output follow Keystone 6 conventions a migrating project depends on but a greenfield project would not want by default (e.g. non-null text columns defaulting to `""`). Conventions that every project wants are the plain default, not part of this mode.
+_Avoid_: legacy mode, compatibility flag, migration mode
+
+### Authentication
+
+**Auth lists**:
+The user/session/account/verification lists the auth plugin derives from the better-auth config — their keys, table/column maps, and database schema all follow that config, so they can be modelled to match (adopt) pre-existing better-auth tables. Distinct from any application domain User.
+_Avoid_: auth tables, auth models, auth schema
+
+**Auth identity**:
+The better-auth-owned record of who a session belongs to (the better-auth user). Separate from, and not assumed to be, the application's own domain User; an app links the two itself when it needs to.
+_Avoid_: auth user, principal, account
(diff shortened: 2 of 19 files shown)


borisno2 (commented):
## Review — `feat(auth): derive Auth lists from better-auth config`

**Verdict: APPROVE** (posted as a Comment review because the GitHub integration identity authored the PR, so a formal Approve is blocked — treat this as an approval).

Reviewed strictly from the PR's 19-file 3-dot diff against the merge-base (1035+/223-), cross-checked against the live field builder and generator in the repo. The design is clean: a single pure derivation module (`deriveAuthLists`) now feeds the legacy helpers, the plugin add-vs-extend logic, and runtime key resolution. I verified every claim in the four scrutiny areas below against the actual code rather than the description.

### Default-behaviour preservation — VERIFIED (no regression)
- `listDb()` returns `undefined` unless `modelName !== default`, so with no overrides no `db.map` is set → generator emits **no `@@map`**. Confirmed by the generator change (`packages/cli/src/generator/prisma.ts:189-194`, only emits when `listConfig.db?.map` is truthy) and the negative test.
- `fieldDb()` returns `undefined` when a field has no column override, so every default field is built byte-identically to the old hand-written lists (e.g. `name: text({ validation: { isRequired: true }, db: undefined })` ≡ old `name: text({ validation: { isRequired: true } })`).
- The existing FK `@map("user")` is preserved correctly. It does **not** come from the auth package — it is emitted unconditionally by the relationship field builder's default branch (`packages/core/src/fields/index.ts:1135-1138`: `@map("${fieldName}")`). `userForeignKeyDb()` returns `undefined` for the default (no `fields.userId`), leaving `field.db` undefined and falling through to that default `@map("user")`. Cross-checked against the committed `examples/auth-demo/prisma/schema.prisma` which still has `userId String? @map("user")` and no `@@map`. The PR's "zero diff on regenerate" claim is sound.
- Keys remain `User`/`Session`/`Account`/`Verification`; relationship refs default to `User.sessions`/`User.accounts`/`Session.user`/`Account.user`. Asserted by `derive-auth-lists.test.ts`.

### No-overwrite guarantee (renamed user model) — VERIFIED
- The add-vs-extend loop only ever iterates the plugin's **own derived keys** (`Object.entries(authLists)`), and extends only when `context.config.lists[listName]` already exists under that derived key. When `user.modelName: 'AuthUser'`, the derived key is `AuthUser`, so an app's separate `User` is never touched. `plugin-derived-keys.test.ts` proves the app `User` keeps `subjectId`, gains no `email/sessions`, and retains its own `beforeOperation` hook. Default case (`User`) still merges — the intended behaviour.

### Runtime key resolution — VERIFIED
- `getUser`/`getCurrentUser` now use `context.db[getDbKey(normalized.models.user.modelName)]`, replacing the hardcoded `'user'` TODO. `getDbKey('AuthUser')` → `authUser`, `getDbKey('User')` → `user` (confirmed via `pascalToCamel`). Tested via the Proxy `accessedKeys` harness.

### Core change (`ListConfig.db.map` + generator `@@map`) — SOUND, no non-auth regression
- New `ListConfig['db']` is additive and optional; it did not previously exist (the other `db?` blocks at lines 364/627 are field/relationship level). Generator gates the `@@map` emission on `listConfig.db?.map`, so non-auth models with no `db.map` are unaffected — covered by the "should not generate @@map" test.
- One-to-one safety: the auth `user` relationships set `db.foreignKey` only as the **object** form `{ map }`, never `=== true`. `shouldHaveForeignKey`/`isOneToOneRelationship` only branch on `=== true`, and the many-to-one classification is driven by the target side (`AuthUser.sessions` is `many:true`), so no spurious `@unique` is added when adopting a `userId` column rename. The relationship validation (`fields/index.ts:1185-1202`) accepts the object form on a bidirectional single rel. Correct.

### `any`/type-casting in public types — ACCEPTABLE
- No `any` leaks into a new public type. `DerivedAuthLists.lists` is `Record<string, ListConfig<any>>`, but that `ListConfig<any>` pattern is the pre-existing project convention for the auth lists (the eslint-disable comment "ListConfig must accept any TypeInfo" matches the prior `lists/index.ts`), not a new escape hatch.
- The `as { userId?: string }` / `as Record<string, unknown>` casts inside the access functions are copied verbatim from the previous hand-written lists — internal to function bodies, not exposed type surface. `AuthModelConfig`, `NormalizedAuthModelConfig`, `NormalizedAuthModels`, `DerivedAuthLists` are all fully concrete.

### Minor / non-blocking
1. **`toBetterAuthModelOptions` always sets `modelName`** (`server/index.ts`): since `NormalizedAuthModelConfig.modelName` is always concrete (defaulted), the `if (model.modelName)` guard is always true, so the function returns `undefined` only when there are no field overrides AND... actually it never returns `undefined` for user/account/verification because `modelName` is always present. This means the running better-auth instance is now always handed an explicit `modelName: 'User'` etc. even in the default case. That should be a no-op (it equals better-auth's own default), but it's a behavioural change vs. previously passing nothing. Worth a one-line confirmation that better-auth treats an explicit default `modelName` identically to omitting it — otherwise harmless.
2. **Doc example mismatch** (`core/src/config/types.ts` `db.map` JSDoc): the example shows `db: { map: 'user' }` producing `@@map("user")` for a list named `AuthUser` — fine as a generic example, but the actual auth derivation pins `@@map` to the *modelName itself* (`@@map("AuthUser")`), not to a lowercase physical name. Slightly confusing next to the auth use case; consider aligning the example.
3. **Changeset/version**: `minor` for auth/core/cli is appropriate (additive feature, default behaviour preserved). Good.

### Tests
Coverage is strong and targets the right invariants: default-unchanged (keys, field shapes, no `@@map`/`@map`, FK untouched), custom `modelName`/`fields`/`userId` mapping, no-overwrite of app `User`, default-key merge, and runtime db-key resolution incl. the no-session null path. The generator gets both positive and negative `@@map` cases.

Nothing blocking. Ship it once the two minor doc nits are considered (optional).

---

Issue #475, opened 2026-06-03
Keystone-compat mode: non-null text without explicit default emits @default("")


borisno2:
## Parent

#470

## What to build

Add an opt-in `db: { keystoneCompat: true }` setting (the **Keystone-compat mode** defined in `CONTEXT.md`). When enabled, every non-null text column that has no explicit `defaultValue` emits `@default("")` — Keystone's implicit empty-string default — so a migrating project doesn't have to hand-set `defaultValue: ''` on dozens of columns. When disabled (the default), no implicit text default is emitted, keeping greenfield schemas clean.

The flag must NOT affect nullable text, fields with an explicit `defaultValue`, or any non-text field.

## Acceptance criteria

- [ ] `db.keystoneCompat` is added to the config types and documented.
- [ ] With `keystoneCompat: true`, a required/non-null `text()` with no `defaultValue` generates `String @default("")`.
- [ ] With `keystoneCompat` off (default), the same field generates `String` with no default.
- [ ] An explicit `text({ defaultValue: 'x' })` still wins over the compat empty-string default.
- [ ] Nullable text and non-text fields are unaffected by the flag.
- [ ] A focused Keystone-compat snapshot test proves empty-string text defaults across a representative model.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

- #471

PR #499, opened 2026-06-03
feat(core,cli): opt-in db.keystoneCompat empty-string text defaults

borisno2:
Implements #475
Part of #470

## What changed

Adds an opt-in `db: { keystoneCompat: true }` setting (the **Keystone-compat mode**). When enabled, every non-null `text()` column that has **no** explicit `defaultValue` emits `String @default("")` — Keystone 6's implicit empty-string text default — so a migrating project reaches **Schema parity** without hand-setting `defaultValue: ''` on dozens of columns.

When off (the default), no implicit text default is emitted, keeping greenfield schemas clean.

The flag:
- only affects **non-null** `text()` **without** an explicit `defaultValue`
- never affects nullable text, fields with an explicit `defaultValue` (which always wins, including an explicit `''`), or any non-text field

## Implementation notes

- `db.keystoneCompat?: boolean` added to `DatabaseConfig` with JSDoc referencing ADR-0004.
- Threaded to `text()`'s `getPrismaType` as a 4th positional argument — the same way `provider`/`listName` already reach fields — via the generator in `packages/cli/src/generator/prisma.ts`. `getPrismaType`'s type signature gains the optional `keystoneCompat` param.
- Builds on the existing `text()` default-emission path: the empty-string literal is produced through the existing pure `formatPrismaDefault` module (no new default-emission code).
- No field-type switch statements; field stays self-contained.

## Tests

- **Keystone-compat generator snapshot** in `packages/cli/src/generator/prisma.test.ts` proving empty-string text defaults across a representative model (required text → `@default("")`, explicit default wins, nullable text untouched, non-text untouched), plus an off-by-default contrast assertion.
- **Unit coverage** in `packages/core/src/fields/text-keystone-compat.test.ts` of the on / off / explicit-default / nullable / non-text matrix on `text().getPrismaType`.

## Verification

- `pnpm build` — pass
- `packages/core`: 558 tests pass
- `packages/cli`: 214 tests pass (1 new snapshot)
- `pnpm lint` — 0 errors (only pre-existing warnings)
- `pnpm manypkg fix` / `pnpm format` — clean

Changeset added (minor, beta) for `@opensaas/stack-core` and `@opensaas/stack-cli`.

---

### DIFF
--- a/.changeset/proud-otters-compat.md
+++ b/.changeset/proud-otters-compat.md
@@ -1,0 +1,36 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Add opt-in `db.keystoneCompat` mode for Keystone-compatible empty-string text defaults
+
+When migrating from Keystone 6, every non-null text column carries an implicit empty-string default. Set `db: { keystoneCompat: true }` to mirror that: any non-null `text()` column without an explicit `defaultValue` now generates `String @default("")`, so a migrating schema reaches parity without hand-setting `defaultValue: ''` on dozens of columns.
+
+The mode is off by default (greenfield schemas stay clean) and never affects nullable text, fields with an explicit `defaultValue`, or any non-text field — an explicit `text({ defaultValue: 'x' })` always wins.
+
+```typescript
+export default config({
+  db: {
+    provider: 'postgresql',
+    keystoneCompat: true, // non-null text without a default → @default("")
+    prismaClientConstructor: (PrismaClient) => {
+      // ... adapter setup
+    },
+  },
+  lists: {
+    Account: list({
+      fields: {
+        // required text → String @default("")
+        name: text({ validation: { isRequired: true } }),
+        // explicit default still wins → String @default("PLEASE_UPDATE")
+        status: text({ validation: { isRequired: true }, defaultValue: 'PLEASE_UPDATE' }),
+        // nullable text is untouched → String?
+        bio: text(),
+      },
+    }),
+  },
+})
+```
+
+See ADR-0004 for the full Keystone-compatible generator defaults.
--- a/packages/cli/src/generator/__snapshots__/prisma.test.ts.snap
+++ b/packages/cli/src/generator/__snapshots__/prisma.test.ts.snap
@@ -1,5 +1,29 @@
 // Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
 
+exports[`Prisma Schema Generator > Keystone-compat mode (db.keystoneCompat) > emits @default("") for non-null text without an explicit default (snapshot) 1`] = `
+"generator client {
+  provider = "prisma-client"
+  output   = "../.opensaas/prisma-client"
+}
+
+datasource db {
+  provider = "postgresql"
+}
+
+model Account {
+  id        String   @id @default(cuid())
+  name         String @default("")
+  status       String @default("PLEASE_UPDATE")
+  nickname     String?
+  note         String?
+  loginCount   Int
+  isActive     Boolean @default(true)
+  createdAt DateTime @default(now())
+  updatedAt DateTime @default(now()) @updatedAt
+}
+"
+`;
+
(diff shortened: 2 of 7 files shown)


borisno2 (commented):
## Code review — `feat(core,cli): opt-in db.keystoneCompat empty-string text defaults`

**Verdict: APPROVE** (submitted as a Comment review because GitHub blocks a formal Approve when the review identity matches the PR author. Treat this as an approval.)

Reviewed strictly from the PR diff plus the files it touches on `refs/pull/499/head` (`packages/core/src/fields/index.ts`, `format-prisma-default.ts`, `config/types.ts`, `packages/cli/src/generator/prisma.ts`, and the two test files). The implementation is correct, well-scoped, and the test matrix matches the stated acceptance criteria. Findings below, ordered by severity.

### Correctness — the four guarantees all hold

1. **Off by default — confirmed.** `const keystoneCompat = config.db.keystoneCompat ?? false` in `generatePrismaSchema`. With it unset, `text()`'s `defaultSource` is `keystoneCompat && !isNullable ? '' : undefined` → `undefined`, so `formatPrismaDefault(undefined, 'text')` returns `undefined` and no `@default` is emitted. The diff adds a *new* snapshot block only; the existing `field defaultValue → @default(...)` snapshot is untouched, and the explicit off-by-default test asserts `not.toContain('@default("")')`. Existing snapshots are genuinely unchanged.

2. **Empty-string scoped correctly — confirmed.** The ternary checks `options?.defaultValue !== undefined` first, so an explicit default always wins — including an explicit `defaultValue: ''`, which short-circuits to the explicit value (no double-emission; the "honours an explicit empty-string defaultValue" test pins exactly one `@default`). Nullable text (`isNullable` true, the `?` branch) never gets the compat default. Non-text fields don't thread the flag into their own default logic at all. Unit tests cover on/off/explicit/explicit-''/nullable/db.isNullable/non-text.

3. **4th positional param is non-breaking — confirmed.** `keystoneCompat?: boolean` is optional in the `BaseFieldConfig.getPrismaType` signature, and the generator passes it uniformly to every field. Other field builders (`integer`, `select`, `calendarDay`, `decimal`, `checkbox`, `timestamp`, `json`, `password`) simply ignore the extra argument — their narrower parameter lists remain valid under an optional-trailing-param signature. No caller or signature breaks.

4. **No `any` / no type-casting in the new public surface — confirmed.** `db.keystoneCompat?: boolean` is clean and documented (JSDoc + `@default false` + ADR-0004 reference). `formatPrismaDefault(value: unknown, …)` uses `unknown` only internally within the package, which is permitted by the repo's typing rules. The pre-existing `as any` in `password()` is not introduced by this PR.

### Quality

- The `text()` modifier ordering (`@db.<native>` → `@default` → `@unique`/`@index` → `@map`) is asserted explicitly by a unit test, which is a good regression guard for a builder that concatenates modifier strings.
- Reusing the existing pure `formatPrismaDefault('', 'text')` (→ `JSON.stringify(String('')) === '""'`) to produce the empty-string literal is the right call — no new default-emission code, and quote-escaping is handled the same way as any other text default.
- Changeset present (minor for `@opensaas/stack-core` + `@opensaas/stack-cli`) with a clear migration example. Doc/JSDoc quality is high.

### Non-blocking notes

- **CI `e2e` red is unrelated.** It fails in `create-opensaas-app`'s `scaffold-first-run-guard` with `Template basic not found` (templates not generated before the guard ran) — a build-ordering/infra issue. This PR touches nothing under `create-opensaas-app` or `templates/`. The `test` job (which runs the new core + CLI suites that actually exercise this change) was still in progress at review time and should be allowed to go green before merge.
- **Cosmetic snapshot alignment.** In the new snapshot, compat-affected columns render with slightly different inter-token spacing than neighbours (e.g. `name         String @default("")`). This is the generator's existing `padEnd(12)`-plus-modifier-spacing behaviour, not a defect of this change — just flagging it so the snapshot isn't mistaken for a formatting bug.

Nice, tightly-scoped feature with a thorough on/off/edge test matrix. No required changes.

---

PR #500, opened 2026-06-03
Re-export query API (defineFragment / runQuery / runQueryOne / ResultOf) from @opensaas/stack-core root

### LINKED ISSUE #496: Re-export query API (defineFragment / runQuery / runQueryOne / ResultOf) from @opensaas/stack-core root
## Context

Discovered during review of #492 (part of PRD #486). The migration guide, `specs/keystone-migration.md`, the `@opensaas/stack-core` CHANGELOG, and the `migrate-context-calls` skill all instruct users to:

```ts
import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'
```

But the package's root `index.ts` does **not** re-export `./query/index.js`. `extend.ts` doesn't export them either, and `internal.ts` exports only the *types* `Fragment`/`FieldSelection`. So the documented import path does not resolve — the runtime functions `defineFragment`/`runQuery`/`runQueryOne` are unreachable from the public root entry point, and `ResultOf` likewise.

This is pre-existing (not introduced by #492), but it breaks the core of the documented `context.graphql.run` → fragments migration story.

## What to build

Make the documented root import resolve: re-export the public query surface (`defineFragment`, `runQuery`, `runQueryOne`, and the `ResultOf` type — plus any other intended-public query exports) from `@opensaas/stack-core`'s root entry point. Keep `unknown`/`any` out of the public types.

## Acceptance criteria

- [ ] `import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'` resolves and type-checks.
- [ ] A test (or type-level test) asserts the root exports are present, so this can't silently regress.
- [ ] The export surface matches what the docs/CHANGELOG/skill advertise (no more, no less than the intended public API).
- [ ] Changeset added (see `pr-changeset` skill) — patch (bug fix).

## Blocked by

None - can start immediately

borisno2:
Fixes #496

## Problem

The migration guide (`specs/keystone-migration.md`), the `@opensaas/stack-core` CHANGELOG, the query module's JSDoc, and the `migrate-context-calls` skill all instruct users to:

```ts
import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'
```

But the package root (`packages/core/src/index.ts`) did not re-export `./query/index.js`, so that import did not resolve. The runtime functions and the `ResultOf` type were unreachable from the public root entry point.

## Change

Re-export the intended public query surface from the root entry point:

- Runtime: `defineFragment`, `runQuery`, `runQueryOne`
- Types: `ResultOf`, `RelationSelector`, `QueryArgs`

These are exactly the symbols advertised by the docs / CHANGELOG / skill. Internal-only helpers (`isFragment`, `buildInclude`, `pickFields`, marked `@internal`) and the `Fragment` / `FieldSelection` types remain off the root surface (the latter stay on `@opensaas/stack-core/internal`, where they already live). No `any` / type-casts were added to public types, and there are no name collisions with existing root exports.

## Regression guard

Added `packages/core/src/index.test.ts` — imports the four documented symbols from the package root and asserts both at runtime (functions are present, `defineFragment` produces a usable fragment) and at the type level (`ResultOf` narrows correctly; `QueryArgs` / `RelationSelector` resolve). If the root stops re-exporting the query API, this fails to compile / run.

## Verification

- `pnpm build` (incl. `tsc` typecheck of core) — pass
- `cd packages/core && pnpm test` — 551 passed (27 files), including the new 4-test file
- `pnpm lint` — no new warnings/errors on changed files
- `pnpm manypkg fix` / `pnpm format` — clean
- Changeset added (`@opensaas/stack-core`: patch)

---

### DIFF
--- a/.changeset/calm-otters-gather.md
+++ b/.changeset/calm-otters-gather.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Re-export the fragment query API (`defineFragment`, `runQuery`, `runQueryOne`, and the `ResultOf`, `RelationSelector`, `QueryArgs` types) from the package root so the documented `import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'` resolves.
--- a/packages/core/src/index.test.ts
+++ b/packages/core/src/index.test.ts
@@ -1,0 +1,50 @@
+import { describe, it, expect, expectTypeOf } from 'vitest'
+// Import from the package root entry point exactly as the docs / CHANGELOG /
+// migrate-context-calls skill instruct consumers to. If the root `index.ts`
+// stops re-exporting the query API, this file fails to type-check / run,
+// preventing a silent regression (issue #496).
+import { defineFragment, runQuery, runQueryOne } from './index.js'
+import type { ResultOf, RelationSelector, QueryArgs } from './index.js'
+
+// ─────────────────────────────────────────────────────────────
+// Stand-in model type (mirrors a Prisma-generated type)
+// ─────────────────────────────────────────────────────────────
+
+type User = {
+  id: string
+  name: string
+  email: string
+}
+
+describe('@opensaas/stack-core root entry point — query API re-exports (issue #496)', () => {
+  it('re-exports the runtime query functions from the package root', () => {
+    expect(typeof defineFragment).toBe('function')
+    expect(typeof runQuery).toBe('function')
+    expect(typeof runQueryOne).toBe('function')
+  })
+
+  it('defineFragment imported from the root produces a usable fragment', () => {
+    const userFragment = defineFragment<User>()({ id: true, name: true } as const)
+
+    expect(userFragment._type).toBe('fragment')
+    expect(userFragment._fields).toEqual({ id: true, name: true })
+  })
+
+  it('exposes ResultOf as a usable type alias from the root', () => {
+    const userFragment = defineFragment<User>()({ id: true, name: true } as const)
+    expect(userFragment._type).toBe('fragment')
+
+    // Type-level assertion: ResultOf<typeof fragment> narrows to the selection.
+    expectTypeOf<ResultOf<typeof userFragment>>().toEqualTypeOf<{ id: string; name: string }>()
+  })
+
+  it('exposes QueryArgs and RelationSelector as usable types from the root', () => {
+    // Type-level usage — these annotations only compile if the types resolve
+    // from the root entry point.
+    const args: QueryArgs = { where: { id: 'abc' }, take: 5 }
+    expect(args.take).toBe(5)
+
+    const selector: RelationSelector<User> = defineFragment<User>()({ id: true } as const)
+    expectTypeOf(selector).not.toBeNever()
+  })
+})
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Code Review — PR #500

**Verdict: APPROVE** (submitted as a Comment review because the reviewer identity is the same as the PR author, so GitHub blocks a formal Approve — the intent is approval.)

### Overview
Re-exports the fragment query API from the `@opensaas/stack-core` root so the documented `import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'` resolves (fixes #496). Adds a regression-guard test plus a `@opensaas/stack-core` patch changeset. Touches 3 files, +64/-0. Source `query/index.ts` is unchanged — purely a re-export surface change.

### Scrutiny checklist — all pass

**1. Only intended public symbols re-exported; no internal leakage — CONFIRMED**
Root re-exports exactly the advertised set: runtime `defineFragment` / `runQuery` / `runQueryOne` and types `ResultOf` / `RelationSelector` / `QueryArgs`. The `@internal`-marked helpers (`isFragment`, `buildInclude`, `pickFields`) and the `Fragment` / `FieldSelection` types are **not** re-exported from root. `Fragment` / `FieldSelection` correctly remain on `@opensaas/stack-core/internal` (verified in `src/internal.ts`), matching the inline comment. `QueryRunnerContext` is also (correctly) kept off the root surface.

**2. No export-name collision — CONFIRMED**
None of the six new names collide with the 19 pre-existing root exports (`config`, `list`, `getContext`, `getDbKey`, `ValidationError`, etc.). `index.ts` uses only explicit named exports — no `export *` barrel that could shadow or duplicate. `RelationSelector` / `QueryArgs` live only on root (not also on `/internal`), so no cross-entry duplication concern either.

**3. Regression guard actually fails if the re-export is removed — CONFIRMED (with one nuance)**
- Runtime: removing the runtime re-export makes `defineFragment`/`runQuery`/`runQueryOne` import as `undefined`, so the `typeof … === 'function'` assertions fail; it would also be a hard `tsc` "no exported member" error.
- Types: removing the type re-export makes the `import type { ResultOf, RelationSelector, QueryArgs }` line a `tsc` compile error.
- Nuance (non-blocking): the `expectTypeOf` type-level assertions are erased at runtime and are only enforced under `tsc` / `vitest --typecheck`, not a bare `vitest run`. That's fine here because core's `build` runs `tsc`, and the import bindings break regardless — but the type-level *narrowing* assertions specifically rely on the build's typecheck step rather than the test runner. Worth a one-line awareness note; no change required.

**4. No `any` / type-casting added to public types — CONFIRMED**
The diff adds no `any` and no unsafe casts. `as const` in the test is a const assertion (standard and required for fragment-literal inference), not a prohibited cast. Public type signatures are untouched.

### Other checks
- Changeset present and correctly scoped (`@opensaas/stack-core`: patch). The changeset-bot's "9 packages" list is the normal dependent-bump fan-out, not a problem.
- JSDoc on the new export block is accurate and points users at the correct entry points.
- No behavioural/runtime risk: this only widens the public surface to match existing docs; it removes a documentation-vs-reality drift rather than introducing new API.

### Suggestions (optional, non-blocking)
- Consider relying on a `tsd`/`vitest typecheck` config (or a brief code comment) to make explicit that the type-level guard is enforced by the build's `tsc` pass, so a future change to the test command can't silently weaken the type guarantee.

Clean, minimal, well-tested fix. Approving.

<sub>Reviewed by Claude Code</sub>

---

Issue #472, opened 2026-06-03
Auto-timestamps off by default: db.timestamps opt-in + per-list override + skip-when-declared


borisno2:
## Parent

#470

## What to build

Stop unconditionally appending `createdAt`/`updatedAt` to every generated model. Auto-timestamps become OFF by default (acceptable breaking change in beta — see ADR-0004). Re-enable globally with `db: { timestamps: true }`, with a per-list override. When timestamps are enabled and a list already declares its own `createdAt` and/or `updatedAt`, the auto pair must be skipped for the declared field(s) so Prisma never sees a duplicate (`P1012`).

Extract the decision into a small, testable predicate (e.g. `resolveListTimestamps(listConfig, dbConfig)`).

## Acceptance criteria

- [ ] By default (no `db.timestamps`), generated models contain NO `createdAt`/`updatedAt` unless the list declares them itself.
- [ ] `db: { timestamps: true }` adds `createdAt DateTime @default(now())` and `updatedAt DateTime @default(now()) @updatedAt` to all lists.
- [ ] A per-list option overrides the global setting (force on or off for one list).
- [ ] With timestamps enabled, a list that declares its own `createdAt` (or `updatedAt`) gets no duplicate auto column (no `P1012`).
- [ ] Generator output tests cover: default off, global on, per-list override, declared-own-no-dup.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

PR #498, opened 2026-06-03
feat(cli,core): configurable generator output paths via output config block

borisno2:
Implements #485
Part of #484

## What

Adds an `output` block to `opensaas.config.ts` so `opensaas generate` can relocate the generated Prisma schema and the `.opensaas` bundle — letting it coexist with an existing Keystone `prisma/` during migration without clobbering it. Defaults are unchanged for greenfield projects.

```typescript
export default config({
  output: {
    prismaSchema: 'prisma-opensaas/schema.prisma',
    opensaasDir: 'generated/opensaas',
  },
  // ...
})
```

## How

- **`OutputConfig` type** added to `@opensaas/stack-core` (`output.prismaSchema`, `output.opensaasDir`; defaults `prisma/schema.prisma` and `.opensaas`).
- **New pure resolver** `packages/cli/src/generator/output-paths.ts` — `resolveOutputPaths(cwd, output)` performs no I/O and returns absolute write paths plus the relative cross-references the generated files embed:
  - `prismaConfigSchema` — the schema *directory* relative to the project root, written into `prisma.config.ts`'s `schema` field so the Prisma CLI resolves the relocated schema.
  - `prismaClientOutput` — the patched Prisma client dir relative to the **schema file's** directory (Prisma resolves `generator { output }` from the schema file).
  - `configImport` — `opensaas.config` relative to the **`.opensaas` bundle** dir (used by `context.ts` and `prisma-extensions.ts`, which live inside the bundle).
- `generate.ts` consumes the resolver instead of the hardcoded paths.
- The generator functions (`prisma`, `prisma-config`, `context`, `prisma-extensions`) take optional cross-reference overrides that default to byte-identical historical output.

## Cross-reference computation (the tricky part)

Each cross-reference is computed from the resolved locations so the emitted code still resolves when paths move:

| relocation | `prismaConfigSchema` | `prismaClientOutput` | `configImport` |
|---|---|---|---|
| default | `prisma` | `../.opensaas/prisma-client` | `../opensaas.config` |
| schema → `prisma-opensaas/schema.prisma` | `prisma-opensaas` | `../.opensaas/prisma-client` | `../opensaas.config` |
| bundle → `generated/opensaas` | `prisma` | `../generated/opensaas/prisma-client` | `../../opensaas.config` |
| both relocated | `prisma-opensaas` | `../generated/opensaas/prisma-client` | `../../opensaas.config` |

[...]

PR #501, opened 2026-06-03
feat(cli,core): auto-timestamps off by default with db.timestamps opt-in

borisno2:
Implements #472

Part of #470

## What changed

Stop unconditionally appending `createdAt`/`updatedAt` to every generated model. Auto-timestamps are now **OFF by default** (matching Keystone 6, which never adds them automatically — see ADR-0004), re-enabled globally via `db: { timestamps: true }` with a per-list override.

- **`resolveListTimestamps(listConfig, dbConfig)`** — a small, pure, testable predicate extracted in `packages/cli/src/generator/prisma.ts`. Returns `{ createdAt, updatedAt }` flags. The Prisma generator and the TypeScript `Output` type generator both call it, so generated types stay consistent with the schema.
- **Per-list override wins** over the global setting. When timestamps resolve to on but the list already declares its own `createdAt`/`updatedAt` field, the auto column is skipped for the declared field(s) so Prisma never sees a duplicate (`P1012`).
- **Config types**: added `db.timestamps?: boolean` to `DatabaseConfig` and a per-list `db.timestamps?: boolean` override to `ListConfig`. `DatabaseConfig` is now re-exported from the `@opensaas/stack-core` root.

## Acceptance criteria

- [x] By default (no `db.timestamps`), models contain no `createdAt`/`updatedAt` unless declared by the list.
- [x] `db: { timestamps: true }` adds the auto pair to all lists.
- [x] Per-list option overrides the global setting (force on or off).
- [x] Declared own `createdAt`/`updatedAt` gets no duplicate auto column when enabled.
- [x] Generator output tests: default off, global on, per-list override (both directions), declared-own-no-dup; plus table-driven unit tests for the predicate.
- [x] Changeset added (minor, both packages).

## Notes

- Breaking change to a long-standing default; acceptable in beta per ADR-0004. Apps relying on auto-timestamps should set `db: { timestamps: true }`.
- Singleton `id Int @id @default(1)` is intentionally untouched here (separate issue under #470).
- Updated only the legitimately-changed generator snapshots (all diffs are pure deletions of the auto timestamp lines).

## Verification

- `pnpm build` — green (full monorepo, 11/11 tasks)
- `packages/cli` tests — 225 passed
- `packages/core` tests — 547 passed
- `pnpm lint` — 0 errors (4 pre-existing unrelated warnings), `pnpm manypkg fix`, `pnpm format` — clean

---

### DIFF
--- a/.changeset/auth-derive-from-better-auth.md
+++ b/.changeset/auth-derive-from-better-auth.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-auth': minor
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Derive the auth plugin's Auth lists from the better-auth config
+
+`authPlugin` now mirrors the better-auth config a developer writes instead of hardcoding the keys `User`/`Session`/`Account`/`Verification`. Per-model `modelName` becomes the OpenSaaS list key (and a table `@@map`), and the `fields` column map becomes per-field `@map`s. The plugin only ever adds/extends its own derived keys, so an app's separate domain `User` is never overwritten. The runtime `getUser`/`getCurrentUser` helpers now resolve the user list key from the configured user model instead of a hardcoded `'user'`.
+
+Default behaviour (no overrides) is unchanged: the lists are still keyed `User`/`Session`/`Account`/`Verification` with the original field shapes and no `@@map`.
+
+```typescript
+// Adopt existing better-auth tables without a destructive migration
+authPlugin({
+  user: { modelName: 'AuthUser', fields: { name: 'full_name' } },
+  session: { modelName: 'AuthSession', fields: { userId: 'user_id' } },
+  account: { modelName: 'AuthAccount' },
+  verification: { modelName: 'AuthVerification' },
+})
+// -> lists keyed AuthUser/AuthSession/AuthAccount/AuthVerification
+//    with @@map + column @map matching the live tables
+```
+
+Lists also gain a model-level `db.map` option, which emits a `@@map("...")` on the generated Prisma model so a list key can differ from its physical table name.
--- a/.changeset/auth-lists-keep-timestamps.md
+++ b/.changeset/auth-lists-keep-timestamps.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-auth': patch
+---
+
+Keep `createdAt`/`updatedAt` on the auth lists now that auto-timestamps are off by default
+
+The derived auth lists (User/Session/Account/Verification) now opt into `db: { timestamps: true }`. better-auth's adapter writes those columns and the schema converter returns `null` for them assuming the generator injects them, so the opt-in keeps the generated auth models intact.
--- a/.changeset/bright-foxes-relocate.md
+++ b/.changeset/bright-foxes-relocate.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Add an `output` config block so `opensaas generate` can relocate the generated Prisma schema and `.opensaas` bundle (e.g. to coexist with an existing Keystone `prisma/` during migration)
+
+Set `output.prismaSchema` and/or `output.opensaasDir` in `opensaas.config.ts` to move where the generator writes. Defaults are unchanged (`prisma/schema.prisma`, `.opensaas/`) when the block is omitted. The generated files' cross-references follow the configured locations: `context.ts`/`prisma-extensions.ts` import `opensaas.config` from the resolved bundle, the Prisma client `generator { output }` points back at the relocated bundle, and the top-level `prisma.config.ts` references the configured schema directory so `prisma` CLI commands keep working.
+
+The pre-existing top-level `opensaasPath` option is preserved: the effective `.opensaas` bundle directory resolves as `output.opensaasDir` > `opensaasPath` > the default `.opensaas`. Setting `opensaasPath` alone still relocates the bundle through the CLI exactly as before; `output.opensaasDir` overrides it when both are set.
+
+```typescript
+export default config({
+  output: {
+    prismaSchema: 'prisma-opensaas/schema.prisma',
+    opensaasDir: 'generated/opensaas',
+  },
+  db: {
+    /* ... */
+  },
+  lists: {
+    /* ... */
+  },
+})
+```
--- a/.changeset/calm-otters-gather.md
+++ b/.changeset/calm-otters-gather.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Re-export the fragment query API (`defineFragment`, `runQuery`, `runQueryOne`, and the `ResultOf`, `RelationSelector`, `QueryArgs` types) from the package root so the documented `import { defineFragment, runQuery, runQueryOne, type ResultOf } from '@opensaas/stack-core'` resolves.
--- a/.changeset/proud-otters-compat.md
+++ b/.changeset/proud-otters-compat.md
@@ -1,0 +1,36 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Add opt-in `db.keystoneCompat` mode for Keystone-compatible empty-string text defaults
+
+When migrating from Keystone 6, every non-null text column carries an implicit empty-string default. Set `db: { keystoneCompat: true }` to mirror that: any non-null `text()` column without an explicit `defaultValue` now generates `String @default("")`, so a migrating schema reaches parity without hand-setting `defaultValue: ''` on dozens of columns.
+
+The mode is off by default (greenfield schemas stay clean) and never affects nullable text, fields with an explicit `defaultValue`, or any non-text field — an explicit `text({ defaultValue: 'x' })` always wins.
+
+```typescript
+export default config({
+  db: {
+    provider: 'postgresql',
+    keystoneCompat: true, // non-null text without a default → @default("")
+    prismaClientConstructor: (PrismaClient) => {
+      // ... adapter setup
+    },
+  },
+  lists: {
+    Account: list({
+      fields: {
+        // required text → String @default("")
+        name: text({ validation: { isRequired: true } }),
+        // explicit default still wins → String @default("PLEASE_UPDATE")
+        status: text({ validation: { isRequired: true }, defaultValue: 'PLEASE_UPDATE' }),
+        // nullable text is untouched → String?
+        bio: text(),
+      },
+    }),
+  },
+})
+```
+
+See ADR-0004 for the full Keystone-compatible generator defaults.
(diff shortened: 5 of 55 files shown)


borisno2 (commented):
## Review: auto-timestamps off by default (`db.timestamps` opt-in)

**Verdict: REQUEST CHANGES** (posted as a Comment because GitHub blocks a formal Request-Changes when the reviewer authored the PR — treat this as Request Changes). The core generator change is clean and well-tested, but it silently breaks the first-party `@opensaas/stack-auth` plugin and the `auth-demo` example.

The mechanism itself is good: `resolveListTimestamps(listConfig, dbConfig)` is a pure, table-tested predicate; the Prisma generator and the TypeScript `Output` generator both call it so schema and types stay in lockstep; per-list override and skip-when-declared (no `P1012`) are correctly implemented; the snapshot churn is genuinely pure deletion of the auto pair (none of the snapshotted lists declared their own `createdAt`/`updatedAt`, so this is not silent column loss). ADR-0004 records the decision.

### Blocker

**1. The auth plugin lists lose `createdAt`/`updatedAt`, which breaks better-auth at runtime.**

`packages/auth/src/lists/index.ts` (`createUserList`/`createSessionList`/`createAccountList`/`createVerificationList`) does not declare `createdAt`/`updatedAt`, and the auth plugin (`packages/auth/src/config/plugin.ts`) never sets `db: { timestamps: true }`. better-auth's adapter writes `createdAt`/`updatedAt` on user/session/account/verification — in fact `packages/auth/src/server/schema-converter.ts` (`convertField`) explicitly returns `null` for `createdAt`/`updatedAt`, treating them as "auto-generated by OpenSaaS." That assumption is exactly what this PR removes.

Consequence: after this merges, any auth app that runs `pnpm generate` (default `timestamps` off, no declared fields) produces user/session/account/verification models with no `createdAt`/`updatedAt`. `db push` drops those columns and better-auth sign-up/session writes fail at runtime.

This is visible in the repo today: the committed `examples/auth-demo/prisma/schema.prisma` still has `createdAt`/`updatedAt` on User/Session/Account/Verification (and Post). That generated file was **not** regenerated in this PR, so the diff looks clean while the example is actually stale and now inconsistent with its own config.

Fix options (either is fine):
- Have the auth lists opt in via the new per-list option — set `db: { timestamps: true }` on each list returned by `getAuthLists` / `convertTableToList` (preferred: it uses the very mechanism this PR adds and keeps the columns the converter assumes exist), **or**
- Declare `createdAt`/`updatedAt` explicitly on those lists and stop skipping them in `schema-converter.ts`.

Then regenerate and commit `examples/auth-demo/prisma/schema.prisma` (and any other example running `authPlugin` / relying on the auto pair) so the checked-in schema matches the new behaviour. Without this, a first-party plugin ships broken on a default-on path.

### Non-blocking

**2. `any` leaks into a public signature.** `resolveListTimestamps(listConfig: ListConfig<any>, …)` is exported from `@opensaas/stack-cli`, so `ListConfig<any>` is now part of the public surface (the repo guideline disallows `any` in external types). The predicate only reads `db`/`fields`, which are invariant across the generic, so `ListConfig<TypeInfo>` (importing `TypeInfo` from `@opensaas/stack-core/extend`, as the test file already does) would express this without `any` or the eslint-disable. The internal `generateModelOutputType` use is less critical but could get the same treatment. Consistent with the existing `ListConfig<any>` pattern in the auth package, so it's a "would be nice," not a gate.

### Confirmed good
- `resolveListTimestamps` is pure and table-driven-tested (default off, global on, both override directions, declared-own skip for one and both). The TS types generator reuses it — Output types track the schema.
- Skip-when-declared verified for single and dual declared timestamps (no duplicate / `P1012`).
- Snapshot deletions are auto-timestamp-only; no declared timestamps were dropped.
- Changeset present, scoped to both packages as `minor`, with an accurate migration note. Breaking-default call-out matches ADR-0004.

Resolve the blocker (auth lists + regenerated example schema) and this is good to merge.

---

borisno2:
Addressed the review feedback and brought the branch up to date.

**Merged `main` in (non-interactive).** Only conflict was `packages/core/src/config/types.ts`. Resolved to preserve both sides: kept #501's `db.timestamps` (global `DatabaseConfig` + per-list `ListConfig.db`) **and** main's `db.map` / `keystoneCompat` / `isNullable` / `joinTableNaming`. The merge had produced two separate `db?:` blocks on `ListConfig` (duplicate-identifier `TS2300`); I merged them into a single `db` block carrying both `map` and `timestamps`.

**Review blocker fixed — auth plugin keeps required timestamps.** With auto-timestamps now OFF by default, the derived auth lists (User/Session/Account/Verification) would have generated without `createdAt`/`updatedAt`, breaking better-auth (its adapter writes those columns and `schema-converter.ts` returns `null` for them, assuming the generator injects them). `listDb()` in `packages/auth/src/config/derive-auth-lists.ts` now always returns `db: { timestamps: true }` (alongside any `@@map`), so the auth models opt back in. The schema converter still returns `null` for those fields — the opt-in re-injects them. Added auth tests asserting `db.timestamps === true` on all four lists (default + renamed-model configs).

**auth-demo schema regenerated.** `examples/auth-demo/prisma/schema.prisma`: the four auth models again show `createdAt`/`updatedAt`; the app's own `Post` list (no `db.timestamps`) correctly drops them — the only diff in that file.

**Non-blocking cleanup.** `resolveListTimestamps` now takes `ListConfig<TypeInfo>` instead of `ListConfig<any>`, so no `any` leaks into the public CLI surface.

Also updated the keystoneCompat generator snapshot (auto-timestamps no longer emitted) and added a `@opensaas/stack-auth` patch changeset.

**Verification:** `pnpm build` green (11/11); tests — cli 229, auth 119, core 562 (all pass); `pnpm lint` 0 errors (4 pre-existing unrelated warnings); `pnpm manypkg fix` + `pnpm format` clean.

---

borisno2 (commented):
## Re-review: blocking finding resolved

**Verdict: APPROVE** (posted as a Comment because GitHub blocks a formal Approve when the reviewer authored the PR — treat this as Approve). The prior blocker — auto-timestamps OFF by default breaking the auth plugin — is fully fixed, and the `main` merge preserved both sides cleanly.

### Blocker fixed — auth lists keep `createdAt`/`updatedAt`
`listDb()` in `packages/auth/src/config/derive-auth-lists.ts` now always returns `{ timestamps: true }` (and `{ timestamps: true, map: modelName }` for the renamed-model case), so all four derived auth lists opt back in. Verified directly:
- `createUserList`/`createSessionList`/`createAccountList`/`createVerificationList` all pass `db: listDb(...)`.
- `schema-converter.ts` still returns `null` for those columns; the per-list opt-in re-injects them — correct division of responsibility.
- New tests assert `db?.timestamps === true` on all four lists in both default and custom-`modelName` configs.

### `auth-demo` schema regenerated correctly
`examples/auth-demo/prisma/schema.prisma` at the PR head: `User`, `Session`, `Account`, `Verification` all show `createdAt DateTime @default(now())` + `updatedAt DateTime @default(now()) @updatedAt` again, while the app's own `Post` list (no opt-in) correctly drops them. That's the only diff in the file. Matches the new behaviour.

### `main` merge preserved both sides (no TS2300)
`ListConfig` in `packages/core/src/config/types.ts` has exactly **one** `db?:` block (line ~1209), carrying both `map` and `timestamps` — no duplicate `db` key, no duplicate-identifier error. The global `DatabaseConfig` retains all of main's options (`provider`, `prismaClientConstructor`, `joinTableNaming`, `keystoneCompat`, `extendPrismaSchema`) alongside the new `timestamps`. Both #501's timestamps logic and main's `db.map`/`keystoneCompat` survive intact.

### `resolveListTimestamps` no longer uses `ListConfig<any>`
Now typed `ListConfig<TypeInfo>` (with `TypeInfo` from `@opensaas/stack-core/extend`), so no `any` leaks into the exported CLI surface.

### Core behaviour confirmed
- Off by default: `listConfig.db?.timestamps ?? dbConfig.timestamps ?? false`.
- Global on + per-list override both directions — covered by generator and predicate table tests.
- Skip-when-declared via `Object.prototype.hasOwnProperty.call(...)` for `createdAt`/`updatedAt` → no duplicate field (P1012); tested for single and dual declared timestamps.
- TS `Output` generator reuses the same predicate, so generated types track the schema.

### Notes (non-blocking)
- `generateModelOutputType` in `types.ts` still uses `ListConfig<any>` with an eslint-disable, but that function is internal (not exported), so no `any` reaches the public surface — acceptable, consistent with the existing `ListConfig<any>` pattern in the auth package.
- Changesets present and correct: `quiet-otters-wander.md` (core + cli, minor) and `auth-lists-keep-timestamps.md` (auth, patch).

All prior findings addressed. Good to merge.

---

borisno2:
Rebased on `main` (merged `origin/main`, now includes #498 output-config) and got the `test` job green.

## Conflict resolution (merging `main`)
- `packages/core/src/index.ts` — the only real conflict. Kept **both** `DatabaseConfig` (this PR) and `OutputConfig` (#498) in the root type re-exports.
- `packages/core/src/config/types.ts`, `packages/cli/src/generator/prisma.ts`, `packages/core/src/config/index.ts` — auto-merged cleanly. Verified the consolidated single `ListConfig.db` block carries both `map` (#498/#497) and `timestamps` (this PR) with no `TS2300`, and that `prisma.ts` retains the timestamps loop, `@@map`/`keystoneCompat` threading, and the `clientOutput` line together.

## Root cause of the `test` failure
It was **not** `format:check` — that passes. The `test` job runs `turbo run test`, whose pipeline builds the examples first (`generate && next build`). Two example builds failed type-checking because they read auto-timestamps that this PR removes by default:
- `examples/blog/test-hooks-type-inference.ts:57` — `args.item.createdAt` no longer exists on `Post`.
- `examples/composable-dashboard/app/page.tsx:24` (and posts/users pages) — `orderBy: { createdAt: 'desc' }` / `post.createdAt` no longer valid.

Several committed example `prisma/schema.prisma` snapshots were also stale (still carrying the old auto-timestamps, and missing #493's `@default()` output).

## Fix
- Opted the two examples that genuinely use timestamps into `db: { timestamps: true }` (blog, composable-dashboard).
- Regenerated the stale example schemas. User-defined lists now correctly drop auto `createdAt`/`updatedAt`; **auth lists keep them** via the plugin opt-in (verified in `examples/starter-auth/prisma/schema.prisma`: only `Post` loses timestamps; `User`/`Session`/`Account`/`Verification` retain them). The regenerated schemas also pick up #493's `@default()` formatting.

Intended behaviour is preserved: auto-timestamps off by default, auth lists keep timestamps.

## CI `test` step results (run locally, exact order)
| Step | Result |
| --- | --- |
| `pnpm manypkg check` | PASS |
| `pnpm lint` | PASS (0 errors, 4 pre-existing warnings) |
| `pnpm format:check` | PASS |
| `turbo run test` | PASS (32/32 tasks) |
| `turbo run test:coverage` | PASS (32/32 tasks) |
| browser tests (`packages/ui` `test:browser:coverage`) | **Could not run** — sandbox blocks the Chromium download (`cdn.playwright.dev` not in network allowlist; `playwright install --with-deps` also blocked). The runner loads and discovers all 5 browser test files and fails only on the missing browser binary, so this is an environment limitation, not a code/merge issue. Will be validated by CI. |

---

Issue #474, opened 2026-06-03
select() db options: isNullable (force ? with default) + enumName (native enum type name)


borisno2:
## Parent

#470

## What to build

Two additive `select()` `db` options for Keystone schema parity:

1. **`db: { isNullable: true }`** forces the nullable `?` on the generated column even when a `defaultValue` is present, so an optional select with a default generates `String? @default("X")` (or the enum equivalent). The default behaviour (NOT NULL when a default is present) is unchanged when `isNullable` is not set — this is an explicit opt-in, not an automatic restoration of `?`.
2. **`db: { type: 'enum', enumName: '...' }`** sets the generated Prisma enum type name (overriding the derived `<List><Field>`), updating every reference to that enum in the owning model.

## Acceptance criteria

- [ ] `select({ defaultValue: 'X', db: { isNullable: true } })` (not isRequired) generates a nullable column with the default (`String? @default("X")` for string-backed, `<Enum>? @default(X)` for enum-backed).
- [ ] Without `db.isNullable`, an optional select with a default still generates NOT NULL (unchanged behaviour).
- [ ] `select({ db: { type: 'enum', enumName: 'AccountNoteStatusType' } })` names the generated `enum` block `AccountNoteStatusType` and the column references it.
- [ ] `db.enumName` is added to the select field's db config types.
- [ ] Generator output tests cover nullable-with-default and custom enum name.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

PR #502, opened 2026-06-03
authPlugin schema placement: relocate Auth lists to a non-public DB schema (clean-diff adoption)

borisno2:
Implements #482
Part of #480

## What

Lets the Auth lists be placed in a non-`public` Postgres schema (e.g. `auth`) so they can adopt a separate-schema better-auth layout and diff CLEAN against the live tables. Builds on the #481/#497 derivation module (`deriveAuthLists`) — it does not reimplement it.

## How `@@schema` + datasource `schemas` are emitted

- **Core (`@opensaas/stack-core`)** — mirroring the existing `db.map` → `@@map`:
  - `ListConfig.db.schema?: string` — a list-level schema.
  - `DatabaseConfig.schemas?: string[]` — the datasource schema list.
- **CLI generator (`packages/cli/src/generator/prisma.ts`)**:
  - When `db.schemas` is set: emits `previewFeatures = ["multiSchema"]` on the `generator` block and `schemas = [...]` on the `datasource` block (both required by Prisma for multi-schema).
  - Per model: emits `@@schema("...")` from `listConfig.db.schema` (right after `@@map`).
- **Auth (`@opensaas/stack-auth`)**:
  - Plugin-level `authPlugin({ schema: 'auth' })` applies `@@schema("auth")` to all derived Auth lists; a per-model override (`verification: { schema: 'auth_internal' }`) relocates a single list.
  - `deriveAuthLists` now sets each list's `db.schema` (per-model override, else plugin-level schema). `normalizeAuthConfig` resolves it.
  - The plugin's new `beforeGenerate` hook wires the datasource: it merges the auth schema(s) into `db.schemas` (always including `public`) and defaults any list without an explicit `db.schema` to `public`, so the generated multi-schema schema is coherent and valid.

## Greenfield default unchanged

With no `schema` option, nothing changes: no `@@schema`, no `schemas` array, no `multiSchema` preview feature — the `beforeGenerate` hook returns the config untouched. Covered by an explicit test.

## Clean-diff adoption proof

- CLI generator test models a representative live `auth`-schema install (custom keys from #481 pinned via `@@map`, placed via `@@schema("auth")`, FK column `@map("user_id")`, app's own `public.User` left alone) and asserts the generated schema is a valid multi-schema schema matching the live tables — i.e. modelled for runtime/types without a migration.
[...]

PR #503, opened 2026-06-03
feat(core,cli): select() db.isNullable + db.enumName

borisno2:
Implements #474. Part of #470.

## What changed

Two additive `select()` `db` options for Keystone schema parity:

1. **`db: { isNullable: true }`** forces the nullable `?` on the generated column even when a `defaultValue` is present, so an optional select with a default generates `String? @default("X")` (or `<Enum>? @default(X)` for enum-backed). The default behaviour (NOT NULL once a default is present) is **unchanged** when `isNullable` is not set — explicit opt-in, not auto-restoration of `?`.
2. **`db: { type: 'enum', enumName: '...' }`** overrides the derived `<List><Field>` Prisma enum type name, renaming both the generated `enum` block and every reference to it in the owning model.

## Implementation notes

- `packages/core/src/config/types.ts`: added `isNullable?: boolean` and `enumName?: string` to the `SelectField` `db` config (documented).
- `packages/core/src/fields/index.ts` (`select().getPrismaType`): nullability is now assembled independently of the default with `+=` (mirroring `text`/`integer`), so a default never overwrites the `?`. Rule: `isNullable = db.isNullable ?? (!isRequired && !hasDefault)` — a present `defaultValue` still yields NOT NULL unless `db.isNullable` is set. For native-enum selects, the type name is `db.enumName ?? <List><Field>`; since the generator derives the enum block name from `result.type`, this renames the block and the column reference together. No generator changes were needed.

## Tests

- `packages/core/src/fields/select.test.ts`: unit tests for NOT-NULL-with-default (unchanged), `db.isNullable` forcing `?` (string + enum), `db.enumName` override, and `enumName` ignored for string selects.
- `packages/cli/src/generator/prisma.test.ts`: generator output tests — `enumName` renames the enum block + column reference; optional select + default + `db.isNullable` emits `String? @default("X")` and `<Enum>? @default(draft)`; without `isNullable` the column stays NOT NULL (string + enum).

## Verification

- `pnpm build` (full monorepo typecheck): pass
- `packages/core`: 565 tests pass
- `packages/cli`: 221 tests pass (no snapshot churn)
- `pnpm lint`: 0 errors (4 pre-existing unrelated warnings); `pnpm manypkg fix`; `pnpm format`: clean

Changeset added (`minor`, beta): `@opensaas/stack-core`, `@opensaas/stack-cli`.

---

### DIFF
--- a/.changeset/select-db-nullable-enumname.md
+++ b/.changeset/select-db-nullable-enumname.md
@@ -1,0 +1,47 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Add `select()` db options for Keystone schema parity: `db.isNullable` and `db.enumName`.
+
+`db.isNullable: true` forces the nullable `?` on the generated column even when a
+`defaultValue` is present. The default behaviour is unchanged — a select with a
+`defaultValue` still generates NOT NULL unless you opt in explicitly:
+
+```typescript
+// Optional select with a default, kept nullable for data containing NULLs
+status: select({
+  options: [
+    { label: 'Draft', value: 'draft' },
+    { label: 'Published', value: 'published' },
+  ],
+  defaultValue: 'draft',
+  db: { isNullable: true },
+})
+// Generates: status String? @default("draft")
+
+// Enum-backed equivalent
+status: select({
+  options: [{ label: 'Open', value: 'open' }],
+  defaultValue: 'open',
+  db: { type: 'enum', isNullable: true },
+})
+// Generates: status <Enum>? @default(open)
+```
+
+`db.enumName` overrides the derived `<List><Field>` name of the generated Prisma
+enum for native-enum selects, renaming both the `enum` block and every reference
+to it in the owning model — useful for matching a live DB enum (e.g. Keystone's
+`…Type` suffix):
+
+```typescript
+status: select({
+  options: [
+    { label: 'Open', value: 'open' },
+    { label: 'Closed', value: 'closed' },
+  ],
+  db: { type: 'enum', enumName: 'AccountNoteStatusType' },
+})
+// Generates: enum AccountNoteStatusType { ... } and the column references it
+```
--- a/packages/cli/src/generator/prisma.test.ts
+++ b/packages/cli/src/generator/prisma.test.ts
@@ -1577,6 +1577,136 @@ describe('Prisma Schema Generator', () => {
       }).not.toThrow()
     })
 
+    it('should rename the enum block and the column reference with db.enumName', () => {
+      const config: OpenSaasConfig = {
+        db: { provider: 'sqlite' },
+        lists: {
+          AccountNote: {
+            fields: {
+              status: select({
+                options: [
+                  { label: 'Open', value: 'open' },
+                  { label: 'Closed', value: 'closed' },
+                ],
+                db: { type: 'enum', enumName: 'AccountNoteStatusType' },
+              }),
+            },
+          },
+        },
+      }
+
+      const schema = generatePrismaSchema(config)
+
+      // The enum block uses the custom name
+      expect(schema).toContain('enum AccountNoteStatusType {')
+      expect(schema).toContain('  open')
+      expect(schema).toContain('  closed')
+      // The column references the custom enum name
+      expect(schema).toMatch(/status\s+AccountNoteStatusType/)
+      // The derived name is not emitted anywhere
+      expect(schema).not.toContain('enum AccountNoteStatus {')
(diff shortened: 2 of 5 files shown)


borisno2 (commented):
## Code Review — PR #503: `feat(core,cli): select() db.isNullable + db.enumName`

**Verdict: APPROVE** (posted as a Comment review because the PR author matches the reviewing identity, so GitHub blocks a formal self-Approve — the intended verdict is Approve.)

Clean, additive, well-tested change. I reviewed strictly from the diff plus the full PR-head versions of `packages/core/src/fields/index.ts`, `packages/cli/src/generator/prisma.ts`, and `packages/core/src/config/types.ts`. All five behaviours called out in the spec hold up.

### Correctness — verified

- **Default behaviour unchanged.** `isNullable = options.db?.isNullable ?? (!isRequired && !hasDefault)`. An optional `select()` with a `defaultValue` but no `db.isNullable` resolves to `false`, so no `?` is emitted — NOT NULL, exactly as before. Confirmed by the unit + generator tests for both string and enum backings. The opt-in is explicit.
- **`db.isNullable: true` forces `?` with a default**, for both backings: `String? @default("X")` and `<Enum>? @default(X)`. Unit and generator tests cover both. Also correctly wins for a required field (`isRequired: true` + `isNullable: true` → still `?`), which is tested.
- **`db.enumName` renames the enum block and the column reference together.** Both flow from a single source: `getPrismaType` returns `type: enumName`, the generator emits the block via `enumDefinitions.set(result.type, …)` and references the column via `mapFieldTypeToPrisma` → `result.type`. One value, two consumers — no chance of drift. Ignored for string selects since the assignment lives inside `if (isNativeEnum)`; the `should ignore db.enumName for string selects` test pins this.
- **Nullability assembled with `+=`.** `?` is appended first, then `@default` via `+=`, so the default never overwrites the `?`. This now mirrors `text`/`integer`, replacing the old `modifiers = …` reassignment that previously clobbered the `?`. The generator's `modStr.startsWith('?')` splitter then renders the spacing correctly (`String? @default(...)`).

### Type safety / hygiene

- New `db` options are plainly typed (`isNullable?: boolean`, `enumName?: string`) with thorough TSDoc. No `any`, no casts, no `unknown` leaking into the public surface.
- Changeset present (`minor`, both `@opensaas/stack-core` and `@opensaas/stack-cli`). PR reports no snapshot churn and full suites green.

### Minor (non-blocking, optional follow-up)

- **`enumName` isn't validated** against the Prisma identifier pattern, whereas enum *values* are (`PRISMA_ENUM_VALUE_PATTERN`). A malformed `enumName` would emit invalid Prisma that only fails later at `prisma generate`. Since this is an advanced escape hatch for matching a live DB enum, deferring validation is reasonable — but a cheap guard mirroring the value check would fail faster with a clearer message.
- **No collision guard** if two enum-backed selects resolve to the same `enumName`: the generator's `Map` keyed by `result.type` would silently merge them (last `enumValues` wins). This already applies to derived names and isn't introduced here, so it's pre-existing — just flagging for awareness.

Nice parity work, and good test coverage of the unchanged-default path specifically. LGTM.

🤖 Generated with [Claude Code](https://claude.ai/code/session_01ULd2HCT8dUve9aa9gKi6sX)

---

borisno2:
Rebased on `main` (merge `7b4202c`, base now `593390c`) — picks up auto-timestamps-off (#501), auth schema placement (#502), output config (#498), and the e2e turbo fix (#508). The branch was `dirty`; it's now clean (`mergeable_state: blocked` = checks/approval only, no conflicts).

`config/types.ts` and `fields/index.ts` merged without conflicts: `SelectField.db.isNullable?`/`enumName?` and the `select().getPrismaType` `+=` nullability + enum-rename logic are all intact alongside main's `db.timestamps`, `db.schemas`/`db.schema`, `OutputConfig`/`output?`, `db.map`, `db.keystoneCompat`, and `joinTableNaming`. The CLI `prisma.ts` generator was only touched by main (this PR adds no generator changes — the logic lives in core), so main's version is taken as-is. No snapshot regeneration was needed.

CI `test` job reproduced locally, all green:
- `pnpm manypkg check` — pass
- `pnpm format:check` — clean
- `turbo run test` — 32/32 tasks pass (core 569, cli 271; all example builds typecheck cleanly)
- `turbo run test:coverage` — 32/32 tasks pass (ui browser tests ran: 164 pass)

Behavioral contract verified unchanged: an optional select with a `defaultValue` and no `isNullable` still generates NOT NULL (`@default("draft")` / `@default(draft)`); `db.isNullable: true` forces `?` for both string and enum-backed selects.

🤖 Generated with [Claude Code](https://claude.ai/code/session_01ULd2HCT8dUve9aa9gKi6sX)

---

Issue #483, opened 2026-06-03
"Adopt existing better-auth tables" preset + docs (app User vs auth identity linking)


borisno2:
## Parent

#480

## What to build

A convenience preset/recipe plus documentation that ties together the keys/field derivation (#481) and schema placement (#482) into a one-stop "adopt existing better-auth tables" path.

- An "adopt existing better-auth tables" preset/recipe that sets sensible defaults (separate schema, existing field names derived from better-auth config) so a migrator doesn't reconstruct the config from scratch.
- Documentation in `packages/auth` (CLAUDE.md/README) and the migration guide covering: configuring the plugin to match a live better-auth install, keeping the app's domain `User` separate from the **Auth identity**, and linking the two from the app side (a relationship/field the app declares).

## Acceptance criteria

- [ ] An "adopt existing better-auth tables" preset/recipe is documented and works end-to-end with #481 + #482 to produce a clean `schema:diff` against a representative live install.
- [ ] Docs clearly distinguish the app's domain `User` from the auth identity and show how to link them.
- [ ] Migration guide references the adoption path (no destructive auth migration).
- [ ] Changeset added (see `pr-changeset` skill); any plugin/skill changes get a `plugin-version` bump if applicable.

## Blocked by

- #481
- #482

Issue #482, opened 2026-06-03
authPlugin schema placement: put Auth lists in a non-public DB schema (clean-diff adoption)


borisno2:
## Parent

#480

## What to build

Let the auth lists be placed in a non-`public` database schema so they can adopt a separate-schema better-auth layout (e.g. `auth`) and diff clean against the live tables.

- A plugin-level `schema` option that applies `@@schema(...)` (and the right `@@map`) to all generated **Auth lists**, reusing the stack's existing multi-schema support.
- A per-list override so an individual auth list can target a different schema.
- Combined with the derived keys/field maps from #481, the generated auth lists diff CLEAN against an existing `auth`-schema better-auth installation (model-only, no migration).

## Acceptance criteria

- [ ] `authPlugin({ schema: 'auth' })` places all generated auth lists in the `auth` schema (`@@schema("auth")`) with correct table `@@map`.
- [ ] A per-list schema override targets a different schema for one auth list.
- [ ] Schema-placement generator test: a representative live `auth`-schema shape (custom keys from #481 + `auth` schema) produces an empty `schema:diff` (adoption).
- [ ] Greenfield default (no `schema` option) is unchanged (`public`).
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

- #481

PR #509, opened 2026-06-03
authPlugin: adoptBetterAuthTables() recipe + adoption docs (app User vs Auth identity)

borisno2:
Implements #483
Part of #480

## What this adds (D3 of PRD #480)

A convenience "adopt existing better-auth tables" recipe plus documentation that ties together the keys/field derivation (#481) and schema placement (#482) into a one-stop adoption path.

- **`adoptBetterAuthTables()` recipe** (`packages/auth/src/config/adopt-better-auth-tables.ts`): a pure function returning the `AuthConfig` adoption fragment — the plugin-level `schema` plus each model's `modelName` (and optional column `fields` maps) — preset to the conventions of a standard separate-schema better-auth install (`AuthUser`/`AuthSession`/`AuthAccount`/`AuthVerification` in an `auth` schema). Spread it into `authPlugin` alongside the rest of your config. Options: `schema`, `modelNamePrefix`, per-model `fields`.
- Exported from the package root (`adoptBetterAuthTables`, `AdoptBetterAuthTablesOptions`, `AdoptBetterAuthTablesConfig`).
- **App `User` ≠ Auth identity**: because the recipe's derived user key is `AuthUser` (not `User`), the plugin only ever adds/extends its derived keys — an app's own domain `User` is never extended or overwritten. Linking the app `User` to the Auth identity is documented as the application's concern (a `relationship({ ref: 'AuthUser' })` the app declares).

## Documentation

- Authentication guide: new "Adopting an Existing better-auth Installation" section (app `User` vs Auth identity, the recipe + customisation, Schema-parity / clean-diff check, and linking the two).
- `packages/auth/CLAUDE.md` + `README.md`: adoption-recipe subsection.
- Migration guides (`migrating-from-keystone.md`, `migration.md`): reference the adoption path (no destructive auth migration).

## Tests (`packages/auth/tests/adopt-better-auth-tables.test.ts`)

- Recipe defaults and `schema`/`modelNamePrefix`/`fields` options.
- Composition: spreads into `authPlugin` alongside the rest of the auth config.
- **Clean-diff adoption assertion**: every Auth list lands in the `auth` schema with `@@map` + `@@schema` (auto-timestamps preserved) while the app's separate `public.User` stays untouched; the datasource lists both schemas.
- Field column maps flow through to the derived field-level `@map` / FK `@map`.

## Verification

- `pnpm build` — pass
- `packages/auth` tests — 133 passed (8 files); `packages/core` — 562 passed; `packages/cli` — 266 passed
- `pnpm lint` — 0 errors (4 pre-existing warnings in untouched files)
- `pnpm manypkg check` — pass
- `pnpm format:check` — pass
- docs `link-check` — pass

Changeset: minor for `@opensaas/stack-auth`. No `claude-plugins/*` changes, so no plugin-version bump.

---

### DIFF
--- a/.changeset/auth-adopt-better-auth-tables.md
+++ b/.changeset/auth-adopt-better-auth-tables.md
@@ -1,0 +1,35 @@
+---
+'@opensaas/stack-auth': minor
+---
+
+Add `adoptBetterAuthTables()` recipe for adopting an existing better-auth installation
+
+A migrating project that already runs better-auth (its `AuthUser`/`AuthSession`/`AuthAccount`/`AuthVerification` tables live in a separate `auth` Postgres schema, and its app `User` is a different model) can now adopt those live tables without rebuilding the auth config by hand. The recipe presets the plugin-level `schema` plus each model's `modelName` (and optional column `fields` maps) to the conventions of a standard separate-schema better-auth install, so the derived Auth lists diff clean (Schema parity) against the live database — no destructive auth migration. The app's own domain `User` is left untouched; linking it to the Auth identity is the application's concern.
+
+```typescript
+import { config } from '@opensaas/stack-core'
+import { authPlugin, adoptBetterAuthTables } from '@opensaas/stack-auth'
+
+export default config({
+  db: { provider: 'postgresql', url: process.env.DATABASE_URL },
+  plugins: [
+    authPlugin({
+      // Defaults: AuthUser/AuthSession/AuthAccount/AuthVerification in the
+      // `auth` schema, pinned to your live table names (@@map) + schema (@@schema).
+      ...adoptBetterAuthTables(),
+      emailAndPassword: { enabled: true },
+    }),
+  ],
+  lists: {
+    // Your own domain User stays in `public` and is NOT touched by the plugin.
+    User: list({ fields: { subjectId: text({ validation: { isRequired: true } }) } }),
+  },
+})
+
+// Customise when your live tables diverge from the defaults:
+adoptBetterAuthTables({
+  schema: 'identity', // default: 'auth'
+  modelNamePrefix: 'BA', // default: 'Auth'
+  fields: { user: { name: 'full_name' }, session: { userId: 'user_id' } },
+})
+```
--- a/docs/content/guides/authentication.md
+++ b/docs/content/guides/authentication.md
@@ -917,6 +917,137 @@ Verification link: http://localhost:3000/api/auth/verify-email?token=abc123def45
 4. Better-auth verifies token and updates password
 5. User can sign in with new password
 
+## Adopting an Existing better-auth Installation
+
+If you are migrating a project that **already runs better-auth** — its tables
+exist, hold live data, and you don't want a destructive auth migration — the
+plugin can adopt those tables instead of recreating them. This is the common
+case when migrating an established app to OpenSaaS Stack.
+
+### App `User` vs the Auth identity
+
+A migrating app almost always has **two** distinct concepts, and conflating them
+is the main pitfall:
+
+- **The application's domain `User`** — your own model (`public.User`), keyed
+  however your app likes (e.g. a `subjectId`), carrying your domain fields
+  (profile, billing, roles, relationships to your other lists).
+- **The Auth identity** — the better-auth-owned user record (`AuthUser`), the
+  thing a session belongs to. It owns sessions, OAuth accounts, and credentials.
+
+The auth plugin models the **Auth identity** (the `User`/`Session`/`Account`/
+`Verification` lists better-auth needs). It does **not** assume its user list is
+your app's `User`. As long as the plugin's user model key differs from your app
+list's key (e.g. the plugin uses `AuthUser` while your app keeps `User`), your
+domain `User` is left completely untouched — never extended, never overwritten.
+
+{% callout type="info" %}
+Keep these separate. The plugin owns the Auth identity; your app owns its domain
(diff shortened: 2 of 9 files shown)


borisno2 (commented):
## Review — PR #509: `adoptBetterAuthTables()` recipe (D3 of #480, implements #483)

**Verdict: APPROVE** (posted as a Comment since the PR was authored by an integration identity). This is a clean, well-scoped, well-documented change. The recipe is a pure function that only sets adoption knobs, it spreads cleanly into `authPlugin`, and the clean-diff guarantee is genuinely asserted by the tests. One minor test-naming/coverage nit and a couple of doc/optional suggestions below — none blocking.

I reviewed the diff against the surrounding (merged) source: `config/types.ts` (`AuthConfig`/`AuthModelConfig`), `config/index.ts` (`normalizeAuthConfig`), `config/derive-auth-lists.ts`, and `config/plugin.ts`.

### Correctness — verified against the real pipeline

- **Returns a correct `AuthConfig` fragment.** `AdoptBetterAuthTablesConfig = Pick<AuthConfig, 'schema' | 'user' | 'session' | 'account' | 'verification'>` is concrete and spreads cleanly. `buildModel()` returns `AuthModelConfig` (`{ modelName, fields? }`); assigning it to the `session` slot (typed `SessionConfig & AuthModelConfig`) is sound because every `SessionConfig` member is optional, so this composes without a cast. Confirmed.
- **Only adoption knobs, no side effects.** The function is pure: it sets `schema` + per-model `modelName` (+ optional `fields`) and nothing else. Providers, `sessionFields`, `extendUserList`, etc. all remain the developer's to add — exactly as documented.
- **Plugin-level `schema` actually reaches the generator.** `normalizeModelConfig(..., defaultSchema)` (in `config/index.ts`) propagates the plugin-level `schema` onto every `model.schema`, and `plugin.ts#beforeGenerate` collects `model.schema` (not the plugin-level field) into the datasource `schemas`. So `schema: 'auth'` correctly lands on all four lists. Confirmed end-to-end.
- **Clean-diff / "app User untouched" guarantee holds and is asserted.** Because the derived user key is `AuthUser` (not `User`), `plugin.ts#init`'s add-vs-extend only extends a list that shares the *derived* key — the app's `public.User` is never touched. The test `lands every Auth list in the auth schema ... and leaves the app User untouched` asserts the real shapes:
  - each Auth list `db` is `{ timestamps: true, map: 'Auth…', schema: 'auth' }` (ADR-0004 auto-timestamps preserved + `@@map` + `@@schema`);
  - the app `User` keeps `subjectId`, gains none of `email`/`emailVerified`/`sessions`, and stays in `public`;
  - `db.schemas` is `['public', 'auth']`.
  This is a genuine assertion of the adoption guarantee, not a rubber-stamp.
- **App-User ≠ Auth-identity separation** is correct in code (separate keys) and clearly documented (auth guide section + linking subsection: linking is the app's concern via `relationship({ ref: 'AuthUser' })`).
- **Field column maps flow through** to field-level `@map` and the FK `@map` — asserted by the `name: 'full_name'` / `session.userId: 'user_id'` test against `fields.name.db?.map` and `fields.user.db?.foreignKey`.
- **No `any` / no casts in the new public surface.** `adopt-better-auth-tables.ts` is fully typed; option and return types are concrete. (The pre-existing `betterAuthPlugins?: any[]` in `AuthConfig` is out of scope here.)
- **Docs are accurate and non-destructive.** The auth-guide section, README, CLAUDE.md, and both migration guides consistently frame this as "modelled for runtime/types, no destructive auth migration" with a Schema-parity diff check — no false claims of a migration.

### Minor (non-blocking)

1. **Test name vs. assertion mismatch — `'public'` schema path is under-covered.** `adopt-better-auth-tables.test.ts`:
   > `it('omits the schema when explicitly set to public', ...)`
   only asserts `fragment.schema` is `'public'` — it does **not** assert anything is "omitted." This matters because the behavior is the opposite of what the name implies: passing `schema: 'public'` sets `model.schema = 'public'` on all four models, so `beforeGenerate` emits `schemas: ['public']` **and** `@@schema("public")` on every Auth list (it is treated as an explicit single-schema install, not as "no schema"). That is arguably the intended behavior for "adopt a public-schema better-auth install," but the test name reads as a no-op-greenfield claim it doesn't verify. Suggest either (a) rename to `keeps the schema when explicitly set to public`, or (b) add a `generationConfig()` assertion showing the actual emitted `db.schemas`/`db.schema` for the `schema: 'public'` case so the behavior is pinned down. The recipe's `schema` JSDoc ("Pass `'public'` … for an install that is not on a separate schema") is consistent with (a)/the explicit-schema behavior.

2. **Greenfield/`schema: undefined` is not directly exercised here.** Not required (the underlying schema-placement behavior is tested in `plugin-schema-placement.test.ts`, and the recipe always defaults `schema: 'auth'`), so this is just a note, not a request.

3. **`modelNamePrefix: ''` escape hatch.** The JSDoc rightly warns that an empty prefix yields `User`/`Session`/... which would collide-and-extend an app `User`. Good that it's documented; no code change needed. Optional: a one-line test asserting `adoptBetterAuthTables({ modelNamePrefix: '' }).user.modelName === 'User'` would lock the documented escape-hatch behavior.

### Process
- Changeset present and correctly scoped (`@opensaas/stack-auth`, minor) with a usage example. No `claude-plugins/*` changes, so no plugin-version bump needed — correct.

Nice work. The single thing worth tidying before merge is the test-name/coverage point (#1); everything else is optional polish.

---

Issue #473, opened 2026-06-03
Singleton lists emit bare `id Int @id` (drop @default(1))


borisno2:
## Parent

#470

## What to build

Singleton lists must generate a bare `id Int @id` with no `@default(1)`, matching Keystone 6 (which emits no column default for singleton ids). Remove the hardcoded `@default(1)` from singleton id generation in the Prisma generator.

## Acceptance criteria

- [ ] A singleton list generates `id Int @id` (no `@default(1)`).
- [ ] Non-singleton lists are unaffected (still `id String @id @default(cuid())`).
- [ ] Generator output test asserts the singleton id line.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

PR #510, opened 2026-06-03
Singleton lists emit bare `id Int @id` (drop @default(1))

borisno2:
Implements #473

Part of #470

## What changed

Singleton lists now generate a bare `id Int @id` with no `@default(1)`, matching Keystone 6 (which emits no column default for singleton ids). This lets singletons reach **Schema parity** from config alone instead of forcing migrators to strip the default via `extendPrismaSchema` (see ADR-0004).

Non-singleton lists are completely unaffected and still emit `id String @id @default(cuid())`.

### Current vs new singleton id output

```prisma
// before
model Settings {
  id        Int      @id @default(1)
  ...
}

// after
model Settings {
  id        Int      @id
  ...
}
```

## Notes

- The stack already injects `id: 1` for singleton creates at runtime (`packages/core/src/context/write-pipeline.ts`), so dropping the Prisma column default is runtime-safe.
- Because a bare `Int @id` has no default, Prisma's generated `CreateInput` now requires `id`. The blog example's `test-singleton.ts` (which deliberately attempts a forbidden second create) was updated to supply an `id`; that create is still rejected by the singleton constraint before reaching the database.

## Tests

- Generator output test asserts a singleton emits bare `id Int @id` (no default) and a normal list's `id String @id @default(cuid())` is unchanged. Snapshot updated accordingly.

## Acceptance criteria

- [x] A singleton list generates `id Int @id` (no `@default(1)`).
- [x] Non-singleton lists are unaffected (still `id String @id @default(cuid())`).
- [x] Generator output test asserts the singleton id line.
- [x] Changeset added (minor, `@opensaas/stack-cli`).

## Verification

- `pnpm build` — pass
- `packages/cli` tests — 271 pass
- `packages/core` tests — 569 pass
- `pnpm turbo run test` (incl. example builds) — pass
- `pnpm manypkg check` — valid
- `pnpm format:check` — clean

---

### DIFF
--- a/.changeset/singleton-id-bare.md
+++ b/.changeset/singleton-id-bare.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-cli': minor
+---
+
+Singleton lists now emit a bare `id Int @id` (no `@default(1)`) to match Keystone 6, so singletons reach Schema parity from config alone instead of needing `extendPrismaSchema` to strip the column default (see ADR-0004).
+
+```ts
+lists: {
+  Settings: list({
+    isSingleton: true,
+    fields: { siteName: text() },
+  }),
+}
+```
+
+Generated Prisma schema:
+
+```prisma
+model Settings {
+  id        Int      @id
+  siteName  String
+}
+```
+
+Non-singleton lists are unaffected and continue to emit `id String @id @default(cuid())`.
--- a/examples/blog/prisma/schema.prisma
+++ b/examples/blog/prisma/schema.prisma
@@ -8,7 +8,7 @@ datasource db {
 }
 
 model Settings {
-  id              Int      @id @default(1)
+  id              Int      @id
   siteName        String   @default("My Blog")
   maintenanceMode Boolean  @default(false)
   maxUploadSize   Int?     @default(10)
--- a/examples/blog/test-singleton.ts
+++ b/examples/blog/test-singleton.ts
@@ -41,8 +41,12 @@ async function main() {
     // Test 3: Try to create a second record - should fail
     console.log('✅ Test 3: Try to create a second record - should fail')
     try {
+      // Singleton ids are now a bare `id Int @id` with no column default (see
+      // ADR-0004), so Prisma's CreateInput requires `id`. The stack still injects
+      // `id: 1` at runtime; this create is rejected by the singleton constraint
+      // before reaching the database, so the supplied id is never used.
       await context.db.settings.create({
-        data: { siteName: 'Second Blog' },
+        data: { id: 2, siteName: 'Second Blog' },
       })
       throw new Error('Should have thrown an error when creating second record')
     } catch (error) {
--- a/packages/cli/src/generator/__snapshots__/prisma.test.ts.snap
+++ b/packages/cli/src/generator/__snapshots__/prisma.test.ts.snap
@@ -197,7 +197,7 @@ datasource db {
 "
 `;
 
-exports[`Prisma Schema Generator > select field with enum type > should generate singleton model with Int @id @default(1) 1`] = `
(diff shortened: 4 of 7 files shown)


borisno2 (commented):
## Review — Verdict: APPROVE (posted as Comment; PR authored by integration identity)

Reviewed strictly from the PR diff plus the runtime code it depends on. This is a clean, correctly-scoped implementation of #473 / ADR-0004. I'd approve formally if GitHub permitted it for this PR author.

### Scope & correctness (all confirmed)

- **Singleton id is now bare, non-singleton untouched.** `packages/cli/src/generator/prisma.ts` emits `id        Int      @id` for `isSingleton` lists and leaves `id        String   @id @default(cuid())` exactly as-is for everything else. The TS id type (`types.ts`, `generateModelOutputType`) stays `number` for singletons / `string` otherwise — unchanged and consistent.
- **Matches ADR-0004** verbatim ("Singleton `id` is bare … no `@default(1)`, matching Keystone 6"). Comments in `prisma.ts` and `types.ts` now cite the ADR.
- **Snapshot change is minimal and accurate.** The only content change in `prisma.test.ts.snap` is the singleton id line (`@default(1)` dropped); the test name was updated to match. The remaining (pre-existing) column-misalignment in that fixture is untouched, as expected.
- **Test assertions improved, not weakened.** The singleton test now asserts `toContain('id        Int      @id\n')` AND `not.toContain('@default(1)')`. The non-singleton test now asserts `not.toContain('id        Int      @id')` — a strictly stronger guard than the old `@default(1)`-specific one. Good.

### Runtime safety (the important bit) — verified

The premise "a bare `Int @id` has no column default, so Prisma's `CreateInput` now requires `id`" is correct: the generated `SettingsCreateArgs` only `Omit`s `select`/`include` from `Prisma.SettingsCreateArgs`, so `data` is still typed by Prisma's own `CreateInput`, which requires `id` for a bare `Int @id`.

Crucially, **normal usage does not break**:
- The stack injects the id at write time in `createWriteStrategy.persist` (`write-pipeline.ts`): `const createData = singleton ? { id: 1, ...data } : data`. Unchanged by this PR.
- The normal singleton entry point `context.db.settings.get()` auto-creates via `createGet` → `createFn({ data: defaultData })`, and `getDefaultData` (`context/index.ts`) explicitly **skips** `id`. So `get()` relies entirely on that runtime injection — which is intact. No `id` is required from callers in the happy path.
- The admin UI create path (`serverAction` → `model.create`) likewise flows through the same injection, so it's unaffected.

### `test-singleton.ts` fix — correct and intent-preserving

Test 3 deliberately attempts a forbidden *second* create. The `count() > 0` guard in `createWriteStrategy.resolveTarget` throws the singleton `ValidationError` **before** `persist`/the DB call, so the supplied `id: 2` is never used — the test still asserts the create is rejected with a `singleton` error message. The added comment explains this accurately. Intent preserved.

### Style / conventions

- No new `any` or type casts introduced by the diff. (Pre-existing `eslint-disable @typescript-eslint/no-explicit-any` lines in `write-pipeline.ts`/`context/index.ts` for runtime-resolved model names are unchanged and out of scope.)
- Changeset present and correct (`@opensaas/stack-cli`, minor) with before/after example. Generator-only change, so cli is the right package.

### Non-blocking note (nit, not introduced here)

`persist` spreads `{ id: 1, ...data }`, so a caller-supplied `id` would override the injected `1`. This is currently harmless for singletons because every "second create" is rejected by the count guard before `persist` runs, and `get()` never supplies an id. Worth keeping in mind if singleton create is ever reachable with a caller id; no change requested for this PR.

### Recommendation

**APPROVE.** Behaviour-preserving for normal singleton usage, faithful to ADR-0004, well-tested, minimal blast radius.

---

Issue #477, opened 2026-06-03
image() multi-column mode: read/write existing Keystone columns (non-destructive parity)


borisno2:
## Parent

#476

## What to build

A multi-column mode for the `image()` field that maps onto an existing Keystone image layout (seven per-part columns) instead of a single `Json?` column, so a migrating project gets a functional image field over its live data with no destructive migration. End-to-end slice:

- Extend the field Prisma-type emission contract so a field can emit **multiple** column lines (the way relationship fields already emit FK + relation lines).
- A new **pure `metadata ↔ columns` mapper**: assemble the per-part columns (`<field>_url`, `<field>_width`, `<field>_height`, `<field>_filesize`, `<field>_contentType`, `<field>_contentDisposition`, `<field>_pathname`) into an `ImageMetadata` on read, and split a metadata value back into columns on write. Handles partially-populated/NULL columns (e.g. only `image_url`).
- `image()` gains a `db`-level multi-column mode with configurable per-part `@map` names. Default remains single-`Json?` (greenfield unchanged).
- Hooks: `resolveOutput` assembles columns → metadata; `resolveInput` splits metadata → columns and preserves the no-re-upload behaviour (an existing metadata value / populated columns is authoritative; only a `File`-like input uploads).

## Acceptance criteria

- [ ] A field can emit multiple Prisma columns through the field-emission contract (covered by a generator test).
- [ ] `image()` in multi-column mode generates the seven `@map`-ped scalar columns with Keystone naming; single-JSON mode is unchanged.
- [ ] Per-part column `@map` names are configurable.
- [ ] `metadata ↔ columns` mapper is a standalone pure module with unit tests: round-trip and partial/NULL-column cases.
- [ ] Reading a record assembles the columns into an `ImageMetadata`; a record with only `image_url` still yields a valid value.
- [ ] No-re-upload guarantee: an existing metadata value / populated columns never triggers a storage upload; a `File` input does.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

Issue #476, opened 2026-06-03
PRD: Non-destructive image/file field migration (multi-column parity over Keystone columns)


borisno2:
## Problem Statement

A team migrating a Keystone 6 project needs a working `image()`/`file()` field over their **existing** database without a destructive migration. Keystone stores an image across seven columns (`<field>_url`, `<field>_width`, `<field>_height`, `<field>_filesize`, `<field>_contentType`, `<field>_contentDisposition`, `<field>_pathname`) and a file across three (`<field>_filename`, `<field>_filesize`, `<field>_url`). The stack's `image()`/`file()` fields emit a single `Json?` column, so adopting them requires consolidating the columns into JSON and **dropping the originals** — which breaks **Schema parity** (the no-destructive-migration guardrail the whole migration program is gated on). The stack's own `migrate-image-fields` skill and `specs/keystone-image-migration.md` currently *recommend* this destructive consolidation, contradicting that guardrail. There was also uncertainty about whether Vercel Blob (vs S3-only) is supported and whether existing stored URLs can be read without re-upload.

## Solution

Give `image()`/`file()` a **multi-column mode** that maps onto the existing Keystone per-part columns in place (configurable `@map` names), assembling them into an `ImageMetadata`/`FileMetadata` on read and splitting back on write. A migrating project reaches a clean diff and a functional field with no data migration and no re-upload of existing assets. The single-`Json?` column stays the greenfield default; JSON consolidation becomes an explicitly-flagged, opt-in destructive alternative. Vercel Blob is documented as a first-class provider, and the no-re-upload guarantee is locked by a test. See ADR-0006.

## User Stories

1. As a migrating developer, I want `image()` to map onto my existing `image_url`/`image_width`/`image_height`/`image_filesize`/`image_contentType`/`image_contentDisposition`/`image_pathname` columns, so that the generated schema diffs clean against my live DB.
2. As a migrating developer, I want `file()` to map onto my existing `file_filename`/`file_filesize`/`file_url` columns, so that file fields migrate the same way.
3. As a migrating developer, I want the multi-column field to read existing stored URLs and present them as a normal image/file value, so that assets display without any re-upload or data migration.
4. As a migrating developer, I want the per-part column names to be configurable (`@map`), so that I can match non-default Keystone column names.
5. As a migrating developer, I want new uploads in multi-column mode to write back into the same column layout, so that the field is fully functional, not read-only.
[...]

PR #511, opened 2026-06-03
image()/file() multi-column mode: read/write existing Keystone columns (non-destructive parity)

borisno2:
Implements #477
Part of #476

## What this does

Adds a **non-destructive multi-column mode** to `image()` / `file()` so a project can adopt a pre-existing Keystone database without dropping columns (ADR-0006). The single-`Json?` column stays the greenfield default; multi-column mode is an opt-in per-field `db.columns: 'keystone'`.

```ts
import { image, file } from '@opensaas/stack-storage/fields'

fields: {
  // Maps onto the live image_url … image_pathname columns in place.
  avatar: image({ storage: 'images', db: { columns: 'keystone' } }),
  // Per-part @map names are configurable.
  cover: image({ storage: 'images', db: { columns: { mode: 'keystone', map: { url: 'cover_link' } } } }),
  resume: file({ storage: 'documents', db: { columns: 'keystone' } }),
}
```

## How it works

- **Multi-column emission** — new field-emission contract `getPrismaColumns` on `BaseFieldConfig` (mirrors how relationship fields emit FK + relation lines). The generator stays neutral: it places one `@map`-ped Prisma line per column it returns. No field-type switches in core/cli.
- **Pure `metadata ↔ columns` mapper** (`packages/storage/src/utils/multi-column.ts`) — `assemble*`/`split*` for image (7 cols) and file (3 cols), handling partial/NULL rows. Pure and unit-tested in isolation.
- **Assemble on read** — the read pipeline (`filterReadableFields`) assembles the per-part columns into the logical `ImageMetadata`/`FileMetadata` and **strips the raw columns** so they never leak. A row with only `image_url` still yields a valid value.
- **Split on write** — the write pipeline splits the logical value back across the per-part columns in place of the single key.
- **No-re-upload guarantee (both modes)** — an already-shaped metadata value / populated columns is authoritative and never triggers a storage upload; only a `File`-like input uploads. Locked by tests.

## Scope note / decision resolved

While wiring the runtime I found the storage fields' `resolveInput`/`afterOperation` hooks read non-existent `inputValue`/`fieldName` args instead of the Keystone-compliant `resolvedData[fieldKey]` / `originalItem`. That latent mismatch had to be corrected for the split/assemble (and the no-re-upload guarantee) to function through the real pipeline; it is now aligned to the documented field-hook contract.

The PRD's documentation/skill rewrite (`migrate-image-fields`, `specs/keystone-image-migration.md`) is a **separate slice of #476** and is intentionally out of scope here — this PR is the code slice (#477).

## Verification

- `pnpm build` — all packages + examples build.
- Tests: core 576, cli 275, storage 105 — all pass. New tests: pure mapper round-trip + partial/NULL; generator multi-column emission + single-column fallback; field-level no-re-upload (both modes) + emission; core generic read-assemble/write-split.
- `pnpm lint` (0 errors), `pnpm manypkg check` (valid), `pnpm format:check` (clean), `pnpm turbo run test` (32/32 tasks, examples unaffected).
- Changeset added (minor; core + cli + storage).

---

### DIFF
--- a/.changeset/mighty-otters-march.md
+++ b/.changeset/mighty-otters-march.md
@@ -1,0 +1,30 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+'@opensaas/stack-storage': minor
+---
+
+Add non-destructive multi-column mode to `image()` / `file()` for adopting an existing Keystone database without dropping columns (ADR-0006).
+
+Keystone stores an image across seven per-part columns (`_url`, `_width`, `_height`, `_filesize`, `_contentType`, `_contentDisposition`, `_pathname`) and a file across three (`_filename`, `_filesize`, `_url`). By default `image()`/`file()` still back a single `Json?` column (greenfield unchanged). Set `db.columns: 'keystone'` to map the field onto the existing per-part columns in place — assembled into an `ImageMetadata`/`FileMetadata` on read and split back on write — so a migrating project reaches a clean schema diff with no data migration and no re-upload of existing assets.
+
+```typescript
+import { image, file } from '@opensaas/stack-storage/fields'
+
+fields: {
+  // Maps onto image_url, image_width, … image_pathname in place.
+  avatar: image({ storage: 'images', db: { columns: 'keystone' } }),
+
+  // Per-part @map names are configurable for non-default column names.
+  cover: image({
+    storage: 'images',
+    db: { columns: { mode: 'keystone', map: { url: 'cover_link' } } },
+  }),
+
+  resume: file({ storage: 'documents', db: { columns: 'keystone' } }),
+}
+```
+
+No-re-upload guarantee (both modes): an already-shaped metadata value — or, in multi-column mode, populated columns — is authoritative and never triggers a storage upload; only a `File`-like input uploads.
+
+Adds a multi-column field-emission contract (`getPrismaColumns`) plus `getColumnNames`/`assembleColumns`/`splitColumns` to the field-authoring surface so any field can map onto several physical columns. The generator emits one `@map`-ped Prisma line per column; reads assemble the logical value from the raw columns and strip them from the result; writes split the logical value back across the columns.
--- a/.changeset/multi-column-write-access.md
+++ b/.changeset/multi-column-write-access.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix field-level write-access bypass for multi-column `image()`/`file()` fields. The per-part column split now respects the field's own `create`/`update` access (denied fields write none of their columns), matching single-column behaviour.
+
+Note the known lossy multi-column round-trip when assembling legacy Keystone columns: `originalFilename` collapses to `filename`, `uploadedAt` is `''`, and a NULL `contentType` reads back as `application/octet-stream`.
--- a/packages/cli/src/generator/prisma.test.ts
+++ b/packages/cli/src/generator/prisma.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect } from 'vitest'
 import { generatePrismaSchema, resolveListTimestamps } from './prisma.js'
-import type { OpenSaasConfig, ListConfig, DatabaseConfig } from '@opensaas/stack-core'
+import type { OpenSaasConfig, ListConfig, DatabaseConfig, FieldConfig } from '@opensaas/stack-core'
 import type { TypeInfo } from '@opensaas/stack-core/extend'
 import {
   text,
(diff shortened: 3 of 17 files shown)


borisno2 (commented):
## Review: image()/file() multi-column mode (#477 / ADR-0006)

**Verdict: APPROVE** (submitted as a Comment review because GitHub blocks a formal Approve when the reviewing identity authored the PR).

Reviewed strictly from the diff + surrounding integration code, against the issue acceptance criteria. Built core + storage (`tsc` clean) and ran the relevant suites locally: storage 105/105, core multi-column read/write 7/7, CLI multi-column generator 4/4. This is a solid, well-tested, non-destructive feature. One robustness gap and a couple of minor lossy-round-trip notes below, none of which block the adoption use case.

### Greenfield default — UNCHANGED (verified)
With no `db.columns`, `image()`/`file()` attach none of the multi-column methods (`getPrismaColumns`/`getColumnNames`/`assembleColumns`/`splitColumns` are `undefined`) and `getPrismaType` still returns `{ type: 'Json', modifiers: '?' }`. Locked by `multi-column-fields.test.ts` ("single-Json? default is unchanged"). The multi-column branches in core (`field-visibility.ts`, `hooks/index.ts`) and the generator are all gated on the presence of those methods, so a default field takes the identical path as before. No regression.

### Field-emission contract — clean, no switches (verified)
`getPrismaColumns`/`getColumnNames`/`assembleColumns`/`splitColumns` added to `BaseFieldConfig`; the generator (`prisma.ts`) and type generator (`types.ts`) delegate generically with zero field-type switches — mirroring the `getPrismaRelation` precedent. The generator correctly falls back to `getPrismaType` when `getPrismaColumns` returns `undefined`/empty (tested). Type generator omits the raw column names from the Prisma payload and re-adds the assembled logical field via the existing `resultExtension`/`TransformedFields` path, so `context.db` types expose `ImageMetadata | null` and hide raw columns.

### Assemble-on-read / split-on-write — correct, raw columns never leak (verified)
`filterReadableFields` assembles into a `workingItem` copy, then `delete`s the raw column keys before the per-field pass, so they cannot reach `filtered`. Partial rows (URL-only) and all-NULL rows are handled (`url`/`filename` is the presence signal → `null` when absent). The field-agnostic `multi-column-read-write.test.ts` explicitly asserts `'m_url' in result === false`, the partial case, the null case, and "field absent when columns unselected". Write-split removes the logical key and merges per-part columns; partial updates that omit the field don't touch its columns (guarded by `if (!(fieldKey in result)) continue`). 

### No-re-upload guarantee — locked in both modes (verified)
`isFileLike()` (duck-typed on `arrayBuffer`) is the *only* trigger for an upload; an already-shaped metadata value short-circuits with `return inputValue`. `multi-column-fields.test.ts` asserts no upload for existing metadata in **both** single-Json and multi-column modes, and that a `File`-like input *does* upload exactly once, for both `image()` and `file()`.

### Latent-bug fix to the storage hooks — genuine correction, contract-aligned (verified closely)
This was the riskiest part, so I checked it against the engine and the documented contract:
- The engine (`executeFieldResolveInputHooks`) has *always* passed `{ listKey, fieldKey, operation, inputData, item, resolvedData, context }` — it never passed `inputValue` or `fieldName`. Confirmed by `git show` of the pre-PR `hooks/index.ts`. So the old storage code read `inputValue` (always `undefined`) and returned early — the field-hook upload path and `cleanupOnReplace`/`cleanupOnDelete` (keyed on the never-passed `fieldName`) were effectively dead before this PR.
- The fix to `resolvedData[fieldKey]` / `fieldKey` matches `FieldResolveInputHookArgs` and the CLAUDE.md example verbatim (`const fieldValue = resolvedData[fieldKey]`).
- The `afterOperation` change `item`→`originalItem` for delete matches `FieldAfterOperationHookArgs` (delete variant carries `originalItem`, not `item`). The old `item[fieldName]` on delete was doubly broken.
- This is a behavior change for single-column users (the hook now actually runs), but the prior behavior stored nothing, so it is strictly a fix, not a regression. The added `typeof fileMetadata === 'object'` guard before `.filename` is a sensible hardening.

### Types — no public `any`/casts (verified)
New public types (`MultiColumnPrismaResult`, `ImageDbConfig`, `FileDbConfig`, the four method signatures) are strongly typed; `unknown` is used only internally. The only `: any` are the field-hook implementation params, each with the pre-existing `eslint-disable ... Field builder hooks are generic` annotation — identical to the established pattern in this file. Changeset present and correct (minor; core + cli + storage).

---

## Findings

### 1. (Medium — robustness gap) Per-part columns bypass field-level write access
Because `splitColumns` runs in the Hook Pipeline (Phase 1.5) *before* `filterWritableFields` (Phase 5), by the time write-access filtering runs the logical `image` key is gone and only the raw `image_url`/`image_width`/… keys remain. `filterWritableFields` looks those up in `fieldConfigs`, finds no config, and `checkFieldAccess(undefined, …)` returns `true` ("no field access means allow", `field-access.ts:39-41`). Net effect: a multi-column field with `access.update: () => false` (or a create deny) will **still be written** — the per-part columns slip through unfiltered. The current tests don't cover field-level access on a multi-column field, so this is untested. Single-column storage fields don't have this gap (their logical key is still present at Phase 5). For the pure Keystone-adoption use case this is unlikely to bite, but it's a real access-control hole worth either (a) re-binding the split columns back to the owning field's access during `filterWritableFields`, or (b) at minimum documenting that field-level write access is not enforced in multi-column mode, plus a test. Not a merge blocker for the adoption slice, but should be tracked.

### 2. (Low) Lossy write→read round-trips (acceptable for adoption, worth a doc note)
- `originalFilename` and `uploadedAt` have no Keystone columns: after a split+assemble round-trip, `originalFilename` collapses to `filename` (the pathname) and `uploadedAt` becomes `''`. Expected given the column layout, but consumers relying on those fields should know.
- `assembleImageMetadata` defaults `mimeType` to `'application/octet-stream'` when `contentType` is NULL, while `splitImageMetadata` writes `metadata.mimeType` straight back. An assemble-then-resave on a row with NULL `contentType` will populate the previously-NULL column with `application/octet-stream` (minor write-amplification, not a correctness issue).

### 3. (Nit) `imagePartFieldName`/`filePartFieldName` are trivial `map[part]` indirections
The Prisma model field name is set equal to the physical column (`name === map`), so these one-line helpers and the descriptor `name`/`map` duplication add a little ceremony. Fine as-is; only note it because it could be simplified if the name/map split is never expected to diverge.

Nothing here blocks merge. The greenfield path is provably untouched, the new behavior is well-tested end-to-end through the real pipeline, and the hook fix is correct. Recommend addressing finding #1 (or explicitly scoping it out) before relying on field-level access for multi-column fields.

---

borisno2:
## Security fix: multi-column writes now respect field-level write access

Pushed `2d056d2` to fix a field-level write-access bypass introduced by multi-column mode.

### The bug

The multi-column split runs during field `resolveInput` (Phase 1.5, in `executeFieldResolveInputHooks`): it removes the logical field key and replaces it with the raw per-part column keys. The canonical writable-field filter (`filterWritableFields`, Phase 5) then runs — but the logical key is already gone, and the raw per-part column keys are not in `listConfig.fields`, so `checkFieldAccess(undefined, …)` returns `true`. Result: a multi-column `image()`/`file()` field declared with `access: { update: () => false }` (or `create: () => false`) was **still written**, unlike its single-column equivalent.

### The fix (localized)

Gate the split on the field's own write access at the split point, using the canonical `checkFieldAccess` evaluator with the **same arguments the write pipeline uses** (`session`, `item`, `context`, `inputData`). If the field is not writable for the current operation, the logical key is dropped and **none** of its per-part columns are written — identical in effect to how `filterWritableFields` drops a denied single-column field. If writable, it splits as before.

- Single-column fields, multi-column fields without field-level access, writable fields, and sudo writes are all unchanged.
- `checkFieldAccess` already short-circuits under sudo, so sudo writes still split.

Touched: `packages/core/src/hooks/index.ts` (the split wiring + the access check it calls).

### Tests (lock the fix)

`packages/core/src/access/multi-column-read-write.test.ts` — "multi-column write split respects field-level write access":
- `access: { update: () => false }` on update → writes no per-part columns (and no logical key)
- `access: { create: () => false }` on create → writes no per-part columns
- granted access → still splits/writes the columns (regression)
- denying the *other* operation does not block the write (update-denied field still writable on create)
- sudo bypasses the gate and still splits
- a multi-column field *without* field-level access splits exactly as before

`packages/storage/tests/multi-column-fields.test.ts` — locks that the real `image()`/`file()` fields in multi-column mode surface the user-provided `access` alongside `splitColumns` (the two inputs the core gate needs), and that single-column fields still carry `access` and have no `splitColumns`.

### Verification

`pnpm build` green. Tests: core **582** (+6), storage **108** (+3), cli **275** — all pass. `pnpm manypkg check` valid, `pnpm lint` 0 errors, `pnpm format:check` clean. Patch changeset added (`@opensaas/stack-core`).

### Documented round-trip limitation

The legacy multi-column round-trip is intentionally lossy when assembling pre-existing Keystone columns (no first-class column backs these): `originalFilename` collapses to `filename`, `uploadedAt` reads back as `''`, and a NULL `contentType` reads back as `application/octet-stream`. Recorded in the changeset.

---

borisno2 (commented):
## Security re-review of the field-level write-access fix (head `2d056d2`)

**Verdict: APPROVE** (posted as a Comment review — the PR is authored by an integration identity, so a formal GitHub Approve is not available to me; treat this as an explicit approval of the security fix delta).

I reviewed the 2nd commit's delta against the canonical write pipeline, typechecked and ran the affected tests on the PR head, and confirm **the field-level write-access bypass is fully closed**.

### The bypass and the fix
In multi-column mode, `splitColumns` runs in `executeFieldResolveInputHooks` (Phase 1.5), dropping the logical field key and substituting raw per-part column keys that are absent from `fieldConfigs`. The canonical `filterWritableFields` (Phase 5) therefore could not enforce the field's own `access`, so a field declared `access: { update/create: () => false }` was still written. The fix gates the split on the field's own write access *before* substituting columns, so a denied field contributes none of its per-part columns — exactly matching how `filterWritableFields` drops a denied single-column field.

### Verification (all confirmed)

**1. Correct evaluator / args / operation** — The fix calls the canonical `checkFieldAccess(fieldConfig.access, operation, { session: context.session, item, context, inputData })`. This is the *same* evaluator `filterWritableFields` uses (`packages/core/src/access/field-access.ts`), with the *same* four args. `operation` is the `'create' | 'update'` value threaded from the pipeline (`hookPipeline.run` → `executeFieldResolveInputHooks`), and it matches the `writeOp` passed to `filterWritableFields` in `write-pipeline.ts`. `item` = `originalItem` and `inputData` = `input` line up with the Phase 5 call. Create-vs-update is handled correctly.

**2. Denied → nothing written; granted → splits** — On deny, the code drops the logical key and `continue`s, writing **no** per-part columns. On grant, it splits as before. Confirmed in code and tests.

**3. No new restriction / no regression** — Sudo bypasses inside `checkFieldAccess` (`_isSudo → true`); a field with no field-level access returns `true` (`!fieldAccess → true`); single-column / normal fields never enter the `splitColumns` branch and are unaffected. Confirmed.

**4. Tests genuinely lock the behaviour (not no-ops)** — `multi-column-read-write.test.ts` adds 6 cases asserting exact `toEqual` output plus negative `in` checks: update-denied → `{}` (no `m_url`/`m_size`), create-denied → `{}`, granted → `{ m_url, m_size }`, deny-update-but-create-op → splits (locks create-vs-update), sudo → splits, no-access → splits. `multi-column-fields.test.ts` adds 3 cases locking that the real `image()`/`file()` builders surface `access` *and* attach `splitColumns` (and that single-column mode has no `splitColumns`). I ran both on the head commit: **13/13 core and 15/15 storage pass.** These would fail against the pre-fix code.

**5. No new `any`/casts; types correct** — No casts or `any` in the new block. `tsc --noEmit` on `packages/core` passes against the head commit. Production `hooks/index.ts` needs no separate `FieldAccess` import because `fieldConfig.access` is already typed as `FieldAccess<…>` on `BaseFieldConfig` and is structurally assignable to `checkFieldAccess`'s `FieldAccess | undefined` parameter (same pattern `filterWritableFields` relies on). The test file's `FieldAccess` import (from `./types.js`) is the correct path.

**6. Reads / assemble-on-read untouched** — The fix commit modifies only `hooks/index.ts` (write split), two test files, and a changeset. `field-visibility.ts` (the read/assemble path via `assembleColumns` + `getColumnNames`) is **not** touched. Reads are not gated, and the raw-column stripping on read is unchanged.

### Informational (not blocking)
- **Upload side effect precedes the gate (parity, not a regression).** A storage field's `resolveInput` upload runs (line ~244) before the access gate (line ~263), so a denied field given a fresh `File` would still trigger a storage upload before its value is dropped. This is *identical* to single-column behaviour — there, the upload hook also runs in Phase 1.5 and `filterWritableFields` drops the value in Phase 5. The fix achieves exact parity and introduces no new discrepancy. If you ever want denied fields to also skip the upload, that's a separate, pre-existing architectural item (resolveInput side effects run before field write-access filtering) and out of scope for this security fix.
- Changeset present, `@opensaas/stack-core` patch, accurate description (incl. the documented lossy legacy round-trip).

**Bottom line: the access-control bypass is truly closed** — denied multi-column fields now write none of their columns, single-column / no-access / sudo paths are unchanged, reads are unaffected, and the new tests lock all of it. Ship it.

— Automated security re-review via Claude Code.

---

Issue #504, opened 2026-06-03
Multi-schema: native enum blocks need @@schema (else P1012 with db.schemas + enum select)


borisno2:
## Context

Found during review of #502 (auth schema placement, part of PRD #480). #482 adds `db.schemas` (datasource multi-schema) and per-model `@@schema`. The generator assigns `@@schema` to **models** but **not to native enum blocks**. Prisma's multi-schema mode requires *every model AND every enum* to declare an `@@schema`, or it errors (`P1012`).

So a config that combines `db.schemas` with a `select({ db: { type: 'enum' } })` field generates an **invalid** Prisma schema. The auth adoption path (#482's target) uses no enums, so it's unaffected — but this gap should be closed before `db.schemas` is relied on as a general multi-schema primitive.

## What to build

In multi-schema mode (`db.schemas` set), emit `@@schema("...")` on generated native `enum` blocks too — defaulting to `public` (or the owning model's schema / a configurable value). Add a provider guard is optional but nice (multi-schema is Postgres-only).

## Acceptance criteria

- [ ] With `db.schemas` set, every generated `enum` block declares an `@@schema(...)` (default `public`).
- [ ] A config combining `db.schemas` with a `select({ db: { type: 'enum' } })` field generates a valid multi-schema Prisma schema (no `P1012`).
- [ ] Greenfield (no `db.schemas`) output unchanged — no `@@schema` on enums.
- [ ] Generator test covering a multi-schema config that includes an enum-backed select.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

- #482 (introduces the `db.schemas` multi-schema machinery)

PR #512, opened 2026-06-03
fix(cli): emit @@schema on native enum blocks in multi-schema mode

borisno2:
Implements #504

Part of #480

## Problem

In multi-schema mode (`db.schemas` set), Prisma requires every model **and** every enum to declare an `@@schema(...)`, or it errors with `P1012`. The generator only emitted `@@schema` on models, so a config combining `db.schemas` with a `select({ db: { type: 'enum' } })` field produced an invalid Prisma schema.

## Fix

`packages/cli/src/generator/prisma.ts` now emits `@@schema(...)` on each generated native `enum` block when multi-schema mode is active. Each enum inherits the schema of its owning model (the list's `db.schema`), defaulting to `"public"`. Greenfield output (no `db.schemas`) is unchanged — no `@@schema` is emitted on enums.

## Tests

Added to `packages/cli/src/generator/prisma.test.ts`:
- A `db.schemas` + enum-backed `select` config emits a valid multi-schema schema where the enum block carries `@@schema("public")`.
- An enum inherits its owning model's `db.schema` (an `auth`-schema model's enum lands in `auth`).
- Greenfield (no `db.schemas`) emits no `@@schema` on enums.

## Verification

- `pnpm build` — pass
- `packages/cli` tests — 274 pass
- `packages/core` tests — 569 pass
- `pnpm lint` — 0 errors
- `pnpm manypkg check` — pass
- `pnpm format:check` — pass
- `pnpm turbo run test` (incl. example builds) — 32/32 pass

Changeset: `.changeset/quiet-enums-settle.md` (patch, `@opensaas/stack-cli`).

---

### DIFF
--- a/.changeset/quiet-enums-settle.md
+++ b/.changeset/quiet-enums-settle.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+Fix multi-schema mode (db.schemas): emit @@schema on generated native enum blocks so an enum-backed select() no longer produces an invalid schema (P1012). Enums inherit their owning model's schema, defaulting to public; greenfield output is unchanged.
--- a/packages/cli/src/generator/prisma.test.ts
+++ b/packages/cli/src/generator/prisma.test.ts
@@ -2062,6 +2062,107 @@ describe('Prisma Schema Generator', () => {
       expect(schema).toContain('enum PostStatus {')
     })
 
+    it('should declare @@schema on generated enum blocks in multi-schema mode (P1012 fix)', () => {
+      // Multi-schema mode (db.schemas) requires every model AND every enum to
+      // declare an @@schema, or Prisma rejects the schema with P1012. Before this
+      // fix only models carried @@schema, so a db.schemas + enum-select config
+      // produced an invalid schema.
+      const config: OpenSaasConfig = {
+        db: {
+          provider: 'postgresql',
+          schemas: ['public', 'auth'],
+        },
+        lists: {
+          Post: {
+            fields: {
+              title: text(),
+              status: select({
+                options: [
+                  { label: 'Draft', value: 'draft' },
+                  { label: 'Published', value: 'published' },
+                ],
+                db: { type: 'enum' },
+              }),
+            },
+          },
+        },
+      }
+
+      const schema = generatePrismaSchema(config)
+
+      // Datasource is multi-schema
+      expect(schema).toContain('previewFeatures = ["multiSchema"]')
+      expect(schema).toContain('schemas  = ["public", "auth"]')
+
+      // The enum block exists and carries @@schema (default public — the list
+      // declares no schema of its own). This is the fix: previously the enum had
+      // no @@schema, which Prisma rejects with P1012 in multi-schema mode.
+      const enumBlock = schema.slice(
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review — PR #512 (fix: emit `@@schema` on native enum blocks in multi-schema mode)

**Verdict: APPROVE** (posted as a Comment review since the PR is authored by an automation identity; treat this as an explicit approval.)

Reviewed strictly from the diff plus the surrounding `prisma.ts` generator and the `select()`/enum-naming logic in `packages/core/src/fields/index.ts`. The change correctly closes the #504 gap and all acceptance criteria are satisfied.

### What it does (correctly)
- `enumDefinitions` is widened from `Map<string, string[]>` to `Map<string, { values: string[]; schema: string }>`, recording each enum's owning schema (`listConfig.db?.schema ?? 'public'`) at collection time.
- The enum-emit loop appends `@@schema("<schema>")` only when `multiSchema` is true. `multiSchema` is the same flag already used for the datasource `schemas = [...]` array and per-model `@@schema`, so the gating is consistent with the existing multi-schema machinery.
- Greenfield path is genuinely untouched: the `@@schema` line is inside `if (multiSchema)`, and the greenfield test asserts `schema).not.toContain('@@schema')` over the whole output.

### Acceptance criteria — all met
- Every enum block declares `@@schema(...)` in multi-schema mode, defaulting to `"public"`. ✓
- `db.schemas` + `select({ db: { type: 'enum' } })` now produces a valid multi-schema schema (test asserts datasource multi-schema markers AND `@@schema("public")` scoped to the `PostStatus` block). ✓
- Enum-to-owning-model mapping is correct: the `AuthUser`/`db: { schema: 'auth' }` test confirms `AuthUserRole` lands in `auth`, not `public`. ✓
- Greenfield output unchanged. ✓
- Generator tests added; changeset (`@opensaas/stack-cli`, patch) present and accurate. ✓

### Code quality
- No `any`, no type casts, no field-type `switch`. Enum detection stays delegated to the field's `getPrismaType()`/`result.enumValues` (consistent with the self-contained-field principle). The new `EnumDefinition` type is properly structured. ✓
- Comments are accurate and explain the P1012 rationale.

### Observations (non-blocking)
1. **Shared enums via explicit `db.enumName` across schemas.** `enumDefinitions` keys on `result.type` (the enum name). The derived name (`listName + field`) is always unique per list, so the common case is safe. But if two lists in *different* schemas declare `select` fields with the same explicit `db.enumName`, the map's last-write-wins means the enum's schema (and values) is taken from whichever list is iterated last. This is an inherent limitation — Prisma can't place one enum in two schemas — and is pre-existing for the values collision, not introduced here. Worth a follow-up validation error if shared-enum-name usage is expected; not required for this fix.
2. **Out of scope, but adjacent:** a model in multi-schema mode that declares *no* `db.schema` emits no `@@schema` (the model loop only emits when `listConfig.db?.schema` is set), which would itself trigger P1012. This PR makes enums *more* robust than models by defaulting to `public`. Aligning the model default to `public` would be a sensible companion change, but it's outside #504's scope.

No required changes. Good, tightly-scoped fix with proportionate test coverage.

🤖 Reviewed via /review

---

Issue #513, opened 2026-06-03
Multi-schema: models without db.schema should default to @@schema("public") (P1012 footgun)


borisno2:
## Context

Found while implementing #504 (which fixed the same class of bug for native enums). In multi-schema mode (`DatabaseConfig.schemas` set), Prisma requires every **model** and every **enum** to declare `@@schema(...)` or it errors with `P1012`. The generator (from #482) only emits `@@schema` on a model when that list has an explicit `db.schema`. So a `db.schemas` config that includes a list WITHOUT `db.schema` produces a model with no `@@schema` → invalid schema.

The auth plugin's `beforeGenerate` hook (#482) defaults its own un-schema'd lists to `public`, so the auth adoption path is fine — but a hand-written `db.schemas` config with a plain list hits this. #504 already does the right thing for enums (default `@@schema("public")`); models should mirror it for consistency.

## What to build

In multi-schema mode (`db.schemas` non-empty), every generated `model` should declare an `@@schema(...)` — using the list's `db.schema` if set, else defaulting to `"public"` (mirroring the enum behaviour added in #504). Greenfield (no `db.schemas`) output unchanged.

## Acceptance criteria

- [ ] With `db.schemas` set, a list WITHOUT `db.schema` emits `@@schema("public")` (no `P1012`).
- [ ] A list WITH `db.schema` is unchanged (uses its schema).
- [ ] Greenfield (no `db.schemas`) emits no `@@schema` on models.
- [ ] Generator test covering a multi-schema config with a mix of schema'd and un-schema'd lists.
- [ ] Changeset added (`@opensaas/stack-cli`, patch).

## Blocked by

None - #482 (multi-schema) and #504 (enum default) are already merged.

PR #519, opened 2026-06-04
Multi-schema: default un-schema'd models to @@schema("public") (fixes P1012)

borisno2:
Implements #513

---

### DIFF
--- a/.changeset/proud-otters-shine.md
+++ b/.changeset/proud-otters-shine.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+Fix multi-schema P1012: models without `db.schema` now default to `@@schema("public")` instead of emitting no `@@schema` (mirrors the enum default). Greenfield output is unchanged.
--- a/packages/cli/src/generator/prisma.test.ts
+++ b/packages/cli/src/generator/prisma.test.ts
@@ -1124,6 +1124,75 @@ describe('Prisma Schema Generator', () => {
       expect(schema).not.toContain('@@schema')
     })
 
+    it('defaults un-schemad models to @@schema("public") in multi-schema mode (mix of schemad and un-schemad lists)', () => {
+      // In multi-schema mode (db.schemas set) Prisma requires every model to
+      // declare an @@schema or it errors with P1012. A list WITH db.schema uses
+      // its own schema; a list WITHOUT db.schema must default to "public"
+      // (mirroring the enum default from #504). See issue #513.
+      const config: OpenSaasConfig = {
+        db: {
+          provider: 'postgresql',
+          schemas: ['public', 'auth'],
+        },
+        lists: {
+          // Explicit schema → used as-is
+          AuthUser: {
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review — PR #519 (implements #513)

**Verdict: APPROVE** (posted as a Comment review because GitHub blocks self-approval — the authenticated identity authored this PR).

### What it does
In Postgres multi-schema mode, every model must carry an `@@schema(...)` or Prisma errors with `P1012`. Previously the model `@@schema` line was gated on `listConfig.db?.schema` being set, so a list with no explicit schema emitted no `@@schema` → P1012. This switches the gate to the existing `multiSchema` flag with a `'public'` fallback, mirroring exactly what #504 already did for enums.

### Correctness — verified against the full file
- **Every model emits `@@schema` in multi-schema mode.** New gate is `if (multiSchema)` → `@@schema("${listConfig.db?.schema ?? 'public'}")`. A list with `db.schema` uses it; a list without defaults to `public`. ✓
- **Greenfield unchanged.** `multiSchema = Array.isArray(schemas) && schemas.length > 0`. With no `db.schemas`, the flag is `false` and no `@@schema` is emitted on models — identical to before. ✓
- **Consistent with #504.** The enum block uses the same `multiSchema` gate and the same `?? 'public'` fallback. The model change is now structurally identical (the enum default derives its schema from the owning list, which is the equivalent semantics). ✓
- **Scope is tight.** Only the model `@@schema` gate changed; no other generator behaviour touched (id/timestamps/relations/`@@map` all untouched). ✓
- **Style.** No `any`, no casts, no field-type switches introduced. `listConfig.db?.schema ?? 'public'` is clean optional-chaining. ✓
- Comment block accurately documents the rationale and the #504 mirror.

### Tests — genuinely assert the contract
- Mixed-config test (`AuthUser` with `db.schema: 'auth'` + `Post` with none): asserts AuthUser keeps `@@schema("auth")`, Post defaults to `@@schema("public")`, and crucially that `modelCount === modelSchemaCount` (2 == 2) — i.e. no model is left without an `@@schema`, which is the actual P1012 footgun being closed. The mixed config has no enums/relationships, so the `@@schema` count isn't inflated and the count equality is meaningful. ✓
- Greenfield test asserts `model Post {` is emitted and the whole schema has no `@@schema` (valid since there are no enums/schemad lists). ✓
- Changeset present (`@opensaas/stack-cli: patch`) with an accurate description. ✓

### Minor / non-blocking observations
- The block-slicing in the mixed test relies on `config.lists` insertion order (AuthUser before Post) to bound the `authUserBlock` slice. That holds (generator iterates `Object.entries` in declaration order), so it's correct — just slightly implicit. The `modelCount === modelSchemaCount` assertion is the load-bearing one and is order-independent, so this is fine as-is.
- The `modelSchemaCount` regex `^\s*@@schema\(` would also match enum `@@schema` lines if an enum existed in the config; here there are none, so it's accurate. Not worth changing.

No required changes. Ship it.

---

Issue #490, opened 2026-06-03
Comprehensive canonical Keystone→stack migration guide (umbrella; reflects A–E)


borisno2:
## Parent

#486

> **HITL — human review/approval required before merge.** This is the public face of migration and spans every area; an agent drafts it but a human reviews and approves.

## What to build

Promote a single canonical, published Keystone→stack guide as the source of truth, folding in the `specs/keystone-migration.md` recipes and reflecting the behaviour delivered across Areas A–E. Reduce `specs/keystone-migration.md` to a design-notes pointer.

The guide must cover and stay consistent with:
- Field-type mapping, hook mapping (Keystone timing semantics vs stack), access mapping.
- `context.graphql.run` → `context.db.*` (links to the hardened skill recipes from #488).
- **Generator parity** (Area A): Keystone-compat mode (empty-string text defaults), auto-timestamps off by default + `db.timestamps`, bare singleton `id`, `select` `db.isNullable`/`db.enumName`, honoured `defaultValue`.
- **Image/file migration** (Area C): link to the non-destructive multi-column recipe (#479) — do not duplicate.
- **Auth adoption** (Area D): link to the adopt-existing-better-auth-tables recipe (#483) — do not duplicate.
- **Configurable output paths** (Area E, #485) for side-by-side coexistence.
- Multi-schema DB preservation and admin-UI parity notes.

## Acceptance criteria

- [ ] One canonical published guide is the source of truth; `specs/keystone-migration.md` points to it.
- [ ] The guide reflects the shipped Area A–E behaviour (no stale pre-change instructions) and links (not duplicates) the #479 image and #483 auth sections.
- [ ] Field-type, hook, and access mapping tables are present and correct.
- [ ] Docs link-check passes (no 404s; internal links resolve).
- [ ] Human reviewer has approved before merge.

## Blocked by

- #487
- #471
- #472
- #473
- #474
- #475
- #479
- #483
- #485

borisno2:
[...]

Issue #485, opened 2026-06-03
Configurable generator output paths: output config block + path resolver


borisno2:
## Parent

#484

## What to build

Let a project relocate the generated Prisma schema and the `.opensaas` output directory via an `output` block in `opensaas.config.ts`, so `opensaas generate` can run alongside an existing Keystone `prisma/` without clobbering it. End-to-end slice:

- `output` config block: `output.prismaSchema` (schema file path) and `output.opensaasDir` (the `.opensaas` bundle directory). Defaults: `prisma/schema.prisma` and `.opensaas/`.
- New **output-path resolver** module: takes the `output` config + defaults and returns the resolved write paths AND the relative cross-references between generated files. `generate.ts` consumes it instead of hardcoding the paths (currently L109-115).
- Cross-references follow the configured locations: `context.ts` imports the generated types/lists from the resolved `.opensaas` dir, and the generated `prisma.config.ts` (kept at top level) points at the configured schema path so `prisma` CLI commands work.

## Acceptance criteria

- [ ] Setting `output.prismaSchema` writes the schema to the configured path (e.g. `prisma-opensaas/schema.prisma`); `output.opensaasDir` writes the `.opensaas` files to the configured dir.
- [ ] With no `output` block, defaults are unchanged (`prisma/schema.prisma`, `.opensaas/`).
- [ ] The generated `prisma.config.ts` references the configured schema path.
- [ ] The output-path resolver is a standalone module with unit tests (defaults and custom paths → resolved paths + cross-references).
- [ ] A generate-into-non-default-paths test asserts files land in the configured locations AND `context.ts` imports + the `prisma.config.ts` schema path resolve.
- [ ] Changeset added (see `pr-changeset` skill).

## Blocked by

None - can start immediately

PR #522, opened 2026-06-04
docs: canonical Keystone → stack migration guide (#490)

borisno2:
Implements #490
Part of #486

## HITL — holds for human review/approval before merge

This is the public face of migration and spans every area, so per #490 it is **human-in-the-loop**: an agent drafted it, but the EM reviews and approves before merge. **Do not merge without human approval.**

## What changed (docs-only)

Promotes a single canonical, published Keystone → stack migration guide as the source of truth, folding in the `specs/keystone-migration.md` recipes and reflecting the shipped Area A–E behaviour.

- **`docs/content/guides/migrating-from-keystone.md`** — expanded from a thin landing page into the comprehensive canonical guide. Covers: overview, config translation, **field-type mapping table**, **generator parity (Area A)** — `db.keystoneCompat` empty-string text defaults, auto-timestamps OFF by default + `db.timestamps` opt-in (incl. per-list override + `P1012` dedupe), bare singleton `id Int @id`, `select` `db.type`/`db.enumName`/`db.isNullable`, honoured `defaultValue`; **access mapping**; **hook mapping** with the real list/field pipeline timing; `context.graphql.run` → `context.db.*` summary that **links** the hardened `migrate-context-calls` skill recipes (#488); **image/file (Area C)** links the non-destructive multi-column recipe (#479); **auth adoption (Area D)** links the `adoptBetterAuthTables()` recipe (#483); **configurable output paths (Area E, #485)**; multi-schema preservation + admin-UI parity; checklist; CLI + plugin tooling.
- **`specs/keystone-migration.md`** — reduced to a short design-notes pointer at the published guide, with a table mapping each topic (A–E) to where it is now maintained.
- **`docs/content/guides/migration.md`** — trimmed the duplicated `context.graphql.run` recipes and the partial KeystoneJS field table to short pointers at the canonical guide (no stale duplicates); kept it as the general multi-source AI-assisted walkthrough.
- **`docs/lib/navigation.ts`** — made "Migrating from Keystone" the first migration entry; relabelled the general one "Migration Guide (all sources)".

## Source of truth & link-only sections

- **One canonical page:** `guides/migrating-from-keystone.md`. The CLI (`MIGRATION_GUIDE_URL`) already points at it.
- Image (#479) and auth (#483) detail are **linked, not duplicated**, per the issue.

## API verification

Every config/API snippet was cross-checked against shipped code on `main`:

- Field/db options vs `packages/core/src/config/types.ts` and `packages/core/src/fields/index.ts` (`db.keystoneCompat`, `db.timestamps` global + per-list, `select` `db.type`/`db.enumName`/`db.isNullable`, `db.schemas`/`db.schema`/`db.map`, `OutputConfig`).
- Query API vs `packages/core/src/query/index.ts` (`defineFragment`, `runQuery`/`runQueryOne`, `ResultOf`, `RelationSelector`).
- Auth `adoptBetterAuthTables()` vs `packages/auth` and the auth guide's adoption section.
- Image multi-column `db.columns: 'keystone'` vs `specs/keystone-image-migration.md` + ADR-0006.
- `migrate-context-calls` recipe summaries vs the skill's `SKILL.md`.
- Output paths vs `packages/cli/src/generator/output-paths.ts`.

## Checks

- `pnpm install --prefer-offline` — clean
- docs **link-check** (`pnpm --filter opensaas-stack-docs link-check`) — **pass** (5 required pages resolve, internal links in 4 migration docs valid)
- docs **build** (`next build`, runs link-check prebuild) — **pass** (all pages generated, incl. `/docs/guides/migrating-from-keystone`)
- `pnpm lint` — **pass** (0 errors; only pre-existing warnings in untouched files)
- `pnpm format:check` — **pass**

No changeset/plugin-version bump: docs-only, no `packages/*` or `claude-plugins/*` changes.

## For the reviewer to focus on

- **Source-of-truth call:** I extended the existing `migrating-from-keystone.md` as canonical (rather than restructuring around `migration.md`) because the CLI and link-check already treat it as canonical. Confirm that's the intended page.
- **`migration.md` trimming:** I kept the `## Migrating context.graphql.run` heading (now a short pointer) so `core-concepts/queries.md`'s existing `#migrating-contextgraphqlrun` anchor link still resolves to a real section. Confirm you're happy keeping that anchor vs removing the section entirely.
- **External links to `specs/` and `docs/adr/`:** these point to GitHub `blob/main/...` URLs (they aren't docs-site pages, so the link-check doesn't verify them). Worth a sanity check that those paths are the intended targets.
- **`bigInt()` mapping:** there is no dedicated `bigInt()` builder, so I documented `integer({ db: { nativeType: 'BigInt' } })` as the mapping — please confirm that's the recommended approach.

---


borisno2 (commented):
## Review — canonical Keystone → stack migration guide (#490)

**Verdict: REQUEST CHANGES** (posted as a Comment review because GitHub blocks a formal Approve/Request-changes when the reviewing identity authored the PR). One factual defect (the `bigInt()` mapping) should be fixed before merge; everything else is accurate and the consolidation is sound.

I cross-checked every config/API snippet against the shipped code on `main` (commit `c97fb46`). The vast majority verified clean. Findings below, ordered by severity.

---

### 🔴 Must fix

**1. `bigInt()` → `integer({ db: { nativeType: 'BigInt' } })` is incorrect (field-type table §2).**
This is the author's open question #4, and the answer is: **no, this is not the recommended mapping and as written it produces an invalid schema.**

- `integer().getPrismaType()` (`packages/core/src/fields/index.ts:192-225`) returns Prisma scalar type `Int` and, when `db.nativeType` is set, appends `@db.<nativeType>` → so this emits `someField Int? @db.BigInt`.
- `@db.BigInt` is **not a valid native-type attribute for the Prisma `Int` scalar**. Prisma models BigInt as a *scalar type* (`BigInt`, which maps to JS `bigint`), not as a `@db.` annotation on `Int`. `Int @db.BigInt` will fail `prisma validate` / `generate` (P1012) on Postgres/MySQL/SQLite.
- Even if it generated, `integer().getTypeScriptType()` always returns `type: 'number'` (line 230), so you'd silently lose `bigint` precision in TS.
- The stack's **own** migration tooling does not treat this as the BigInt path: `packages/cli/src/migration/introspectors/prisma-introspector.ts:203` maps Prisma `BigInt` → `text()` with the comment `// No native support`, and `migration-generator.ts:527` special-cases `BigInt`/`Decimal`/`Bytes`. There is no `bigInt()` builder.

Recommendation: don't present `integer({ db: { nativeType: 'BigInt' } })` as the mapping. Either (a) state that there is no native BigInt field and the tool maps it to `text()` (matching the introspector), and/or (b) recommend `db.extendPrismaSchema` to emit a real `BigInt` scalar column if the live DB has one. As written it will mislead migrators into an un-generatable schema.

---

### 🟡 Worth addressing

**2. `float()` → `decimal()` mapping changes the column type (field-type table §2).**
Keystone's `float()` is a `Float` (double) column; `decimal()` emits `Decimal @db.Decimal(p,s)` (`fields/index.ts:342-378`). Mapping `float → decimal` is a **destructive column-type change**, which cuts against the guide's central "Schema parity / no destructive changes" promise. There is no `float()` builder in the stack, so some workaround is unavoidable — but the table should flag that `Float → Decimal` is a type change (consider `decimal` *or* `extendPrismaSchema`/`nativeType` to keep a real float), not present it as a clean same-shape mapping like the rows above it.

---

### 🟢 Verified accurate (no change needed)

Spot-checked and confirmed against source:

- **Generator parity (Area A):** `db.keystoneCompat` (empty-string non-null text default, opt-in, explicit `defaultValue` wins, nullable/non-text untouched) — `config/types.ts:1582`, `fields/index.ts:120-129`. `db.timestamps` global + per-list override + `P1012` dedupe — `types.ts:1369,1554`. Bare singleton `id Int @id` — matches. `select` `db.type`/`db.enumName` (default `<List><Field>`)/`db.isNullable` semantics and the `→ status AccountNoteStatusType? @default(open)` example output — `types.ts:643-700`, `fields/index.ts:854-915`. Honoured `defaultValue` → `@default(...)` on text/integer/json/checkbox/decimal/timestamp — confirmed.
- **Query API (Area B):** `defineFragment` / `runQuery` / `runQueryOne` / `ResultOf` / `RelationSelector` all exist with the documented signatures and the nested `{ query, where, orderBy, take, skip }` shape (`query/index.ts`), and **are exported from the root `@opensaas/stack-core`** (`src/index.ts`) — so the guide's `import { defineFragment, type ResultOf } from '@opensaas/stack-core'` is correct.
- **Output paths (Area E):** `OutputConfig.prismaSchema` (default `prisma/schema.prisma`) and `opensaasDir` (default `.opensaas`) match `types.ts:2087`; exported from root.
- **Auth (Area D) & image/file (Area C):** correctly **linked, not duplicated**. `adoptBetterAuthTables()` exists (`packages/auth/src/config/adopt-better-auth-tables.ts`). Image/file use `@opensaas/stack-storage/fields` with `db.columns: 'keystone'` (`storage/src/fields/index.ts`). The "7 image columns / 3 file columns" claim and the `avatar_url`/`resume_filename`… example names match `IMAGE_COLUMN_PARTS` / `FILE_COLUMN_PARTS` exactly (`storage/src/utils/multi-column.ts`).
- **Mapping tables:** Access §4 — `item` is provided to operation access on update/delete (`access/types.ts:270,304`), so the `access.item.*` row is right. Hook §7 — the **read pipeline (3 steps: DB → field access control → field `resolveOutput`) is correct**; reads do not run `afterOperation` (`context/index.ts` findUnique/findMany only filter fields + resolveOutput). Note this means the repo's root `CLAUDE.md` read-order list (which adds a 4th "field afterOperation" step) is the stale one — the guide is right.
- **Consolidation soundness:** single canonical page; `migration.md` trimmed to pointers while **preserving the `## Migrating context.graphql.run` heading (line 877)**, so `core-concepts/queries.md`'s two `#migrating-contextgraphqlrun` anchors (lines 14, 178) still resolve. `specs/keystone-migration.md` reduced to a pointer + topic map; nav updated.
- **Tooling:** CLI `MIGRATION_GUIDE_URL` === `https://stack.opensaas.au/docs/guides/migrating-from-keystone` (`commands/migrate.ts:23`, asserted in tests), `--with-ai`/`--type keystone` are valid flags, and `/plugin install opensaas-migration@opensaas-stack-marketplace` matches the marketplace (`.claude-plugin/marketplace.json`). The `migrate-context-calls` Recipe 1–4 summaries (where-shape, connect/disconnect/`set: []`, "never write scalar FK — silently stripped", gql.tada/`VariablesOf`→factory, fragment→include + null-on-access-denied) all faithfully match the skill's `SKILL.md`.
- **Links:** all four ADR `blob/main` URLs (0004–0007) and both `specs/*.md` targets exist on disk.

---

### Recommendations on the four flagged open questions

1. **Is `migrating-from-keystone.md` the right canonical page?** — **Yes.** The CLI's `MIGRATION_GUIDE_URL` already points at exactly this page (and a test locks the value), so extending it rather than restructuring around `migration.md` is the correct call.
2. **Keep vs delete the `## Migrating context.graphql.run` pointer section in `migration.md`?** — **Keep it.** `core-concepts/queries.md` links into `#migrating-contextgraphqlrun` in two places; the heading must survive or those anchors 404. Leaving it as a short pointer is the right move.
3. **GitHub `blob/main` URLs to `specs/` and `docs/adr/`.** — **Fine to keep.** All targets exist and these aren't docs-site pages, so blob URLs are appropriate (link-check correctly doesn't validate them). Minor optional nit: `blob/main` URLs aren't version-pinned, so they can drift if a file is renamed — acceptable for design-notes/ADR references.
4. **`bigInt()` mapping.** — **Not correct as written** (see 🔴 #1). Needs revision before this ships as the public source of truth.

---

**Summary:** Strong, accurate, well-consolidated guide — the Area A–E technical content, query API, links, and tooling references all check out against shipped code. The single blocking item is the `bigInt()` field mapping (#1), which would lead a migrator to a schema Prisma rejects; the `float()→decimal()` row (#2) should also be caveated. Fix those two and this is good to merge. Holding for human approval per the HITL note.

---

borisno2:
Follow-up accuracy fix pushed to `claude/issue-490-canonical-migration-guide` (`2808ee8`) — field-type mapping table only, no restructuring, still HITL/no merge.

**`bigInt()` correction (was wrong).** The previous row documented `bigInt()` → `integer({ db: { nativeType: 'BigInt' } })`. That's invalid: per `packages/core/src/fields/index.ts` (the `db.nativeType` branch around L204-207, locked by the test at `packages/core/tests/field-types.test.ts:299-303`) it emits `Int @db.BigInt`, a combination Prisma rejects — `@db.BigInt` is not a valid native-type attribute for the `Int` scalar — and the TS type stays `number` regardless.

I grounded the corrected mapping in the migration introspector: `packages/cli/src/migration/introspectors/prisma-introspector.ts:203` maps Prisma `BigInt` → `text()` with a `// No native support` comment, and `getWarnings()` (L220-223, mirrored in `migration-generator.ts:527-530`) emits `uses unsupported type "BigInt" - will be mapped to text()`. There is no `bigInt()` builder anywhere in `packages/`. The row now reads `bigInt()` → `text()`, states there is **no native BigInt field**, notes values are stored/returned as strings (exact preservation for IDs but no numeric operators), and explicitly warns against the invalid `integer({ db: { nativeType: 'BigInt' } })` snippet.

**`float()` caveat.** There is no `float()` builder, and `decimal()` (`packages/core/src/fields/index.ts:280`) emits `Decimal @db.Decimal(precision, scale)` (defaults `18, 4`) with a `decimal.js` `Decimal` runtime type. The row now flags `float()` → `decimal()` as a **type change**, not 1:1 parity (column `Float`→`Decimal`, runtime `number`→`Decimal`), and tells you to size `precision`/`scale` for your existing values and review rounding/precision differences.

This also supersedes the open "please confirm that's the recommended approach" question in the PR description for `bigInt()` — the answer is no, and the row is fixed accordingly.

**Checks (re-run on this branch after the change):**
- `pnpm install --prefer-offline` — clean
- docs link-check (`pnpm --filter opensaas-stack-docs link-check`) — pass (5 required pages, internal links in 4 migration docs valid)
- `pnpm build` (incl. docs `next build` + link-check prebuild) — pass (11/11 tasks, all 34 docs pages generated)
- `pnpm lint` — pass (0 errors; only pre-existing warnings in untouched files)
- `pnpm format:check` — pass

Note: `format:check` was actually failing on the prior PR head (`3a01cfb`) for both `docs/content/guides/migrating-from-keystone.md` and `specs/keystone-migration.md` (misaligned markdown tables). I ran `prettier --write` on both, which is why the diff includes whitespace-only table-alignment churn beyond the two row edits — no other content changed. Docs-only, no changeset/plugin-version bump needed.

---

Issue #581, opened 2026-06-27
[PRD] Access-control fidelity: Keystone fail-loud parity across read/write/nested pipelines


borisno2:
> *This was generated by AI during triage.*

## PRD — Access-control fidelity & Keystone fail-loud parity

**Goal:** Bring the access-control engine, read/write pipeline, and nested-operations to Keystone-parity: **fail loud on writes** where the stack currently fails quiet, close read/connect **access bypasses**, and tighten lookup semantics. Surfaced across a production Keystone→stack migration access-parity audit.

**Shared policy (decided during triage):** adopt Keystone fail-loud — writes throw on access/shape violations; reads keep the silent-failure (filter/empty) contract; **`sudo` is the single trusted bypass** (no new per-call opt-out flags).

**Why these are grouped:** they all touch the same surfaces — the writable-field filter / write pipeline, the context read paths (`findUnique`/`findMany` include building), and `nested-operations.ts`. Working them independently would collide; sequencing avoids rework.

### Child issues
**Write side (fail-loud):**
- **#564** (bug, ready-for-agent) — undeclared data keys fail OPEN → reject for non-sudo writes.
- **#568** (bug, ready-for-agent) — field-access-denied keys silently stripped → throw on caller-supplied denied fields.

**Read side:**
- **#566** (bug, ready-for-agent, HIGH) — explicit `include`/fragment replaces the access-controlled include → merge, not replace (row-level bypass).
- **#565** (enhancement, ready-for-agent) — add `findFirst` to the delegates (prerequisite for #567).
- **#567** (bug, ready-for-agent) — `findUnique` accepts a non-unique `where` → enforce unique-where (depends on #565 for the non-unique path).

**Nested operations:**
- **#578** (bug, ready-for-agent, HIGH) — nested-connect false denial + wrong access phase → read-access + DB-reachability check (absorbs #569's connect half).
- **#569** (bug, ready-for-human) — nested writes skip before/afterOperation hooks → run the full pipeline (architectural).

### Suggested sequencing
1. **#564 + #568** together (write-pipeline fail-loud — same `filterWritableFields` area).
2. **#565 → #567**, and **#566** (context read paths).
3. **#578** (nested-connect; contained), then **#569** (nested full-pipeline; human, larger).

### Definition of done
- A Keystone-parity access test suite passes: no silent fail-open on writes; no read or connect access bypass; unique-`where` enforced; nested writes run the full hook pipeline; nested connect uses read access + DB reachability.
[...]

Issue #568, opened 2026-06-10
Write pipeline silently strips fields denied by field access instead of erroring (Keystone threw)


borisno2:
## Summary

On create/update, `filterWritableFields` in `@opensaas/stack-core@0.24.0` silently DROPS any `data` key the session's field access denies, then proceeds with the rest of the write. Keystone threw a GraphQL access error for the same situation.

Two consequences for migrating apps:

1. A formerly-failing call now "succeeds" while doing less than asked — the caller cannot observe the denial (no error, no result marker).
2. When the stripped field is also a hook trigger (e.g. a gated `status` whose transition fires an `afterOperation` workflow), the side-effect chain is silently skipped rather than loudly refused — the worst failure mode for billing/notification triggers.

## Ask

Throw on denied fields by default (Keystone parity), or provide a strict mode / per-call option — and either way document the strip behaviour prominently in the access-control docs. Silent partial writes are the kind of behaviour an agent-built app can't debug.

## Context

Found during a semantic-fidelity audit of a Keystone→stack migration. Related observations from the same audit: #564 (undeclared keys fail open), #566 (explicit include bypasses relation filters) — together these make the write/read pipeline fail quiet where Keystone failed loud.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** On create/update, declared fields whose field-level write access denies the caller are silently dropped and the write proceeds. Adopt Keystone parity: throw when the caller supplies a field they're not allowed to write.

**Current behavior:**
The writable-field filtering step evaluates field-level write access per field and keeps only the allowed ones (`if (canWrite) filtered[field] = value`), silently discarding denied fields before the DB write. The operation then "succeeds" while doing less than asked — the caller gets no error and no result marker indicating a field was dropped. Worse, when the stripped field is also a hook trigger (e.g. a gated `status` whose transition fires an `afterOperation` workflow/notification/billing side effect), the side-effect chain is silently skipped rather than refused. Keystone threw a field-access error for the same situation.

**Desired behavior:**
For an **access-scoped (non-sudo)** create/update, if the caller's supplied input includes a declared field whose field-level write access (`create`/`update`) denies them, the operation should fail loudly with a clear error naming the field, list, and operation — before the DB write and before `beforeOperation` side effects — rather than silently stripping the field and proceeding.
[...]

Issue #564, opened 2026-06-10
Access control fails OPEN for data keys not declared in the list config — unknown keys pass through create/update to Prisma unchecked


borisno2:
## Summary

In `@opensaas/stack-core@0.24.0`, the write pipeline silently forwards `data` keys that are **not declared as fields in the list config** straight to the Prisma client, with no access evaluation:

- `dist/access/field-access.js` `checkFieldAccess`: `if (!fieldAccess) return true` — correct for a *declared* field with no access config, but the caller resolves `fieldAccess` from the list's field map, so an **undeclared** key also arrives here as `undefined` → allow.
- `dist/context/nested-operations.js`: an undeclared key is not recognised as a relationship field → no nested-operation access processing → passthrough.

Because the generated Prisma model can have **more** fields than the stack list config declares (back-relations always exist on the Prisma side; hosts may also deliberately model a subset), a caller can reach Prisma-level writes the config never exposed:

```ts
// Student list config declares no `from_Enrolment_student` field, and no field gate runs:
await context.db.student.update({
  where: { id },
  data: { from_Enrolment_student: { disconnect: [...] /* or delete/update */ } },
})
```

…which mutates Enrolment rows through an access-scoped (non-sudo) context, bypassing field-level access entirely. Found while auditing a Keystone migration: Keystone **fails closed** here — its GraphQL input types only contain declared fields, so an unknown key is a hard error. A team porting Keystone field gates can drop a gated relationship field from the stack config and get *less* enforcement instead of a loud failure.

## Ask

Fail closed: reject (throw) any `data` key on create/update that is not a declared field of the list — mirroring the GraphQL-schema behaviour the config-driven model implies. If passthrough is ever intentional, make it an explicit opt-in (`allowUnknownDataKeys`) and document it as access-bypassing.

## Severity

We'd rate this HIGH for migrating apps: the row-level `where` scoping still applies to the *outer* record, but nested relation operations on undeclared back-relations execute with no field/op gate.

## Context

Keystone→stack migration of a production app (35 lists). The concrete in-repo trigger was a `Student.enrolments` Keystone field gate (non-admins: nested-create only) whose stack port omitted the field — the gate function existed and was unit-tested, but was wired to nothing, and the pipeline let nested writes through silently.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
[...]

PR #583, opened 2026-06-27
fix(core): fail loud on non-sudo writes in filterWritableFields (#564, #568)

borisno2:
## Summary

Brings the write pipeline's field filter (`filterWritableFields`, Phase 5 of `write-pipeline.ts`) to **Keystone fail-loud parity**. Both fixes touch the same function, so they ship together to avoid collision.

Implements #564
Implements #568
Part of #581

### #564 — undeclared data keys fail CLOSED (throw)
Previously a `data` key with no entry in `fieldConfigs` resolved `fieldConfig === undefined`, `checkFieldAccess(undefined, …)` returned `true`, and the key passed through to Prisma unchecked. Because the generated Prisma model has more fields than the list config declares (back-relations like `from_Enrolment_student`), a non-sudo caller could drive ungated nested writes on undeclared back-relations.

Now: an undeclared key on create/update **throws** for non-sudo writes. Since the filter runs at Phase 5 (before `processNestedOperations` at Phase 5.5), this also closes the undeclared-back-relation nested-write hole — no change to `nested-operations.ts` required.

### #568 — field-access-denied keys throw instead of silent strip
Previously `if (canWrite) filtered[fieldName] = value` silently dropped denied fields, letting a write "succeed" while doing less than asked (and skipping hook side effects gated on that field).

Now: a caller-supplied declared field denied by field-level access **throws**.

### Preserved behaviour
- **`sudo` is the single trusted bypass** — undeclared keys pass through under sudo, and `checkFieldAccess` already returns `true` under sudo (no parallel sudo path added).
- System fields (`id`/`createdAt`/`updatedAt`), virtual fields, relationship foreign keys (`${field}Id`), and multi-column split columns (from `getColumnNames`) still pass through cleanly.
- Declared relationship fields still pass through to nested operations.

## How the throw is surfaced
Uses the project's existing `ValidationError` (from `hooks/index.js`) — the same type the write pipeline already throws for validation failures, so callers/tests have a single convention for non-silent write rejections. The field-access ⇄ hooks import is a cycle, but `ValidationError` is referenced only inside the function body (call-time), so it is safe under ESM live bindings.

## Tests
- New focused unit tests in `packages/core/src/access/field-access.test.ts`: undeclared key throws (non-sudo create + update); undeclared key passes under sudo; system/FK fields skip cleanly; multi-column split columns pass through; declared field denied throws; sudo would-be-denied field does NOT throw; declared relationship field passes through.
- Updated 3 existing tests that encoded the OLD silent-strip contract to the new fail-loud contract: 2 in `tests/access.test.ts` and 1 nested-create test in `tests/nested-access-and-hooks.test.ts`.

## Verification
- `pnpm lint`: pass (0 errors; only pre-existing warnings)
- `pnpm manypkg fix` / `pnpm format`: clean
- `pnpm -C packages/core build` (tsc typecheck): pass
- `pnpm -C packages/core test`: 596 passed
- Changeset added (`patch` for `@opensaas/stack-core`)

---

### DIFF
--- a/.changeset/loud-foxes-write.md
+++ b/.changeset/loud-foxes-write.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Make non-sudo writes fail loud in `filterWritableFields` (Keystone parity).
+
+Undeclared `data` keys on create/update now throw instead of passing through unchecked (#564), and fields denied by field-level access now throw instead of being silently stripped (#568). `sudo` remains the single trusted bypass; system fields, relationship foreign keys, and multi-column split columns still pass through.
--- a/.changeset/loud-foxes-write.md
+++ b/.changeset/loud-foxes-write.md
@@ -4,4 +4,6 @@
 
 Make non-sudo writes fail loud in `filterWritableFields` (Keystone parity).
 
-Undeclared `data` keys on create/update now throw instead of passing through unchecked (#564), and fields denied by field-level access now throw instead of being silently stripped (#568). `sudo` remains the single trusted bypass; system fields, relationship foreign keys, and multi-column split columns still pass through.
+Undeclared `data` keys on create/update now throw instead of passing through unchecked (#564), and fields denied by field-level access now throw instead of being silently stripped (#568). `sudo` remains the single trusted bypass; system fields and relationship foreign keys still pass through. Raw multi-column split columns (e.g. `media_url`/`media_size` from an `image()`/`file()` field) are now gated by their owning field's write access — supplying them directly under non-sudo when that field denies the write throws, instead of bypassing the field's `access.create`/`access.update`.
+
+Behavioural narrowing: a list-level `resolveInput` hook that adds keys to `resolvedData` which are not declared fields will now be rejected by the undeclared-key throw. No production hook does this today.
--- a/packages/core/src/access/field-access.test.ts
+++ b/packages/core/src/access/field-access.test.ts
@@ -254,9 +254,10 @@ describe('filterWritableFields', () => {
     expect(filtered).toHaveProperty('title', 'Test')
   })
 
-  it('passes through raw per-part columns from a multi-column field (non-sudo)', async () => {
+  it('passes through raw per-part columns from a multi-column field whose write access ALLOWS (non-sudo)', async () => {
     // Multi-column fields inject raw columns (e.g. m_url/m_size) that are not
     // declared in fieldConfigs; they must not trip the undeclared-key reject.
+    // With no field-level access (allow), they pass through.
     const fieldConfigs = {
       media: {
         type: 'image',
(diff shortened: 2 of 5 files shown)


borisno2 (commented):
## Review: fail-loud writes in `filterWritableFields` (#564, #568)

**Verdict: REQUEST CHANGES** (posted as a comment because the API blocks requesting changes on one's own PR).

Reviewed strictly from the PR diff + surrounding code. The core change is well-reasoned and the bulk of it is correct: undeclared keys throw on non-sudo create/update, denied declared fields throw instead of being silently stripped, and `sudo` remains the single trusted bypass with no parallel sudo path. I verified the claims and they hold — with **one important exception** in the split-column allow-set that warrants changes before merge.

### Verified correct
- **Sudo bypass is single-path and complete.** Undeclared keys: explicit `if (isSudo) { passthrough }`. Denied declared fields: `checkFieldAccess` returns `true` under sudo (field-access.ts L35-37), so the denial throw is never reached under sudo. No sudo path hits a throw. Nested writes inherit `_isSudo` via the `context` threaded from write-pipeline (`{ ...context, prisma }`).
- **Legitimate keys still pass:** system fields (id/createdAt/updatedAt), virtual fields (declared, `virtual:true`), and `${field}Id` FKs are skipped *before* the undeclared throw; declared relationship fields fall through to nested ops. Ordering confirmed.
- **Import cycle is safe.** `field-access.ts` references `ValidationError` only inside the function body (call-time); `hooks/index.ts` declares `ValidationError` as a top-level class and only uses `checkFieldAccess` inside function bodies. Under ESM live bindings both modules are fully evaluated before either function is called. No top-level cross-reference. Safe.
- **Changed tests reflect the new contract, not a masked regression.** The 3 edited tests correctly flip silent-strip → throw, and the nested test additionally asserts `mockPrisma.post.update` was NOT called (write aborts). New tests cover non-sudo throw + sudo passthrough for both #564 and #568, plus FK/system/multi-column passthrough.
- **Error messages** name the field but carry no record-existence info — fine for input-shape / own-field-access errors.

### Findings

**1. [High — security] Raw split-column keys bypass the logical field's write-access gate (`splitColumnFields` allow-set is over-broad relative to #564's goal).**
`field-access.ts` (new `splitColumnFields` passthrough): any data key matching a declared multi-column field's `getColumnNames(fieldName)` (e.g. `media_url`, `media_size`) is passed straight through with **no `checkFieldAccess` call**.

The field-level write gate for multi-column fields lives in `hooks/index.ts` `executeFieldResolveInputHooks` (L262-289) and only fires when the **logical** key (`media`) is present in the payload — that loop iterates *declared fields*, not data keys. A non-sudo caller who supplies the **raw physical columns directly** — `context.db.x.update({ where, data: { media_url: 'https://evil', media_size: 1 } })` — never triggers that gate, and the new allow-set then waves the raw columns through unchecked, straight to Prisma. So a field whose `access.update`/`access.create` denies writes can still be written via its raw columns.

This is a pre-existing hole (raw columns previously passed via the old `checkFieldAccess(undefined) === true` default-allow), so it is not a *regression introduced here* — but #564's whole premise is closing ungated undeclared-key writes, and this PR now *explicitly codifies* the raw-column path as allowed. Suggested fix: gate split columns by the owning field's write access (map each column name back to its owning field, run `checkFieldAccess` before passthrough, throw on deny — mirroring #568) rather than blanket-allowing them. At minimum add a test asserting raw split columns for an access-denied field throw for non-sudo.

**2. [Low — contract/behaviour note] The undeclared-key throw will reject list-level `resolveInput` hooks that add non-field keys to `resolvedData`.**
`resolvedData` from list-level `resolveInput` flows into `filterWritableFields`. After this change, any key a hook adds that is not a declared field (nor system/virtual/FK/split-column) throws on non-sudo writes. I checked the repo: all *production* `resolveInput` hooks only add declared fields, so nothing breaks today (the `plugin-engine.test.ts` cases adding `field1`/`modified` invoke hooks in isolation, not through the pipeline). Consistent with Keystone parity and likely intended — but a real behavioural narrowing worth a line in the changeset/docs so plugin authors know transient non-field keys in `resolveInput` output are now rejected.

### Summary
The fail-loud rework is solid and correctly preserves the sudo bypass, system/FK/virtual skips, the import-cycle handling, and the test contract updates. The blocking concern is finding #1: the split-column allow-set leaves the logical field's write access ungated via its raw columns, which undercuts the security goal of #564.

---

borisno2:
Addressed the blocking security finding on the split-column passthrough.

Raw per-part columns from a multi-column field (e.g. `media_url`/`media_size`) are no longer blanket-forwarded. In `filterWritableFields` each raw column is now mapped back to its **owning declared field** (the field whose `getColumnNames()` produces that column) and gated by that field's write access via the same `checkFieldAccess(owningField.access, operation, args)` used for normal fields:

- Denied (non-sudo) -> **throws** `ValidationError`, naming the owning field (consistent with #568).
- Allowed (or sudo, since `checkFieldAccess` returns `true` under sudo) -> passes through, preserving the legitimate multi-column write path.
- A raw column mapping to no declared split-column field remains an undeclared key, so the #564 throw still applies.

This closes the bypass where a non-sudo caller supplying the raw columns directly (`data: { media_url, media_size }`) skipped the owning field's gate — the logical-key gate in `executeFieldResolveInputHooks` only fires when the `media` key is present.

Also updated the existing changeset (`.changeset/loud-foxes-write.md`) with the Finding 2 note: a list-level `resolveInput` hook that adds non-field keys to `resolvedData` will now be rejected by the undeclared-key throw (no production hook does this today).

Tests added/updated in `packages/core/src/access/field-access.test.ts`: denied owning-field raw columns now throw (create + update); allowed pass through; sudo passes through regardless; existing legitimate multi-column passthrough still passes.

Verification: `pnpm lint` (0 errors). All package test suites green run directly — core 598, ui 190, auth 133, cli 289, rag 356, storage 123, storage-s3 43, storage-vercel 30, create-opensaas-app 41. (`turbo test` surfaced unrelated example `build` failures because `prisma generate` can't fetch its schema-engine binary in this sandboxed network — not related to this change.)

---

Issue #565, opened 2026-06-10
context.db has no findFirst — non-unique single-row lookups have no direct equivalent


borisno2:
## Summary

The generated `context.db.<list>` delegates expose `findUnique` / `findMany` / `create` / `update` / `delete` / `deleteMany` / `count`, but not Prisma's `findFirst`. A non-unique single-row lookup (`findFirst({ where, orderBy })`) is one of the most common Prisma patterns, and code being migrated from raw Prisma (or written by devs/agents who know Prisma) reaches for it naturally — and gets a runtime `TypeError: context.db.<list>.findFirst is not a function` that type-mocked tests don't catch.

Found in a Keystone→stack migration where a list hook called `context.db.account.findFirst(...)`: it type-checked against a permissive seam, passed mocked tests, and failed only against the real generated client.

## Ask

Either:
1. Add `findFirst(args)` to the access-controlled delegates (sugar over the existing access-filtered `findMany` with `take: 1`, so no new access surface); or
2. If intentionally omitted, make it a hard TYPE-level absence in the generated delegate types so misuse is a compile error rather than a runtime TypeError (today the gap is only discoverable at runtime when code is typed against a structural seam), and document the `findMany({ take: 1 })` idiom in the context API docs.

## Context

@opensaas/stack-core 0.24.0. Related to the migration-recipe docs: agents converting Prisma/Keystone code need the supported-method list to be explicit.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Add `findFirst` to the access-controlled `context.db.<list>` delegates as sugar over the access-filtered `findMany` with `take: 1`, so the common Prisma single-row-by-non-unique-filter pattern works (and fails at compile time if misused, not runtime).

**Current behavior:**
The generated/access-controlled list delegates expose `findUnique`, `findMany`, `create`, `update`, `delete`, `count`, `createMany`, `updateMany` (and `get` for singletons), but not `findFirst`. Code migrated from raw Prisma or Keystone (or written by anyone who knows Prisma) reaches for `context.db.<list>.findFirst({ where, orderBy })` and gets a runtime `TypeError: context.db.<list>.findFirst is not a function`. Because call sites are often typed against a permissive structural seam, this passes type-checking and mocked tests and only fails against the real client.

**Desired behavior:**
Each list delegate should expose `findFirst(args)` with the same access guarantees as the other read operations:
- Returns a single record or `null`.
[...]

PR #584, opened 2026-06-27
feat(core): add findFirst to access-controlled context.db delegates

borisno2:
Implements #565. Part of #581.

## What

Adds `findFirst(args)` to the access-controlled `context.db.<list>` delegates. Per the issue, this is **option 1: sugar over the existing access-filtered `findMany` with `take: 1`** — it introduces **no new access surface**. It applies exactly the same query-access checks and access-controlled include building as `findMany`, then returns the first row or `null`.

This honours the shared read-side policy from #581: reads keep the silent-failure contract (access-denied query → `null`, not a throw), and `sudo` remains the single bypass (inherited unchanged from `findMany`).

## How

- **`packages/core/src/context/index.ts`** — new `createFindFirst(findManyOp)` factory that delegates to the already-built `findManyOp` with `take: 1` and returns `result[0] ?? null`. Wired into the delegates `operations` object. Zero duplication of access logic — identical semantics by construction. Supports the same `where`/`orderBy`/`include`/`query`-fragment args as `findMany`.
- **`packages/core/src/access/types.ts`** — added `AugmentedFindFirst` interface (mirrors `AugmentedFindUnique`/`AugmentedFindMany`): a `query`-fragment overload returning `Promise<ResultOf<fragment> | null>` plus the original Prisma `findFirst` signature passthrough. Wired `findFirst` into the `AccessControlledDB` mapped type. The generic `PrismaModelDelegate` already declared `findFirst`.
- **`packages/cli/src/generator/types.ts`** — the generated `CustomDB` explicitly enumerates delegate methods (it is **not** purely derived from `AccessControlledDB`), so `findFirst` and a `<List>FindFirstArgs` type are now emitted for each list in `.opensaas/types.ts`, giving migrated apps full type support for the familiar Prisma `findFirst` pattern.

## Tests

Added to `packages/core/tests/context.test.ts` (the existing context/findMany suite):
- returns the first matching row (and delegates with `take: 1`)
- returns `null` (not `undefined`, not throw) when nothing matches
- respects `where` / `orderBy`
- honours query-access denial the same as `findMany` (denied → `null`, prisma not hit)
- respects a `query` fragment, narrowing the returned single result

## Verification

- `pnpm lint` — 0 errors (only pre-existing unrelated warnings)
- `pnpm manypkg fix` + `pnpm format` — clean
- `pnpm -C packages/core build` and `pnpm -C packages/cli build` — pass
- `pnpm -C packages/core test` — 593 passed (incl. 5 new)
- Regenerated the blog example to confirm `findFirst`/`FindFirstArgs` appear in generated types
- Changeset added (minor on `@opensaas/stack-core` and `@opensaas/stack-cli`)

---

### DIFF
--- a/.changeset/swift-otters-findfirst.md
+++ b/.changeset/swift-otters-findfirst.md
@@ -1,0 +1,31 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+Add `findFirst` to access-controlled `context.db.<list>` delegates
+
+`findFirst` is sugar over the existing access-filtered `findMany` (`take: 1`), so
+it introduces no new access surface: it applies the exact same query-access checks
+and access-controlled include building as `findMany`, then returns the first
+matching row or `null`. It honours the read-side silent-failure contract — an
+access-denied query yields `null` rather than throwing.
+
+```ts
+// Non-unique single-row lookup
+const account = await context.db.account.findFirst({
+  where: { userId: '123' },
+  orderBy: { createdAt: 'desc' },
+})
+
+// Narrow the single result with a query fragment
+const post = await context.db.post.findFirst({
+  where: { published: true },
+  query: postFragment,
+})
+// post: ResultOf<typeof postFragment> | null
+```
+
+The CLI type generator now emits a `findFirst` method (and `<List>FindFirstArgs`
+type) for each list in the generated `.opensaas/types.ts`, so migrated apps that
+reach for the familiar Prisma `findFirst` pattern get full type support.
--- a/packages/cli/src/commands/__snapshots__/generate.test.ts.snap
+++ b/packages/cli/src/commands/__snapshots__/generate.test.ts.snap
@@ -281,6 +281,14 @@ export type UserFindManyArgs = Omit<Prisma.UserFindManyArgs, 'select'> & {
 }
 
 /**
+ * Custom FindFirstArgs for User with virtual field support
+ */
+export type UserFindFirstArgs = Omit<Prisma.UserFindFirstArgs, 'select'> & {
+  select?: UserSelect | null
+  query?: Fragment<UserOutput, FieldSelection<UserOutput>>
+}
+
+/**
  * Custom CreateArgs for User with virtual field support
  */
 export type UserCreateArgs = Omit<Prisma.UserCreateArgs, 'select'> & {
@@ -329,6 +337,9 @@ export type CustomDB = Omit<AccessControlledDB<PrismaClient>,
   user: {
     findUnique: <T extends UserFindUniqueArgs>(
       args: Prisma.SelectSubset<T, UserFindUniqueArgs>
+    ) => Promise<UserGetPayload<T> | null>
+    findFirst: <T extends UserFindFirstArgs>(
+      args?: Prisma.SelectSubset<T, UserFindFirstArgs>
     ) => Promise<UserGetPayload<T> | null>
     findMany: <T extends UserFindManyArgs>(
       args?: Prisma.SelectSubset<T, UserFindManyArgs>
--- a/packages/cli/src/generator/__snapshots__/types.test.ts.snap
+++ b/packages/cli/src/generator/__snapshots__/types.test.ts.snap
@@ -140,6 +140,14 @@ export type UserFindManyArgs = Omit<Prisma.UserFindManyArgs, 'select'> & {
 }
 
 /**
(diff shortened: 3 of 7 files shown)


borisno2 (commented):
## Review: `feat(core): add findFirst to access-controlled context.db delegates`

**Verdict: APPROVE** ✅ (posted as a comment — the GitHub token here owns this PR and cannot submit a formal Approve.)

Reviewed strictly from the PR diff and the code it touches. This is a tight, low-risk change that does exactly what it claims: `findFirst` is sugar over the access-controlled `findMany` (`take: 1`, return `result[0] ?? null`) with zero duplicated access logic.

### Risk areas verified

- **No new access surface.** `createFindFirst(findManyOp)` delegates entirely to the already-built `findManyOp`. All query-access checks, filter merging, access-controlled `include` building, fragment narrowing, and the `sudo` bypass live in `findMany` and are inherited by construction — there is no parallel access path that can drift (`packages/core/src/context/index.ts:614-628`).
- **Silent-failure contract preserved.** Query-access denial in `findMany` returns `[]`, and `[][0] ?? null` yields `null` — not a throw, not `[]`, not `undefined`. Covered by the "denied -> null, prisma not hit" test.
- **`null` on no match.** `result[0] ?? null` correctly returns `null` (test explicitly asserts `not.toBeUndefined()`).
- **Arg forwarding.** `{ ...args, take: 1 }` forwards `where`/`orderBy`/`include`/`query`/`skip`; `take` is intentionally pinned to 1 (correct for findFirst).
- **Type surface.** `AugmentedFindFirst` mirrors `AugmentedFindUnique`/`AugmentedFindMany`: overload 1 is the `query`-fragment form returning `Promise<ResultOf<fragment> | null>`, overload 2 is the original Prisma passthrough. `FindFirstQueryArgs` (where/orderBy/skip) matches the `FindManyQueryArgs` convention. The only `any` is the existing localized `eslint-disable` pattern on the conditional-type guard — no new `any`/`unknown` leaks into the public surface.
- **Generated CLI type.** `${listName}FindFirstArgs` and the `findFirst` method are consistent with the generated `findMany` (`args?` optional, `Promise<${listName}GetPayload<T> | null>`, custom select/include/query). Relationship vs non-relationship branches match the `FindManyArgs` generator.
- **Changeset.** Correctly bumps both `@opensaas/stack-core` (minor) and `@opensaas/stack-cli` (minor).
- **Tests.** All five required cases present: first-row + `take: 1`, null-on-no-match, where+orderBy, access-denial→null, query-fragment narrowing.

### Notable positive: this closes a latent bug

`packages/core/src/query/index.ts:500` (`runQueryOne`) already calls `context.db[dbKey].findFirst(...)`, but `findFirst` was **never registered** in the `operations` object built by `getContext` (`context/index.ts:247-264`), and `db[dbKey]` is a plain object with no prisma fallback. So prior to this PR any real `runQueryOne` call would throw `TypeError: ...findFirst is not a function`. The existing `query/index.test.ts` masks this because it uses a hand-mocked delegate rather than the real context. This PR fixes that gap as a side effect — nice.

### Findings

**Low severity**

1. `packages/core/src/context/index.ts:522-523` — singleton lists throw with a message hardcoded to `findMany`. Because `findFirst` delegates to `findManyOp`, calling `findFirst` on a singleton list throws `"Cannot use findMany: <List> is a singleton list. Use get() instead."`, which names an operation the caller never invoked. Cosmetic, but mildly confusing. Optional: thread the caller's op name into the error, or accept as-is.

Nothing else surfaced.

---

borisno2:
CI fix: the failing `@opensaas/stack-cli#test` job was caused by stale generator snapshots in the CLI package (`packages/cli/src/generator/__snapshots__/types.test.ts.snap` and `packages/cli/src/commands/__snapshots__/generate.test.ts.snap`), which still encoded the old generated output without the new `findFirst` method and `<List>FindFirstArgs` type.

I regenerated those snapshots to expect the additive `findFirst` output. The diff is purely additive (228 insertions, 0 deletions) — it adds the per-list `<List>FindFirstArgs` type and the `findFirst` delegate signature, mirroring the existing `findMany`/`findUnique` blocks. No generator logic, core logic, or runtime behaviour was touched.

Verification:
- `pnpm -C packages/cli test` → 289 passed (17 files), previously-failing package now green
- `pnpm -C packages/core test` → 593 passed
- ui/auth/rag package suites all pass (190 / 133 / 356)
- `pnpm lint` → 0 errors

No changeset added (this PR already has one).

---

Issue #567, opened 2026-06-10
findUnique executes findFirst and silently accepts a non-unique where (nondeterministic row)


borisno2:
## Summary

`context.db.<list>.findUnique` in `@opensaas/stack-core@0.24.0` executes Prisma `findFirst` under the hood and accepts ANY `where` shape — a non-unique filter is not rejected at type level or runtime, and with no `orderBy` the returned row is nondeterministic. It also honours only `where`/`include` (other args silently ignored).

Keystone's `findOne` enforced unique-where via the GraphQL schema, so migrating code can silently change semantics: a converted `findOne({ where: { someNonUniqueField } })` type-checks, runs, and returns an arbitrary matching row.

## Ask

Either constrain `where` to the model's unique inputs (Prisma's own `WhereUniqueInput` typing + a runtime guard), or execute real `findUnique` semantics, or — if findFirst behaviour is intentional — name/document it as such and add real `findFirst` to the API (see #565) so intent is explicit.

## Context

@opensaas/stack-core 0.24.0, dist/context/index.js (createFindUnique). Found auditing a Keystone→stack migration where converted call sites rely on de-facto-unique columns (e.g. `where: { subjectId }`) that nothing now enforces.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** `context.db.<list>.findUnique` accepts any `where` shape and resolves it as a `findFirst`, so a non-unique filter silently returns a nondeterministic matching row. Enforce unique-`where` semantics (Keystone `findOne` parity).

**Current behavior:**
`findUnique` merges the list's query-access filter into the caller's `where` and then runs Prisma `findFirst` under the hood (it must use `findFirst` so the access filter can be AND-ed in — real `findUnique` only accepts unique fields). But it accepts **any** `where`, with no unique constraint at the type level or at runtime, and supplies no `orderBy`. A caller migrating Keystone `findOne({ where: { someNonUniqueField } })` (or relying on a de-facto-unique column) type-checks, runs, and gets an arbitrary matching row when more than one matches. It also honours only `where`/`include`/`query`, silently ignoring other args.

**Desired behavior:**
`findUnique` should require a unique identifier and fail loudly otherwise, while preserving access scoping:
- The caller's `where` must identify a row by a unique input (the `id`, a single `@unique` field, or a declared compound-unique). At the type level, the generated delegate should constrain `findUnique`'s `where` to the model's unique inputs.
[...]

PR #586, opened 2026-06-27
fix(core): enforce unique-where for findUnique

borisno2:
Implements #567. Part of #581.

## Problem

`context.db.<list>.findUnique` executed Prisma `findFirst` under the hood and accepted **any** `where` shape. A non-unique filter was neither rejected nor deterministic (no `orderBy`), silently diverging from Keystone's `findOne` unique-where semantics. A migrated `findOne({ where: { someNonUniqueField } })` would type-check, run, and return an arbitrary matching row.

## Change

Added a runtime guard to `createFindUnique` (`packages/core/src/context/index.ts`) that enforces a unique `where`:

- The `where` must contain **exactly one** recognised unique key and no other keys.
- The unique-key set is derived from the list config **at runtime**: `id` (always) plus any field declared `isIndexed: 'unique'`. For a unique `relationship` field, the FK column `<field>Id` is the recognised key (that's where Prisma puts `@unique`).
- A non-unique `where` is a **caller-shape error**, so the guard **throws** a clear `ValidationError` (fail-loud-on-misuse, per the PRD) rather than masking it as a silent `null`. The message points callers to `findFirst` (#565) for non-unique single-row lookups.

**Documented rule (in code):** the config API exposes no list-level compound (`@@unique`) declaration at runtime, so compound selectors can't be validated. The guard therefore enforces the tractable subset — exactly one single-field unique key — which fixes the bug and rejects extra non-unique keys, without ever falsely rejecting a valid single-field unique lookup.

## Type level

No generator change needed. The generated `<List>FindUniqueArgs` already `Omit`s only `select`/`include` from `Prisma.<List>FindUniqueArgs`, so its `where` stays Prisma's `<List>WhereUniqueInput` — the type is already correctly constrained. The runtime guard backstops untyped callers. A code comment records this.

## Preserved behaviour

Query-access check, filter merge, access-controlled `include` / `query`-fragment narrowing, field visibility, `sudo` bypass, and `null` on access denial / no match are all unchanged.

## Fold-in (#565 review note)

The singleton "Cannot use findMany … Use get() instead" message reachable via `findFirst` was **skipped**: `createFindFirst` only receives `findManyOp` (no `listConfig`/`listName`), and the existing singleton test asserts the literal `'Cannot use findMany'`. Making it findFirst-accurate would require threading config + editing an unrelated test, exceeding the "clean 1-line" bar — left out to avoid scope creep.

## Tests

Added to the existing context/findUnique tests in `packages/core/tests/context.test.ts`:
- valid unique `where` (id) works as before — access + include intact
- configured-unique field (email) accepted
- query-fragment narrowing with a unique `where`
- non-unique `where` THROWS (and message references `findFirst`); guard runs before DB access
- unique key mixed with extra non-unique keys THROWS
- empty `where` THROWS
- access-denied findUnique still returns `null`
- no-match returns `null`

## Verification

- `pnpm lint` — 0 errors (only pre-existing warnings)
- `pnpm manypkg fix`, `pnpm format` — clean
- `pnpm build` — full repo typecheck/build passes
- Full test suite green: core (601), cli (289), auth (133), ui (190), rag (178), storage (123), storage-s3 (43), storage-vercel (30), create-opensaas-app (41)
- Changeset added (`@opensaas/stack-core`: patch)

---

### DIFF
--- a/.changeset/swift-foxes-cheer.md
+++ b/.changeset/swift-foxes-cheer.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Enforce unique-`where` for `context.db.<list>.findUnique` — a non-unique `where` now throws a clear error instead of silently returning a nondeterministic row. Use `findFirst` for non-unique single-row lookups.
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -66,6 +66,72 @@ function isSingletonList(listConfig: ListConfig<any>): boolean {
 }
 
 /**
+ * Compute the set of single-field unique selectors a `findUnique` `where` may be
+ * keyed by, derived from what the list config exposes at runtime.
+ *
+ * The set is:
+ * - `id` — always a unique identifier on every list.
+ * - Any field declared `isIndexed: 'unique'` in the config (e.g. `text({ isIndexed: 'unique' })`).
+ * - For a `relationship` field declared `isIndexed: 'unique'`, the foreign-key
+ *   column name (`<field>Id`) — that is the column Prisma marks `@unique`, so the
+ *   unique `where` is keyed by `<field>Id`, not the relation field itself.
+ *
+ * Chosen rule (documented intentionally): the config does NOT expose compound
+ * (`@@unique`) keys at runtime — there is no list-level unique declaration in the
+ * config API — so we cannot validate compound `<Model>_<a>_<b>` selectors. We
+ * therefore enforce the tractable subset: `where` must contain EXACTLY ONE
+ * recognised single-field unique key and NO other keys. This rejects non-unique
+ * filters (the bug in #567) and rejects extra non-unique keys alongside a unique
+ * one, while never falsely rejecting a valid single-field unique lookup. If a
+ * project legitimately needs a compound-unique lookup, that path is not covered
+ * here and would need explicit config support; the safe escape hatch for any
+ * non-unique single-row lookup is `findFirst` (see #565).
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
+function getUniqueWhereKeys(listConfig: ListConfig<any>): Set<string> {
+  const keys = new Set<string>(['id'])
+
+  for (const [fieldKey, fieldConfig] of Object.entries(listConfig.fields)) {
+    if (!fieldConfig || typeof fieldConfig !== 'object') continue
+    if (!('isIndexed' in fieldConfig) || fieldConfig.isIndexed !== 'unique') continue
+
+    if (fieldConfig.type === 'relationship') {
+      // A unique relationship's `@unique` lives on the FK column `<field>Id`.
+      keys.add(`${fieldKey}Id`)
+    } else {
+      keys.add(fieldKey)
+    }
+  }
+
+  return keys
+}
+
+/**
+ * Enforce Keystone `findOne` semantics for `findUnique`: the caller-supplied
+ * `where` must be a valid unique selector. A non-unique `where` is a caller-shape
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review: enforce unique-`where` for `findUnique` (#567 / #581)

**Verdict: APPROVE (posted as COMMENT — GitHub blocks self-approval).** The implementation is correct, well-scoped, and the primary over-strictness / ordering risks are handled correctly. No blocking issues found. A few non-blocking notes below.

### What it does
Adds a runtime guard (`getUniqueWhereKeys` + `assertUniqueWhere`) to `createFindUnique` in `packages/core/src/context/index.ts`. The `where` must contain exactly one recognised single-field unique key (`id`, any `isIndexed: 'unique'` field, or `<field>Id` for a unique relationship) and no other keys; otherwise it throws `ValidationError` pointing the caller to `findFirst`. Type-level behaviour is unchanged (generated `FindUniqueArgs.where` is already `WhereUniqueInput`).

### Verified correct (the high-risk areas)
- **Ordering vs access-filter merge is correct.** `assertUniqueWhere(args.where, …)` runs on the caller's raw `where` *before* the query-access check and `mergeFilters` merge. It can never misclassify an AND-merged access filter as "non-unique", because it never sees the merged shape. This was the biggest risk and it's handled right.
- **Access-denial → `null` preserved**, no-match → `null` preserved, sudo bypass / include / query-fragment narrowing all unchanged (guard is independent of those paths).
- **No internal-caller regression.** Every framework caller of the *wrapped* `context.db.<list>.findUnique` passes `{ where: { id } }` (auth `getUser`/`getCurrentUser` in `packages/auth/src/config/plugin.ts:158,171`; `packages/ui/src/components/ItemForm.tsx:55`; singleton test). All pass the guard. `update`/`delete`/`count`/`get` and the nested-ops / write-pipeline / rag-search lookups call the **raw Prisma delegate** (`model.findUnique`), not the wrapper, so they're unaffected. MCP `query` uses `findMany`, not `findUnique`.
- **Unique-key derivation is correct.** FK column name is `${fieldName}Id` (confirmed `packages/core/src/fields/index.ts:1169`), and `@unique` sits on that FK column — so adding `<field>Id` for unique relationships is right. A valid single-field unique lookup (e.g. `email`) is accepted (test covers it).
- **Throw-on-misuse** is the intended fail-loud contract per PRD #581, and the message references `findFirst`. Sudo also being subject to the shape guard is defensible (a sudo caller passing a non-unique `where` is still misusing `findUnique`).
- **Tests** cover the full matrix: valid-unique (id + configured email), query-fragment narrowing, non-unique throw (with `findFirst` message + DB-not-called), extra-key throw, empty-where throw, access-denied→null, no-match→null. No generator/snapshot change needed (claim holds — type was already constrained). Changeset present (`@opensaas/stack-core` patch).

### Non-blocking notes
1. **Modern-Prisma `WhereUniqueInput` permits a unique key + extra non-unique filters** (e.g. `{ id, tenantId }`). The guard rejects the extra key (throws). This is intentional and documented, and is *stricter* than Prisma — safe (never falsely rejects a bare single-field unique lookup), but if any downstream code relies on the unique-plus-filter form it would now throw. No such caller exists in this repo today; worth a one-line mention in user-facing docs.
2. **Compound `@@unique` selectors are unsupported** (`{ a_b: {…} }` → single key `a_b` not in the set → throws). The PR documents this explicitly and routes such cases to `findFirst`; acceptable given the config API exposes no list-level compound declaration at runtime.
3. **Edge case — unique one-to-one where this side does not own the FK** (`relationship({ isIndexed: 'unique', db: { foreignKey: false } })`): the derived key `<field>Id` wouldn't exist on this model, so a lookup keyed by the actual unique column could be rejected. Very niche, not exercised by current lists; flagging for awareness, not as a defect.

Nice, tightly-scoped fix with the ordering done correctly.

---

Issue #578, opened 2026-06-26
processNestedConnect falsely DENIES nested-relation access filters — in-memory matcher only handles flat scalar filters


borisno2:
## Summary

In `@opensaas/stack-core@0.24.0`, `processNestedConnect` (`dist/context/nested-operations.js`) evaluates whether the caller may `connect` an existing row by running the connected list's access filter through an **in-memory matcher** that only understands FLAT scalar filters. When the access filter is a **nested relation filter** — e.g. `{ user: { id: { equals: session.userId } } }` — the matcher hits its fallback `else if (item[key] !== value)` branch, compares an **object** (`{ id: { equals } }`) against the row's field with strict `!==`, and **always fails** → `Error: Access denied: Cannot connect to this item`, even when the caller is legitimately allowed.

This breaks **every access-scoped (non-sudo) nested `connect` to a list whose `access.operation.query` (or the field/relationship access) returns a nested relation filter.**

## Concrete repro (from a Keystone→stack migration)

An account-holder (ACCOUNT role) creating a Student connected to their OWN account:
```ts
context.db.student.create({ data: { ..., account: { connect: { id: ownAccountId } } } })
```
`Account`'s row filter is `{ user: { id: { equals: session.userId } } }` (the account-holder owns the account via account→user). `processNestedConnect`'s matcher can't evaluate the nested `user.id.equals`, so it denies — 500ing the dashboard student-create / registration flow for a row the caller demonstrably owns. Under Keystone the same connect worked (Keystone evaluates the filter against the DB).

## Root cause

The matcher does a JS-object walk of the access filter and only compares scalar leaves; nested-relation predicates (`{ relation: { field: { equals } } }`), and likely `AND`/`OR`/`some`/`none`/`not`, fall through to a strict object `!==` and never match.

## Proposed fix

Evaluate the access filter in the DATABASE, not in memory: `model.findFirst({ where: { AND: [connection, accessFilter] } })` — if it returns the row, the connect is allowed. This handles arbitrary nested/relation/boolean filters correctly and mirrors Keystone's nested-connect enforcement.

## Severity

HIGH — silently breaks legitimate user flows (any account-holder/self-scoped nested connect) on a real DB. Not caught by mocked unit tests; only surfaces against a database.

## Relation to #569

[...]

PR #587, opened 2026-06-27
Fix nested connect false denial: read access + DB reachability

borisno2:
## Summary

Implements #578. Part of #581.

`processNestedConnect` (and the connect branch of `processNestedConnectOrCreate`) had two related bugs that falsely denied legitimate access-scoped nested `connect`s:

1. **Wrong access phase.** It checked the target list's `operation.update`. Connecting references an existing row; it does not modify it. Keystone required **read/query** access on the target. The check now uses `access.operation.query`.
2. **In-memory matcher could not evaluate nested filters.** When access returned a filter object, the old code walked it in memory and only compared flat scalar leaves. A nested-relation filter like `{ user: { id: { equals: session.userId } } }` (and `AND`/`OR`/`some`/`none`/`not`) fell through to a strict object `!==` and **always** failed → false denial.

## The fix

Connect now mirrors Keystone semantics and evaluates the access filter in the **database**, not in memory:

- `accessResult === false` → deny (throw)
- `accessResult === true` → allow, no re-check (still verifies the row exists via `findUnique`, preserving the "Item not found" behaviour)
- `accessResult` is a filter object → **DB-reachability check**: `model.findFirst({ where: { AND: [connection, accessFilter] } })`. Allowed iff a row is returned. A non-existent id or an unreachable row both yield no row → denied (security preserved).

The connection identifier and the access filter are AND-combined with an explicit `{ AND: [connection, accessResult] }`, which folds existence into the reachability query and correctly handles arbitrary nested-relation predicates and boolean combinators.

The same corrected logic is applied to `connectOrCreate`'s connect branch. That branch additionally no longer swallows access-denied errors in its existence fallback — only a genuinely absent row falls through to create.

**Sudo** bypasses the whole check, exactly as before (`if (!context._isSudo)`).

### Owning-side field-level access (deferred)

The brief mentions also applying the owning relationship field's field-level access on connect. `NestedOpHandlerArgs` does not currently carry the owning field's config/access, so threading it through would be significant new plumbing. This is deferred with a `// TODO(#578)` in the code. The read-access + DB-reachability change here is the must-have that fixes the reported false-denial.

## Tests

Added to `packages/core/tests/nested-access-and-hooks.test.ts`:

- Nested connect to a target whose query access returns a **nested-relation filter** succeeds when reachable, denied when not (account-holder self-connect repro).
- Connect requires **read** access not target **update** (read-not-update repro succeeds).
- **AND/OR/some/none/not** boolean-combinator filter evaluated via DB reachability (no false denial/allow).
- `connectOrCreate`'s connect branch denies an existing-but-unreachable row and allows a reachable one.
- Sudo connect bypasses the check entirely.

## Verification

- `pnpm -C packages/core test`: 610 passed
- `pnpm -C packages/cli test`: 289 passed
- `pnpm lint`: 0 errors (pre-existing warnings only)
- `pnpm manypkg fix` + `pnpm format`: clean
- core `tsc` build: passes
- Changeset: `@opensaas/stack-core` patch

---

### DIFF
--- a/packages/core/src/context/nested-operations.ts
+++ b/packages/core/src/context/nested-operations.ts
@@ -117,8 +117,76 @@ async function processNestedCreate(
 }
 
 /**
+ * Verify that a single connection target is reachable for the caller.
+ *
+ * Connecting an existing row references it; it does not modify the row's own
+ * data. Mirroring Keystone, this requires **read/query** access on the target
+ * list (not `update`). When query access returns a filter object, the filter is
+ * evaluated in the DATABASE (not in memory) via
+ * `findFirst({ where: { AND: [connection, accessFilter] } })`. The connect is
+ * allowed iff that query returns a row, which correctly handles arbitrary
+ * nested-relation predicates and boolean combinators (`AND`/`OR`/`some`/
+ * `none`/`not`). The existence check is folded into the reachability query so a
+ * non-existent id is still denied.
+ *
+ * Sudo bypasses the entire check (handled by the caller).
+ *
+ * TODO(#578): also gate connect by the owning relationship field's field-level
+ * access. `NestedOpHandlerArgs` does not currently carry the owning field's
+ * config/access, so threading it through is deferred; the read-access +
+ * DB-reachability change here is the must-have correctness fix.
+ */
+async function verifyConnectReachable(
+  connection: Record<string, unknown>,
+  relatedListName: string,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
+  relatedListConfig: ListConfig<any>,
+  context: AccessContext,
+  prisma: unknown,
+): Promise<void> {
+  // Access Prisma model dynamically - required because model names are generated at runtime
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const model = (prisma as any)[getDbKey(relatedListName)]
+
+  // Connecting references an existing row; it requires READ (query) access on
+  // the target, not update access.
+  const queryAccess = relatedListConfig.access?.operation?.query
+  const accessResult = await checkAccess(queryAccess, {
+    session: context.session,
+    context,
+  })
+
+  // Explicit denial.
+  if (accessResult === false) {
+    throw new Error('Access denied: Cannot connect to this item')
+  }
+
+  // Full access: still verify the row exists (keep "Item not found" behaviour).
+  if (accessResult === true) {
+    const item = await model.findUnique({ where: connection })
+    if (!item) {
+      throw new Error(`Cannot connect: Item not found`)
+    }
+    return
+  }
+
+  // Filter result: confirm the row is reachable under the access filter by
+  // AND-combining the connection identifier with the filter and querying the DB.
+  // A non-existent id and an unreachable row both yield no row → denied. This
+  // correctly evaluates arbitrary nested-relation predicates and boolean
+  // combinators because the database does the matching, not an in-memory walk.
+  const reachable = await model.findFirst({
+    where: { AND: [connection, accessResult] },
+  })
+
+  if (!reachable) {
+    throw new Error('Access denied: Cannot connect to this item')
+  }
+}
+
+/**
  * Process nested connect operations
- * Verifies update access to the items being connected
+ * Verifies read (query) access to the items being connected via DB reachability
  */
 async function processNestedConnect(
   connections: Record<string, unknown> | Array<Record<string, unknown>>,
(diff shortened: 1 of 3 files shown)


packages/core/src/context/nested-operations.ts
```diff
@@ -116,9 +116,77 @@ async function processNestedCreate(
...
+  // Explicit denial.
+  if (accessResult === false) {
+    throw new Error('Access denied: Cannot connect to this item')
+  }
+
+  // Full access: still verify the row exists (keep "Item not found" behaviour).
+  if (accessResult === true) {
+    const item = await model.findUnique({ where: connection })
+    if (!item) {
+      throw new Error(`Cannot connect: Item not found`)
```
borisno2, line 168: **Behavioral change worth documenting (not a blocker).** Switching the gate from `operation.update` to `operation.query` interacts with `checkAccess`'s deny-by-default: when `queryAccess` is `undefined`, `checkAccess` returns `false` (engine.ts L73-76: `if (!accessControl) return false`), so this throws.

Concretely, a related list with `query: undefined` but a permissive `update` (e.g. `update: () => true`) previously allowed nested `connect` and now denies it. This is the correct direction (connect = read), and it's consistent with reads already being deny-by-default — a list you can't read shouldn't be connectable. But it is a stricter result than Keystone's allow-all operation default, and it can bite simple lists that never declared `query`.

Suggest: call this out explicitly in the changeset (the current changeset describes the filter-evaluation fix but not the update→query phase shift's deny-by-default implication), and confirm with maintainers that "no `query` rule ⇒ connect denied" is the intended contract.

---

packages/core/src/context/nested-operations.ts
```diff
@@ -116,9 +116,77 @@ async function processNestedCreate(
...
+  // A non-existent id and an unreachable row both yield no row → denied. This
+  // correctly evaluates arbitrary nested-relation predicates and boolean
+  // combinators because the database does the matching, not an in-memory walk.
+  const reachable = await model.findFirst({
+    where: { AND: [connection, accessResult] },
+  })
+
+  if (!reachable) {
+    throw new Error('Access denied: Cannot connect to this item')
+  }
```
borisno2, line 184: Optional hardening (low): the `else` branch trusts the static type `boolean | PrismaFilter` and passes `accessResult` straight into `findFirst({ where: { AND: [connection, accessResult] } })`. A misbehaving access fn that returns e.g. `null`/a string at runtime would land here; Prisma would reject it (fails closed — error, not a silent allow), so there's no security hole, but using the existing `isPrismaFilter()` helper from `engine.ts` would make the contract explicit and the error message clearer.

---

packages/core/src/context/nested-operations.ts
```diff
@@ -316,31 +347,47 @@ async function processNestedConnectOrCreate(
...
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-          const model = (prisma as any)[getDbKey(relatedListName)]
-          const existingItem = await model.findUnique({
-            where: opRecord.where,
+        // Access Prisma model dynamically - required because model names are generated at runtime
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        const model = (prisma as any)[getDbKey(relatedListName)]
+        const where = opRecord.where as Record<string, unknown>
+
+        const existingItem = await model.findUnique({ where })
```
borisno2, line 364: Verified the security-critical claim for the connectOrCreate connect branch: removing the old `catch {}` is correct and does **not** introduce a premature-create bug.

- `processNestedCreate` (called just above this block) only runs hooks/validation and returns shaped data — it performs **no** DB write. The single actual write for the whole operation is `strategy.persist` in `write-pipeline.ts` (Phase 8), which runs *after* all nested processing completes.
- So an access-denied existing row now `throw`s here and aborts the entire operation before any Prisma write — no orphaned/silently-created row.
- The fallback to CREATE is correctly reached only when `existingItem` is falsy (genuine row-absent), not when access is denied. The test `connectOrCreate's connect branch ... denies unreachable rows` asserts `student.create` was NOT called. 

The old `catch {}` would have swallowed the `accessResult === false` throw and silently downgraded a denial into a create — this PR fixes that real bug.

---

borisno2 (commented):
## Review: Fix nested connect false denial (read access + DB reachability)

**Verdict: APPROVE (posted as COMMENT — GitHub blocks self-approval on an author's own PR).** The core fix is correct, the security direction is sound, and test coverage is strong. One behavioral-change note and one optional hardening suggestion below; neither is a blocker.

### What this PR does
Fixes false denial of nested `connect` (and `connectOrCreate`'s connect branch) by (1) checking the target list's `operation.query` (read) access instead of `operation.update` — connect references a row, it doesn't modify it — and (2) replacing a broken in-memory filter matcher with a DB-reachability check `findFirst({ where: { AND: [connection, accessFilter] } })` that correctly evaluates nested-relation and `AND`/`OR`/`some`/`none`/`not` filters.

### Security assessment (the critical part) — PASS
- **Direction is correct.** A row not reachable under the query filter yields no row from `findFirst` → denied (no false ALLOW). A reachable row matching the id → allowed (no false DENY). `AND: [connection, accessFilter]` folds existence into reachability, so a non-existent id is also denied.
- **Three outcomes handled:** `false` → throw; `true` → allow but still `findUnique` to preserve "Item not found"; filter object → DB reachability. **Sudo** bypasses the whole check (`if (!context._isSudo)`), and the sudo test asserts neither the access fn nor `findFirst` runs.
- **`catch {}` removal is a genuine fix, not a regression.** `processNestedCreate` performs no DB write (only hooks/validation/shaping); the single write is `strategy.persist` in `write-pipeline.ts`, which runs after all nested processing. So an access-denied existing row now throws and aborts before any write — it does NOT silently create a new row. The old `catch {}` had been swallowing the denial throw and downgrading it to a create. Confirmed by the `connectOrCreate ... denies unreachable rows` test asserting `student.create` was not called.
- **Empty access filter `{}`** correctly reduces `AND: [connection, {}]` to just the connection identifier (allow any matching existing row), consistent with `mergeFilters` semantics — not an accidental over-allow.

### Tests — strong
Allow + deny for scalar and nested-relation filters; read-not-update repro; AND/OR/some/none/not via DB reachability; connectOrCreate connect-branch deny + allow; sudo bypass. They assert the exact `findFirst({ where: { AND: [connection, filter] } })` shape and that denied-existing-row does not create. Good.

### Findings (ranked)
1. **Behavioral change (medium, document it):** update→query switch + `checkAccess` deny-by-default means a related list with `query: undefined` but a permissive `update` now denies nested connect (previously allowed). This is the right direction and consistent with reads being deny-by-default, but it's stricter than Keystone's allow-all operation default and should be called out in the changeset (currently it only describes the filter fix, not the phase-shift implication). Inline comment added.
2. **Optional hardening (low):** the filter `else` branch trusts the static type rather than validating with the existing `isPrismaFilter()` helper. Fails closed today (Prisma rejects bad input), so no security hole — just clarity. Inline comment added.

### Deferred owning-side field-level access (`// TODO(#578)`)
Acceptable for this PR's scope. The reported bug is false-denial from the wrong access phase + broken matcher, which this fully addresses. The deferral is clearly flagged with a tracking reference and a sound rationale (`NestedOpHandlerArgs` doesn't carry the owning field config). Not a blocker; recommend it stays tracked under #578/#581.

---

Issue #566, opened 2026-06-10
Explicit include bypasses relation-level access filtering on reads — caller include REPLACES the access-controlled include


borisno2:
## Summary

In `@opensaas/stack-core@0.24.0`, `findUnique`/`findMany` build an access-controlled include (`buildIncludeWithAccessControl` — per-relation `where` filters from the related list's `query` access, denied relations omitted, recursive to depth 5), but then:

```js
// dist/context/index.js:321 (findUnique) and :390 (findMany)
include = args.include || accessControlledInclude;
```

A caller-supplied `include` **replaces** the access-controlled one wholesale. Nested relations named in an explicit `include` are fetched with NO row-level access filter and NO denied-relation exclusion — phase-1 scoping is silently skipped for exactly the calls that ask for specific relations. (Phase-2 field stripping still runs, but which ROWS of a relation are visible is the phase-1 job.)

So the API is safest when used naively (bare read → fully filtered auto-include) and unsafe when used deliberately (explicit include → unfiltered relations) — inverted from what callers expect, and a divergence from Keystone, where nested GraphQL selections were always access-filtered.

## Repro shape

List B's `query` access returns a row filter (e.g. tenant scoping). `context.db.a.findMany({ include: { bRelation: true } })` on a non-sudo context returns ALL related B rows, including rows the same session cannot read via `context.db.b.findMany()`.

## Ask

Merge instead of replace: intersect the caller's `include` with the access-controlled constraints — apply the relation's access `where` to each caller-included relation (AND-ing with any caller-supplied nested `where`), drop relations whose query access is `false`, and recurse into nested includes. The caller's include should select WHICH relations to fetch, never WHETHER access filters apply.

## Severity

HIGH — silent row-level access bypass on any non-sudo read that passes `include`.

## Context

Found during an access-parity audit of a Keystone→stack migration (production app, 35 lists). Related: #564 (write-side fail-open for undeclared keys). Happy to provide a failing test case against the starter example if useful.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** A caller-supplied `include` (and a `query` fragment) on a read replaces the access-controlled relation include wholesale, so explicitly-requested relations are fetched with no row-level access filtering and no denied-relation exclusion — a silent row-level read bypass on any non-sudo read that names relations.

**Severity:** HIGH — silent cross-scope data exposure on reads.

**Current behavior:**
[...]

PR #589, opened 2026-06-27
fix(core): merge explicit include with access control on reads (#566)

borisno2:
Implements #566. Part of #581.

## Problem (HIGH — silent row-level access bypass)

On non-sudo `findUnique`/`findMany`, a caller-supplied `include` previously **replaced** the access-controlled include wholesale:

```js
include = args.include || accessControlledInclude
```

So any nested relation named in an explicit `include` was fetched with **no** per-relation access `where` filter and **no** denied-relation exclusion. The API was safest used naively (bare read → fully filtered auto-include) and unsafe used deliberately (explicit include → unfiltered relations).

## Fix — merge, not replace

New `mergeIncludeWithAccessControl` (in `packages/core/src/access/access-filter.ts`, exported from `access/index.ts`) intersects the caller's `include` with the access-controlled include. The caller's include selects WHICH relations to fetch; access control decides WHETHER and WITH WHAT filter.

For each relation the caller asks to include:
- **Denied relation** (query access `false`, so absent from the access-controlled include) → **dropped** (not fetched).
- **Allowed with a filter** → the relation's access `where` is applied, **AND-combined** with any caller-supplied nested `where` (caller can only narrow, never widen).
- **Bare `true`** → becomes the access-controlled shape (filter + nested filtered include), never bare `true`.
- **Nested includes** → recursively merged using the related list's field configs, so deeply-nested caller includes are access-filtered at every level (the access include is recursive to depth 5 via `buildIncludeWithAccessControl`; beyond that no auto-include exists and deeper caller keys pass through, matching existing behaviour).
- A key that is **not** a config-declared relationship is passed through unchanged (access control does not govern it — preserves existing behaviour).

### Behaviour preserved
- **Sudo**: caller `include` is used as-is (no filtering, nothing dropped) — sudo is never made more restrictive than before.
- **Query fragment path**: unchanged — the fragment-built include is still used directly; the merge only applies to the explicit `args.include` path.
- **`assertUniqueWhere` guard (#567)**: untouched — the merge is downstream of it.

## Tests

Added in `packages/core/tests/context.test.ts` (`#566` describe block), all red-before/green-after:
- `findMany`/`findUnique` with `include: { rel: true }` where `rel` query access returns a row filter → relation carries that access `where`, not bare `true`.
- A relation with query access `false` named in the caller include is dropped.
- A caller nested `where` is AND-combined with the access `where`.
- Nested (2-level) includes are access-filtered at each level.
- Sudo with explicit include returns the include unfiltered.
- The query-fragment path is unaffected.

## Verification
- `pnpm -C packages/core test` → 618 passed
- `pnpm -C packages/cli test` → 289 passed
- `pnpm build` (all 11 tasks), `pnpm lint` (0 errors), `pnpm manypkg fix`, `pnpm format` → clean
- Changeset: `@opensaas/stack-core` patch

---

### DIFF
--- a/.changeset/spotted-falcons-merge.md
+++ b/.changeset/spotted-falcons-merge.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix row-level access bypass when an explicit `include` is passed to non-sudo `findUnique`/`findMany`. The caller's `include` is now merged with (not replaced by) the access-controlled include: denied relations are dropped, each relation's access `where` is AND-combined with any caller nested `where`, and nested includes are filtered at every level. Sudo and query-fragment paths are unchanged.
--- a/.changeset/spotted-falcons-merge.md
+++ b/.changeset/spotted-falcons-merge.md
@@ -2,4 +2,4 @@
 '@opensaas/stack-core': patch
 ---
 
-Fix row-level access bypass when an explicit `include` is passed to non-sudo `findUnique`/`findMany`. The caller's `include` is now merged with (not replaced by) the access-controlled include: denied relations are dropped, each relation's access `where` is AND-combined with any caller nested `where`, and nested includes are filtered at every level. Sudo and query-fragment paths are unchanged.
+Fix row-level access bypass when an explicit `include` is passed to non-sudo `findUnique`/`findMany`. The caller's `include` is now merged with (not replaced by) the access-controlled include: denied relations are dropped, each relation's access `where` is AND-combined with any caller nested `where`, and nested includes are filtered at every level. Sudo and query-fragment paths are unchanged. When no access-controlled include is computed (inside a `resolveOutput`/virtual-field context, at max include depth, or for a list with no relationships), the caller's `include` is passed through unchanged rather than dropped — avoiding fail-closed data loss.
--- a/packages/core/src/access/access-filter.ts
+++ b/packages/core/src/access/access-filter.ts
@@ -95,3 +95,140 @@ export async function buildIncludeWithAccessControl(
 
   return hasRelationships ? include : undefined
 }
+
+/**
+ * A single relation entry in a Prisma `include` object: either a bare `true`
+ * (fetch with no extra constraints) or an object that scopes the fetch with a
+ * `where` filter and/or a nested `include`.
+ */
+type IncludeEntry = boolean | { where?: PrismaFilter; include?: IncludeObject }
+type IncludeObject = Record<string, IncludeEntry>
+
+/** The structured (object) form of a relation include entry. */
+type IncludeEntryObject = { where?: PrismaFilter; include?: IncludeObject }
+
+/**
+ * Narrow an unknown include value to the structured object form (vs bare `true`
+ * or any other primitive). Caller-supplied includes arrive untyped at the
+ * runtime boundary, so we validate the shape here rather than casting.
+ */
+function asEntryObject(value: unknown): IncludeEntryObject | null {
+  if (value && typeof value === 'object') {
+    const obj = value as Record<string, unknown>
+    const where = obj.where
+    const include = obj.include
+    const entry: IncludeEntryObject = {}
+    if (where && typeof where === 'object') entry.where = where as PrismaFilter
+    if (include && typeof include === 'object') entry.include = include as IncludeObject
+    return entry
+  }
+  return null
+}
+
+/**
+ * AND-combine an access `where` with a caller-supplied nested `where`.
+ *
+ * The access filter is authoritative: the caller's filter may only NARROW the
+ * result further, never widen past what access permits. We therefore wrap both
+ * in a Prisma `AND` so neither can override the other. If only one side is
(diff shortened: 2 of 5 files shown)


borisno2 (commented):
## Review: merge explicit `include` with access control (#566)

**Verdict: REQUEST CHANGES** (posted as a comment — the review API won't let me formally request changes on my own org token's PR). The security objective is met correctly, but the merge introduces a functional regression for reads issued inside `resolveOutput` hooks / virtual fields (and at MAX_DEPTH).

### What's right (security goal achieved)
I verified the core anti-bypass behaviour and it holds:
- Caller `include: { rel: true }` for a filtered relation now comes back with the access `where` applied, not bare `true`. ✅
- A relation whose query access is `false` (absent from the access include) is **dropped**. ✅
- Caller nested `where` is **AND-combined** with the access `where` (`{ AND: [accessWhere, callerWhere] }`) — no widening path exists. When access permitted all rows (bare `true`) a caller `where` only narrows. ✅
- The produced shape matches `buildIncludeWithAccessControl` (`true` vs `{ where, include }`) and `{ AND: [...] }` is valid Prisma for to-many relation `where`. ✅
- Sudo path uses the caller include as-is; query-fragment path is untouched; the `args.include` merge is downstream of the where-access checks. ✅
- No unsafe casts on public surfaces — `asEntryObject` validates the runtime boundary rather than casting blindly. Good.
- Tests are genuine red-before/green-after and cover the listed cases.

### Blocking finding (must fix)

**`packages/core/src/access/access-filter.ts` — `mergeIncludeWithAccessControl` drops ALL caller relations whenever `accessControlledInclude` is `undefined`.**

The merge does `const accessInclude = accessControlledInclude ?? {}`. When the access include is `undefined`, every declared relationship in the caller's `include` resolves to `accessValue === undefined` and hits the `continue` (drop) branch. But `buildIncludeWithAccessControl` returns `undefined` in three non-denial cases:

1. `_resolveOutputCounter.depth > 0` — i.e. **any read performed inside a `resolveOutput` hook or virtual field** (`access-filter.ts:43-45`; the counter is incremented around every hook in `field-visibility.ts:81`).
2. `depth >= MAX_DEPTH` (5) — `access-filter.ts:35`.
3. The list has no relationship fields at all (`hasRelationships` false) — `access-filter.ts:96`.

**Concrete failure (case 1):** a virtual field whose `resolveOutput` calls `context.db.post.findMany({ include: { comments: true } })`.
- Before this PR: `args.include || accessControlledInclude` → caller include honoured, `comments` fetched.
- After this PR: `accessControlledInclude` is `undefined` → `accessInclude = {}` → `comments` is **silently dropped**, so the hook receives posts with no comments. This breaks existing virtual-field / resolveOutput patterns that read with `include`.

This is a fail-closed *data-loss* regression (not a leak), so it does not reopen the bypass — but it will silently break working code. The same over-drop occurs one level down inside the recursion when an access entry is bare `true` (depth/hook-bailed) but the caller requested a nested include.

**Fix direction:** distinguish "access actively denied this relation" from "the builder declined to expand the include (depth cap / resolveOutput guard / no relationships)". When `accessControlledInclude === undefined`, the caller's `include` should pass through (matching pre-PR behaviour) rather than being treated as "everything denied". Inside the recursion, the same applies when `accessEntry?.include` is undefined but the relation is allowed.

### Test gap
No test exercises a non-sudo `include` query while `_resolveOutputCounter.depth > 0` (inside a `resolveOutput` hook) or at MAX_DEPTH — every `#566` test runs at top-level depth 0, so this regression passes CI undetected. Please add a test for an `include` issued from within a `resolveOutput` hook (and ideally a deep/MAX_DEPTH case).

---
Once the `undefined` access-include case passes the caller's selection through (instead of dropping it) and a regression test is added, this is good to merge — the bypass fix itself is correct.

*Reviewed from the PR diff only.*

---

borisno2:
Addressed the blocking regression: when `buildIncludeWithAccessControl` returns `undefined` (a NON-denial outcome — inside a `resolveOutput`/virtual-field context, at `MAX_DEPTH`, or for a list with no relationships), `mergeIncludeWithAccessControl` no longer treats it as "all relations denied". It now passes the caller `include` through unchanged, matching the prior `args.include || …` fallback and fixing the fail-closed data loss where, e.g., a `resolveOutput` hook calling `context.db.post.findMany({ include: { comments: true } })` had its `comments` silently dropped.

The distinction is precise: a whole-object `undefined` access include = passthrough; an `undefined` ENTRY inside a DEFINED access include still means the relation was denied and is dropped (the real merge — denied-relation drop, access `where`, AND-combined caller nested `where` — is untouched).

Regression tests added in the `#566` block exercising the `accessControlledInclude === undefined` path via `findUnique`/`findMany` issued from within a `resolveOutput` hook (bare and nested caller include). Both are red-before / green-after. All existing #566 tests still pass.

Full suite green: core 620, cli 289, ui 190, auth 133, rag 178, storage 123, storage-s3 43, storage-vercel 30, create-opensaas-app 41. `pnpm lint` 0 errors.

---

Issue #580, opened 2026-06-27
[PRD] Generated .opensaas bundle: type-clean & loadable in every runtime


borisno2:
> *This was generated by AI during triage.*

## PRD — Generated `.opensaas` bundle: correctness & loadability

**Goal:** The generated `.opensaas` bundle (context/types/lists/extensions + prisma-client subtree) should type-check clean and load in every target runtime — bundler (Next/webpack) and plain Node — with consistent, statically-resolvable import extensions. Today it is `.ts`-source-only, has inconsistent import extensions, and carries one residual type error, which forces `@ts-nocheck` and blocks Node-only consumers (e.g. the better-auth live-adapter swap).

**Why these are grouped:** all three touch the CLI generator's bundle emission and module-resolution story; they share the same files and decisions (especially the import-extension strategy) and would collide if worked independently.

### Child issues
- **#559** (bug, ready-for-agent) — keystone-columns `image()`/`file()` emit `needs: { <field>: true }` typed `never`; the last thing forcing `@ts-nocheck` on the bundle.
- **#560** (enhancement, ready-for-agent) — default `importFileExtension="ts"` for the prisma-client subtree + a `db.prismaGeneratorOptions` passthrough, so the whole graph is statically resolvable.
- **#579** (enhancement, ready-for-human) — a sanctioned Node-loadable build of the bundle (compiled `.mjs`/`.cjs` emit or `package.json` `exports`/`type:module`) so a live module (auth) loads it across bundle + plain Node + build-time.

### Suggested sequencing
1. **#559** and **#560** in parallel (agent) — produce a type-clean, extension-consistent bundle.
2. **#579** (human) builds on #560's extension/module-resolution decisions to add the Node-loadable form.

### Definition of done
- The generated bundle passes `tsc --noEmit` with no `@ts-nocheck`.
- The bundle (including the prisma-client subtree) is statically resolvable by a bundler.
- A live module can `import()` the generated context under plain Node, verified by a better-auth-creates-user e2e in a non-bundled runtime.

borisno2:
## QA sign-off — PRD #580: Generated `.opensaas` bundle: type-clean & loadable in every runtime

**Verdict: ✅ PASS**

[...]

Issue #559, opened 2026-06-10
Generated prisma-extensions.ts types `needs: { <field>: true }` as `never` for storage fields in db.columns:'keystone' mode


borisno2:
## Summary

With `@opensaas/stack-*@0.24.0`, the generated `.opensaas/prisma-extensions.ts` fails `tsc --noEmit` when a list uses a storage `image()`/`file()` field in **multi-column keystone mode** (`db: { columns: 'keystone' }`).

## Repro

A list with:

```ts
image({ storage: 'productionImages', db: { columns: 'keystone' } })
```

In keystone-columns mode the Prisma model has only the seven `image_*` part columns (`image_filesize`, `image_pathname`, …) — there is **no `image` scalar** on the model. But the generated result-extension still emits:

```ts
needs: { image: true }
```

Prisma's extension typing resolves `image` against the model's scalar fields, so the literal `true` is typed against `never`:

```
prisma-extensions.ts(354,18): error TS2322: Type 'true' is not assignable to type 'never'.
prisma-extensions.ts(382,18): error TS2322: Type 'true' is not assignable to type 'never'.
```

(Two errors = two lists using the field — `Teacher` and `Production` in our app.)

## Why it matters

Since #553 the generated bundle is meant to be statically imported by the host (`import { getContext } from '@/.opensaas/context'`), which pulls it into the host's type-check graph. These two errors are now the **only** thing forcing hosts to stamp `// @ts-nocheck` onto generated files — everything else in the 0.24.0 bundle type-checks clean (verified empirically: stripped all `@ts-nocheck` from a generated bundle and ran `tsc --noEmit`; only these 2 errors remain).

## Ask

In keystone-columns mode, the extension's `needs` should reference the part columns that actually exist on the model (e.g. `needs: { image_pathname: true, image_url: true, … }`) — or the virtual-assembly extension should be typed against the assembled output rather than model scalars.

## Context

Found completing the Keystone→stack migration of a production app (35 lists, multi-schema Postgres). Follows SF-14/#548/#553.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Generated Prisma result extensions emit `needs: { <field>: true }` for storage `image()`/`file()` fields in multi-column (`db: { columns: 'keystone' }`) mode, where no such scalar exists on the model — producing `TS2322: Type 'true' is not assignable to type 'never'` and forcing `@ts-nocheck` on the generated bundle.

[...]

PR #591, opened 2026-06-27
fix(cli): derive keystone-mode storage `needs` from part columns

borisno2:
Implements #559. Part of #580.

## Problem

In `db: { columns: 'keystone' }` mode, storage `image()`/`file()` fields are stored as multiple part columns on the Prisma model (e.g. `image_url`, `image_filesize`, `image_pathname`, …) with **no** single `image` scalar. But the generated `.opensaas/prisma-extensions.ts` still emitted a result extension with `needs: { image: true }`. Prisma types `needs` against the model's scalar fields, so the literal `true` was typed against `never`:

```
prisma-extensions.ts(354,18): error TS2322: Type 'true' is not assignable to type 'never'.
```

These were the last errors forcing `@ts-nocheck` on the generated bundle.

## Fix

In `packages/cli/src/generator/prisma-extensions.ts`, when a field exposes `getColumnNames` (the multi-column contract that `image()`/`file()` implement in keystone-columns mode):

- `needs` is built from `getColumnNames(fieldName)` — each resolved part column → `true` — instead of `{ [fieldName]: true }`.
- `compute` no longer reads the non-existent logical scalar (`production.image`); it returns `undefined`. The logical value is re-assembled by core's read pipeline (`assembleColumns`) and re-added via `TransformedFields`, so runtime behaviour is unchanged (the scalar was already absent → `undefined`).

### How the part-column names are derived

`getColumnNames` is the single source of truth: it already returns the resolved/overridden physical names (honouring custom `db: { columns: { mode: 'keystone', map: { url: 'custom_name' } } }` maps). The fix derives `needs` directly from it and does **not** hardcode the `_url`/`_pathname`/… suffixes. This mirrors the existing pattern in `packages/cli/src/generator/types.ts`. The emitted `needs` keys exactly match the `@map`-ped columns the Prisma generator produces for the same field.

For an `image` field named `image`:
```
needs: { image_url: true, image_width: true, image_height: true, image_filesize: true, image_contentType: true, image_contentDisposition: true, image_pathname: true }
```
For a `file` field named `doc`:
```
needs: { doc_filename: true, doc_filesize: true, doc_url: true }
```

### Unchanged

- Single-column (non-keystone) storage fields and all other fields still emit `needs: { <fieldName>: true }`.
- Virtual fields still emit `needs: {}`.

## Tests

Added `packages/cli/src/generator/prisma-extensions.test.ts` covering: keystone-mode image and file fields (part columns), custom/overridden column names flowing through `getColumnNames`, single-column fields unchanged, and virtual fields unchanged. Full cli suite: 294 passing. `tsc` (build) clean.

https://claude.ai/code/session_01RyPQQD4zvAiBFJJyKYhvCH

---
_Generated by [Claude Code](https://claude.ai/code/session_01RyPQQD4zvAiBFJJyKYhvCH)_

### DIFF
--- a/.changeset/wise-points-cheer.md
+++ b/.changeset/wise-points-cheer.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+Fix `tsc` failure in generated `prisma-extensions.ts` for multi-column storage fields in `db: { columns: 'keystone' }` mode. The result extension's `needs` now references the physical part columns (e.g. `image_url`, `image_pathname`, …) derived from the field's `getColumnNames`, instead of the logical field name which has no scalar on the model (previously typed `true` against `never`). This removes the last error forcing `@ts-nocheck` on the generated bundle (#559).
--- a/packages/cli/src/generator/prisma-extensions.test.ts
+++ b/packages/cli/src/generator/prisma-extensions.test.ts
@@ -1,0 +1,146 @@
+import { describe, it, expect } from 'vitest'
+import { generatePrismaExtensions } from './prisma-extensions.js'
+import type { OpenSaasConfig, FieldConfig } from '@opensaas/stack-core'
+import { text, virtual } from '@opensaas/stack-core/fields'
+
+/**
+ * Builds a synthetic multi-column storage field (the shape image()/file()
+ * produce in `db: { columns: 'keystone' }` mode). The CLI package does not
+ * depend on @opensaas/stack-storage, so we mimic the relevant surface directly:
+ * a field with a `resultExtension` (so it participates in the result extension)
+ * and a `getColumnNames` that returns the physical part columns. See #559.
+ */
+function multiColumnStorageField(columns: string[]): FieldConfig {
+  return {
+    type: 'image',
+    resultExtension: {
+      outputType: "import('@opensaas/stack-storage').ImageMetadata | null",
+    },
+    getColumnNames: (fieldName: string) => columns.map((part) => `${fieldName}_${part}`),
+  }
+}
+
+describe('Prisma Extensions Generator', () => {
+  it('derives `needs` from the part columns for keystone-mode image fields', () => {
+    const config: OpenSaasConfig = {
+      db: { provider: 'postgresql' },
+      lists: {
+        Production: {
+          fields: {
+            image: multiColumnStorageField([
+              'url',
+              'width',
+              'height',
+              'filesize',
+              'contentType',
+              'contentDisposition',
+              'pathname',
+            ]),
+          },
+        },
+      },
+    }
+
+    const output = generatePrismaExtensions(config)
+
+    // needs references the physical part columns that actually exist on the
+    // model, NOT the logical `image` field (which has no scalar in this mode).
+    expect(output).toContain(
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review verdict: APPROVE (posting as COMMENT — GitHub blocks self-approval by the authoring identity)

Reviewed strictly from the PR diff plus the surrounding code it touches (`getColumnNames` contract, core's `assembleColumns` read pipeline). This is a clean, correctly-scoped bug fix that removes the last `@ts-nocheck`-forcing error in the generated `prisma-extensions.ts`.

### What it does
For multi-column storage fields (`image()`/`file()` in `db: { columns: 'keystone' }` mode), the generated result extension previously emitted `needs: { image: true }` — but no `image` scalar exists on the model, so Prisma typed the `true` literal against `never` (TS2322). The fix derives `needs` from the field's part columns and has `compute` return `undefined`, deferring the logical value to core's read pipeline.

### Verified correct
- **`needs` is derived from the contract, not hardcoded suffixes.** It calls `fieldConfig.getColumnNames(fieldName)` (`packages/core/src/config/types.ts:539`, `(fieldName: string) => string[]`). The storage impl resolves through `resolveFileColumnMap`/`resolveImageColumnMap` (`packages/storage/src/utils/multi-column.ts`), so custom `db: { columns: { map: {...} } }` overrides flow through. Mirrors the existing `packages/cli/src/generator/types.ts:485-487` usage. ✔
- **Single-column and virtual cases unchanged.** The `isMultiColumn` guard (`!isVirtual && typeof fieldConfig.getColumnNames === 'function'`) correctly excludes single-column fields (still `needs: { <fieldName>: true }`, `compute` reads the scalar) and virtual fields (still `needs: {}`). ✔
- **`compute` → `undefined` is a true no-op at runtime.** Old code did `const value = production.image` → `image` is absent → `undefined` → `return undefined`. New code returns `undefined` directly. Either way, core's `assembleColumns` (`packages/core/src/access/field-visibility.ts:128-141`) runs *after* the Prisma result extension (the extended client is what core queries — `context/index.ts:558` → `filterReadableFields` at `:569`) and **unconditionally overwrites** `workingItem[fieldName]` with the assembled value. The result extension cannot clobber it; the new `needs` is in fact what guarantees the part columns are present so assembly fires. ✔
- **Conventions.** No `any`/type-casts added to exported types (the pre-existing `null as any` is untouched); relative imports use `.js` ESM extensions; changeset present at `patch` (appropriate for a bug fix). ✔
- **Tests.** New `prisma-extensions.test.ts` asserts the exact part-column `needs` keys for both image and file, custom/overridden names flowing through, single-column unchanged (`needs: { title: true }` + `const value = post.title`), and virtual unchanged (`needs: {}`). ✔

### Non-blocking notes
- The test builds a synthetic multi-column field locally rather than importing `@opensaas/stack-storage` — intentional, since the CLI package has no such dependency. The runtime assembly path is covered separately by `packages/core/src/access/multi-column-read-write.test.ts`, so combined coverage is adequate.
- `getColumnNames` is a generic `BaseFieldConfig` contract (any field could implement it), and the guard treats it as "owns multiple physical columns" — the right interpretation. Today only storage fields implement it, always alongside `assembleColumns`, so there's no gap.

No correctness, regression, or convention issues found. LGTM.

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

---

Issue #590, opened 2026-06-27
Add beforeTransaction / afterTransaction transaction-boundary hooks (out-of-transaction side effects)


borisno2:
Follow-up to #569 / ADR-0010 (part of the #581 access-fidelity effort). #569 introduces a transaction boundary around every write (decompose nested writes into per-record full-pipeline runs inside one `prisma.$transaction`). This issue adds a **new** hook pair that runs *outside* that transaction.

## Motivation
`beforeOperation`/`afterOperation` now run **inside** the transaction (they roll back with it), so they are unsuitable for non-transactional side effects like external API calls — you don't want to hold a DB transaction open across a network call, and such a call can't be rolled back. We need hooks that bracket the transaction from outside.

## Design (agreed in the #569 grill; see CONTEXT.md glossary "Transaction-boundary hooks")
Add list- and field-level `beforeTransaction` / `afterTransaction` hooks:
- **`beforeTransaction`** runs *before* the transaction opens.
- **`afterTransaction`** runs *after* the transaction settles and **always runs**, receiving the outcome:
  - `status: 'committed' | 'rolled-back'`
  - `error?` — the error that caused the rollback (absent on commit)
  - on committed: the persisted `item` (and `originalItem` for update/delete), like `afterOperation`
  - on rolled-back: no `item`; gets `inputData`/`resolvedData` and `error` so it can compensate for whatever `beforeTransaction` did externally.
- **Per list involved** in the write (including nested lists), mirroring how `beforeOperation`/`afterOperation` fire per record.
- **Symmetric-bracket "always-run" rule:** `afterTransaction[L]` runs **iff `beforeTransaction[L]` already ran** — so every external action taken has its paired compensator; lists whose `beforeTransaction` never ran get neither. A `beforeTransaction` throwing aborts the write (transaction never opens) and triggers `afterTransaction` (status `rolled-back`) only for the lists whose `beforeTransaction` already completed.
- If an `afterTransaction` hook itself throws, still run the remaining `afterTransaction` hooks (all compensations fire), then surface the error(s) to the caller. The DB state is already final.

## Acceptance criteria
- [ ] `beforeTransaction`/`afterTransaction` available at list and field level, typed, with the args above.
- [ ] `afterTransaction` always runs and receives `committed | rolled-back` + `error`; persisted `item` present only on commit.
- [ ] Per-list firing (incl. nested lists), symmetric-bracket rule honored on `beforeTransaction` failure and on DB rollback.
[...]

Issue #569, opened 2026-06-10
Nested relation writes skip beforeOperation/afterOperation hooks; nested connect requires update-access on the target list


borisno2:
## Summary

Two semantic divergences from Keystone in `@opensaas/stack-core@0.24.0`'s nested-operations handling (`dist/context/nested-operations.js`), found while auditing a migration:

1. **Nested creates run a partial hook pipeline.** A nested `{ relation: { create: {...} } }` runs the target list's `resolveInput`/`validate` but NOT `beforeOperation`/`afterOperation`. Keystone ran the full hook pipeline for nested writes. Any list whose side effects (workflows, notifications, billing) hang off before/afterOperation silently loses them when a record is created nested instead of top-level — same logical operation, different side effects depending on call shape.

2. **Nested `connect` requires `update` operation access on the TARGET list.** Keystone required read access on the target (+ field-level access on the relation field). Requiring update-access on the target means e.g. "attach an existing tag to my post" needs update rights on Tag — stricter in a way that breaks legitimate flows, while (1) is looser in a way that drops side effects.

## Ask

Run the full hook pipeline for nested creates/updates/deletes (or document loudly that nested writes are hook-partial and recommend top-level writes for hook-bearing lists), and align/document the connect access rule (read + field-gate is the Keystone-compatible semantic).

## Context

Production Keystone→stack migration; the audit's concrete instances were benign today only by luck (the nested-created list had no operation hooks yet). Related: #564, #566, #568.

borisno2:
**Confirmed live-flow casualty (upgrading from "benign by luck").** While completing a Keystone→stack migration we hit this as a real broken user feature, not a latent risk.

The account-holder self-enrolment flow does: `context.db.student.update({ data: { enrolments: { create: { lessonTerm: { connect: { id } }, ... } } } })` on a NON-sudo (ACCOUNT-role) context. `processNestedConnect` (`dist/context/nested-operations.js:95-96`) checks the connected `LessonTerm`'s `access.operation.update` (admin-only here) and throws `Access denied: Cannot connect to this item`. Under Keystone the same flow worked because connecting checked the related list's READ/query access (`LessonTerm` query = allowAll), not update.

So the migration faithfully ports every access rule yet the feature breaks, purely because nested-connect uses the wrong access phase. A non-admin legitimately allowed to READ a `LessonTerm` (and create an Enrolment referencing it) cannot connect it.

[...]

PR #593, opened 2026-06-27
feat(core): run full hook pipeline for nested writes inside one transaction (#569)

borisno2:
## Summary

Implements #569. Nested relation writes (`create`/`update`/`delete`) now run the **same full hook pipeline** as the equivalent top-level write — list- and field-level `beforeOperation`/`afterOperation`, not just `resolveInput`/`validate`/field-rules — so side effects (workflows, notifications, billing) are identical whether a record is written nested or top-level.

The whole operation (parent + all nested writes) now runs inside **one `prisma.$transaction`**, so it is atomic, and a throwing `beforeOperation`/`afterOperation` (or validation) **rolls the whole write back**.

Part of #581. Decision: ADR-0010. Follow-up: #590.

## Observable contract delivered

1. **Full hook pipeline for nested writes** — nested create fires `beforeOperation` (create) → persist → `afterOperation` with the created `item`; nested update fires `afterOperation` with `originalItem` + updated `item`; nested delete fires before/after with `originalItem`. Side effects are observably identical nested vs top-level.
2. **One transaction per operation, atomic** — every write (not only nested) runs inside one interactive transaction; the `tx` client is the persistence target for parent + nested.
3. **`afterOperation`/`beforeOperation` throw rolls back** — hooks run inside the transaction.
4. **Preserved semantics** — operation/field access, validation, silent-failure on reads, sudo bypass, Field Visibility, MAX_DEPTH 5, and the #578 nested-`connect`/`connectOrCreate` read-access + DB-reachability behavior are unchanged.

## Mechanism (per ADR-0010)

"Hooks around a single nested persist": Prisma's single nested write still performs persistence (Prisma keeps owning FK ordering / intra-statement atomicity). Around it, the pipeline runs each nested record's `beforeOperation` before the persist and registers a deferred `afterOperation` task that fires after, all against the transaction client.

- **`originalItem`** (nested update/delete): the row fetched for access doubles as `originalItem`.
- **`item`** (nested create/update): the parent write `include`s the affected relations, and the persisted nested row is recovered from that result.
- **Transaction threading**: `runWritePipeline` opens `prisma.$transaction`; the `tx` client replaces the prior `getModel(prisma, …)` target for the parent and is passed to `processNestedOperations` for all nested reads/persistence. When a client has no `$transaction` (test mocks), it falls back to running directly — hook ordering/args are identical; only the rollback guarantee needs a real transaction.

ADR-0010's prose was updated to describe this realized mechanism (the decision — single transaction, full hooks for nested, every write transactional, throw-rolls-back, transaction-boundary hooks deferred — is unchanged).

## Out of scope

- `beforeTransaction`/`afterTransaction` transaction-boundary hooks — Follow-up: #590.
- Explicit `rollback()` helper — throw-to-rollback is the mechanism.
- Pass-through nested kinds (`disconnect`/`set`/`updateMany`/`deleteMany`) behave as before.

## Tests

New `packages/core/tests/nested-write-hooks.test.ts` (transaction-aware in-memory Prisma mock): nested create fires list+field before/after with correct args; nested update afterOperation gets `originalItem`+`item`; nested delete before/after get `originalItem`; side effect identical nested vs top-level; throwing nested afterOperation **and** beforeOperation roll back the parent write; non-sudo nested-create access denial rolls back; sudo bypasses access but still runs hooks; every write uses `$transaction`.

Updated `nested-operation-registry.test.ts`: `delete` is no longer a pass-through kind (now runs the delete hook pipeline), so it is asserted separately; `deleteMany`/`set`/`updateMany` remain pass-through.

## Behavior change (called out in changeset)

An `afterOperation` that throws previously left the row committed; it now rolls back with the transaction. Every write is now transactional.

## Verification

- `pnpm lint` — no new errors (only pre-existing warnings)
- `pnpm build` — all packages
- `pnpm -r test` — all packages green (core 637, cli 289, plus auth/ui/rag/storage*)

---


packages/core/src/context/nested-operations.ts
```diff
@@ -18,15 +72,70 @@ function isRelationshipField(fieldConfig: FieldConfig | undefined): boolean {
...
+  if (included == null) return undefined
+
+  if (Array.isArray(included)) {
+    const rows = included as Array<Record<string, unknown>>
+    if (knownId !== undefined) {
+      const match = rows.find((r) => r.id === knownId)
+      if (match) return match
+    }
+    const fresh = rows.filter((r) => typeof r.id === 'string' && !excludeIds.has(r.id as string))
+    if (fresh.length > 0) return fresh[fresh.length - 1]
```
borisno2: **Blocker (correctness): multiple nested creates on a to-many relation all recover the SAME row for `afterOperation`.**

For `posts: { create: [{title:'A'},{title:'B'}] }`, `processNestedCreate` pushes one afterTask per item, all with `knownId=undefined` and the same `excludeIds` Set. Creates never add their new ids to `excludeIds` (only update/delete do), so at run time `excludeIds` is empty. In `recoverPersistedRow`:

- `knownId===undefined` → the id-match branch is skipped
- `fresh = rows.filter(... !excludeIds.has(r.id))` → **every** row (both A and B, plus any pre-existing)
- returns `fresh[fresh.length-1]` → the last row

So both A's and B's `afterOperation` / field `afterOperation` / deeper `childAfterTasks` receive the *same* (last) row, not their own. Any after-hook reading `item.id` (search index, webhook, audit log, CDN invalidation) runs against the wrong record for every create except the coincidentally-last one. `inputData`/`resolvedData` are still per-item correct, so the mismatch is silent.

The recovery is only reliable for to-one and for single-create. Creates have no per-item identity to recover by; the heuristic can't disambiguate siblings. Recommend recovering created rows by some stable identity (e.g. capture pre-existing ids per field before the write into `excludeIds`, then pair each create with a distinct fresh id) or, per ADR-0010's rejected "full decomposition" option, fetch each created row by its returned id.

---

packages/core/src/context/nested-operations.ts
```diff
@@ -18,15 +72,70 @@ function isRelationshipField(fieldConfig: FieldConfig | undefined): boolean {
...
+  const included = parentResult[fieldName]
+  if (included == null) return undefined
+
+  if (Array.isArray(included)) {
+    const rows = included as Array<Record<string, unknown>>
+    if (knownId !== undefined) {
+      const match = rows.find((r) => r.id === knownId)
+      if (match) return match
+    }
+    const fresh = rows.filter((r) => typeof r.id === 'string' && !excludeIds.has(r.id as string))
```
borisno2: **Blocker (correctness): a nested create on a to-many relation that already has pre-existing linked rows can recover a PRE-EXISTING row as the "created" item.**

`excludeIds` is created fresh per field and starts empty; creates and `connect` never populate it (the `NestedOpHandlerArgs.excludeIds` doc claims it holds "connect targets" but `processNestedConnect` neither receives nor populates it). The parent write `include`s the whole relation, so `parentResult[fieldName]` returns pre-existing rows + the new one. With `excludeIds` empty, `fresh` includes the pre-existing rows, and `recoverPersistedRow` returns `fresh[fresh.length-1]` — whichever row Prisma returns last.

Prisma does **not** guarantee `include` ordering (no `orderBy`), so the genuinely-created row is not reliably last. Example: a `User` with 5 existing `posts`, then `user.update({ data: { posts: { create: [{title:'New'}] } } })` — the create's `afterOperation` can fire with one of the 5 pre-existing posts as `item`.

Fix together with the multi-create issue: snapshot the relation's pre-existing ids into `excludeIds` before the write (and include connect targets), so `fresh` is exactly the newly-created rows.

---

packages/core/src/context/write-pipeline.ts
```diff
@@ -112,6 +123,43 @@ function getModel<TPrisma extends PrismaClientLike>(
...
+  if (typeof client.$transaction === 'function') {
+    return (await client.$transaction(async (tx) => fn(tx as TPrisma))) as Record<
+      string,
+      unknown
+    > | null
+  }
+  return fn(prisma)
+}
+
 /**
```
borisno2, line 168: **Atomicity hole vs ADR-0010: user hooks that call `context.db.*` write OUTSIDE this transaction.**

`runInTransaction` correctly threads `tx` as `args.prisma` and into `processNestedOperations` (`{ ...context, prisma: tx }`). But the `context.db` operations are built once in `getContext` closing over the **original** `prisma` (context/index.ts ~L311-331), and `context.db` is never rebuilt from `tx`. Spreading `{ ...context, prisma: tx }` overrides `context.prisma` but not `context.db`.

So a `beforeOperation`/`afterOperation` that does `await context.db.foo.create(...)` runs against the original client — a *separate* interactive transaction — which will **not** roll back when this transaction aborts. ADR-0010 states "every write is atomic / a throw rolls the whole write back"; that guarantee holds for the relation payload the pipeline owns, but not for hook-initiated `context.db` writes. Given the ADR's strong atomicity framing, this gap should either be closed (rebuild `context.db` against `tx` for the hook context) or explicitly documented as a limitation in ADR-0010.

---

packages/core/src/context/write-pipeline.ts
```diff
@@ -112,6 +123,43 @@ function getModel<TPrisma extends PrismaClientLike>(
...
  * Check if a list is configured as a singleton.
  */
```
borisno2, line 170: **Nit (robustness): the no-`$transaction` fallback is silent.** `return fn(prisma)` runs whenever `$transaction` isn't a function. Real `PrismaClient` (and the driver-adapter/`$extends` clients in CLAUDE.md) all expose `$transaction`, so production should not hit this — but if any wrapper ever drops it, atomicity is lost with zero signal. A one-time `console.warn` here (mirroring the `select`-ignored warning pattern) would make a misconfigured production client detectable rather than silently non-atomic.

---

packages/core/src/context/nested-operations.ts
```diff
@@ -102,14 +212,68 @@ async function processNestedCreate(
...
+      })
+
+      // 10. Register afterOperation: fires once the parent (and thus this nested
+      // row) has persisted. The persisted row is recovered from the parent's
+      // included relation.
+      afterTasks.push({
+        fieldName,
+        run: async (parentResult) => {
+          const persisted = recoverPersistedRow(parentResult, fieldName, undefined, excludeIds)
+          const createdItem = persisted ?? {}
```
borisno2: **Minor (silent degradation):** `const createdItem = persisted ?? {}`. When recovery returns `undefined` (relation not included / empty array — see the heuristic issues above), `afterOperation` and field `afterOperation` fire with `item = {}` (no `id`). A hook keying off `item.id` (cache key, search-doc id, webhook payload) silently gets `undefined`. There's no invariant check that a created `item` has an id. Consider failing loudly (or at least warning) when a `create`/`update` afterTask can't recover its persisted row, rather than passing `{}`. (The update path's `persisted ?? originalItem` fallback is safer but returns a *stale* pre-update row, so `originalItem` vs `item` diffs vanish.)

---

packages/core/tests/nested-write-hooks.test.ts
```diff
@@ -0,0 +1,498 @@
...
+        item: expect.objectContaining({ id: 'c1', body: 'doomed' }),
+      }),
+    )
+    expect(listAfter).toHaveBeenCalledWith(
+      expect.objectContaining({
+        operation: 'delete',
+        originalItem: expect.objectContaining({ id: 'c1', body: 'doomed' }),
+      }),
+    )
+  })
```
borisno2, line 300: **Test coverage gap.** The suite is solid on the headline guarantees — the rollback tests use a `$transaction` mock that genuinely snapshots/restores tables on throw (so they really prove atomicity, not just propagation), and the nested-vs-top-level parity test is good. But every nested-create test creates exactly **one** related row, which is precisely the case the `recoverPersistedRow` heuristic handles correctly. Please add:

1. A nested create of **multiple** rows on a to-many relation (`create: [{...},{...}]`) asserting each row's `afterOperation` receives its **own** `item` — this would catch the "all siblings get the last row" bug.
2. A nested create on a to-many relation that **already has pre-existing rows**, asserting the created row (not a pre-existing one) is passed to `afterOperation`.

Both currently fail with the heuristic as written.

---

borisno2 (commented):
## Review — nested writes: full hook pipeline in one transaction (#569)

**Verdict: REQUEST CHANGES** (submitted as COMMENT — GitHub blocks REQUEST_CHANGES on one's own PR; treat this as a change request).

The core architecture is sound and well-documented. The transaction threading is correct: `runWritePipeline` opens one interactive `$transaction`, `tx` is the persistence target for parent + all nested writes, and there is **no try/catch inside the transaction body** — a throwing nested `beforeOperation`/`afterOperation` (or validation/access denial) propagates out of the `$transaction` callback, so Prisma rolls everything back. The rollback tests genuinely prove this (the mock snapshots and restores on throw). ADR-0010 honestly describes the realized "hooks around a single nested persist" mechanism, the changeset correctly documents the behavior change (afterOperation-throw-rolls-back; every write transactional) as a `minor` bump, and the delete-pipeline change + `deleteMany`/`set`/`updateMany` pass-through assertions are correct. Sudo-bypasses-access-but-runs-hooks, field-level hook parity, and MAX_DEPTH 5 are all preserved.

Requesting changes for the **item-recovery heuristic**, the load-bearing weakness.

### Must fix (correctness)
1. **Multiple nested creates on a to-many relation all get the SAME `item`** (`recoverPersistedRow`). With N≥2 creates, every created row's `afterOperation` receives `fresh[last]` (the same row), not its own — creates have no `knownId` and never populate `excludeIds`. Verified by walkthrough. Inline comment.
2. **Nested create on a to-many relation with pre-existing rows can recover a PRE-EXISTING row.** `excludeIds` starts empty and is never seeded with pre-existing/connect ids, so `fresh[last]` can be a pre-existing row; Prisma `include` ordering is not guaranteed. Inline comment.

Both are fixable by snapshotting the relation's pre-existing ids (plus connect targets) into `excludeIds` before the write so `fresh` is exactly the new rows — though recovering each created row by its returned id (closer to ADR-0010's rejected full-decomposition path) is more durable.

### Should address
3. **Atomicity gap vs ADR-0010:** `context.db` is built over the original `prisma` and is not rebuilt from `tx`; `{ ...context, prisma: tx }` overrides `context.prisma` but not `context.db`. A hook calling `context.db.x.create()` writes in a *separate* transaction that won't roll back with the parent. Either rebind `context.db` to `tx` for the hook context, or document this limitation in ADR-0010 (whose framing currently implies full atomicity).
4. **Silent `item = {}` fallback** on failed create-row recovery — `afterOperation` runs with no `id`. Fail loud or warn.
5. **Test gap:** add multi-create and create-with-pre-existing-rows cases (both fail today). Inline comment.

### Nits
- Silent no-`$transaction` fallback could emit a one-time warn.
- `NestedOpHandlerArgs.excludeIds` doc says it holds "connect targets" — it does not (`processNestedConnect` never populates it).

The transaction/rollback core is correct; findings 1 & 2 are the blockers because they silently feed wrong/empty rows to nested `afterOperation` side effects — the exact parity guarantee this PR exists to deliver.

---

borisno2:
Addressed the blocking review findings (pushed as `edcc8af`).

**Findings 1 + 2 — created-row recovery (item correctness).** Replaced the `fresh[last]`/`excludeIds` heuristic in `recoverPersistedRow` with **id-diff recovery**:
- Before the parent persist, for each to-many relation that has nested `create`s, the pipeline captures the set of related-row ids already linked to the parent (in-transaction read; empty when the parent is itself being created) via `capturePreExistingIds`.
- After the persist, the created rows are exactly the included rows whose ids are **not** in that captured set (`recoverCreatedRows`).
- Each nested create now fires its `afterOperation` **once per newly-created row, against its OWN distinct row** — a to-many `create: [{A},{B}]` no longer collapses to one row, and a pre-existing sibling is never passed as the "created" item.
- Updated rows are recovered by their known id (`recoverUpdatedRow`). Created rows are paired to create-payload entries by **position** (Prisma preserves create-array order), so `item` is always correct; `inputData↔item` pairing is positional/best-effort (documented in code).
- **Finding 4:** if a created row can't be identified by id-diff (the client didn't echo the `include`d relation), the task **skips** rather than firing `afterOperation` with a fabricated id-less `{}` (documented). The misleading `excludeIds`/"connect targets" doc comment (finding 6) is gone.

**Finding 3 — `context.db` atomicity (fixed, not documented-around).** Chose to **bind `context.db` to the transaction** rather than weaken the ADR. The request-time `db` delegates capture the non-transactional client, so I extracted the delegate construction into `buildDbDelegate` (reused by `getContext`) and the Write Pipeline now rebuilds the context's `db`/`prisma` against the tx client (`bindContextToTransaction`) for the duration of the write. A `context.db` write performed inside a `before`/`afterOperation` hook now participates in — and rolls back with — the same transaction. Plugin runtimes are not re-executed. ADR-0010 + the existing changeset were updated to state this.

**Tests (finding 5)** added to `nested-write-hooks.test.ts` (with a richer to-many mock that echoes `include` and tracks links):
- Multiple to-many nested creates → each created row's `afterOperation` fires exactly once with its own distinct row (verified failing against the old `fresh[last]` logic).
- Nested create on a to-many relation with pre-existing rows → fires only for the new row, never a pre-existing one.
- A hook writing via `context.db` rolls back when the transaction later throws, and commits when it succeeds (covers finding 3).

The transaction/rollback machinery from the original PR is untouched. Existing tests stay green; no test asserted the old recovery contract, so none needed updating.

Full suite green: core 641, cli 289, ui 190, rag 178, auth 133, storage-s3 43, storage-vercel 30. `pnpm lint` (no new warnings), `pnpm format`, `pnpm manypkg fix` clean.

One irreducible limitation: `inputData`↔created-row pairing for to-many creates is by array position (the only reliable signal Prisma's nested-write result gives us). `item` correctness is guaranteed regardless.

---

borisno2 (commented):
## Re-review: fix delta `d1e8672..edcc8af` — APPROVE (posted as comment; cannot formally approve own PR)

Focused re-review of only the fix commit (`edcc8af`) addressing the three blockers. The transaction/rollback core (verified in the prior pass) was not re-checked. All prior blockers are resolved and I found no regressions.

### Blockers — resolved

**#1 + #2 created-row recovery (`nested-operations.ts`)** — Resolved correctly.
- `capturePreExistingIds` (L191) reads the parent's current related ids via the `prisma` threaded down, which is the tx client (`processFieldNestedOps` ← `args.prisma` ← `runWriteInTransaction`'s `{ ...context, prisma: tx }`), so the snapshot is captured BEFORE the persist and inside the transaction. ✅
- `recoverCreatedRows` (L131) is a true set-difference against `preExistingIds`; created rows are paired to create-payload entries by position via `CreatedRowRecovery.rowAt` (L168), memoised per parent result and per field. Multiple to-many creates each get a distinct row; a create alongside pre-existing rows can never recover a pre-existing row.
- I empirically confirmed the new tests genuinely fail against the old logic: forcing `rowAt` back to `fresh[last]` fails the multi-create test (`new Set(ids).size` = 1 ≠ 2); forcing `preExistingIds` empty fails the pre-existing-rows test (recovers `pre-1`). Both reproduce the original bugs.

**#4 no fabricated `{}` item** — Resolved. The create after-task (L343-360) returns early (skips the after-hook) when `rowAt` yields `undefined`, with a clear comment; it never hands an id-less `{}` to a hook. `runNestedAfterTasks` docstring updated to match.

**#3 `context.db` inside hooks is transactional (`index.ts` / `write-pipeline.ts`)** — Resolved.
- Delegate construction extracted to exported `populateDbDelegate` + `buildDbDelegate` (`index.ts` L441/L489); `getContext` now calls `populateDbDelegate` in place — non-tx path behavior is preserved byte-for-byte.
- `bindContextToTransaction` (`write-pipeline.ts` L256) rebuilds `db` AND `prisma` against `tx`, reusing `session`/`storage`/`plugins`/`_isSudo` and the shared `_resolveOutputCounter` reference. Plugin runtimes are not re-run; sudo and access flow through unchanged.

### Reentrant-transaction question (the flagged risk) — OK, with a test-fidelity caveat

A hook's `context.db` write goes through `runWritePipeline` → `runInTransaction(tx, …)`, which guards on `typeof client.$transaction === 'function'` (`write-pipeline.ts` L154). Real Prisma's interactive `TransactionClient` does NOT expose `$transaction`, so the guard falls through to `fn(prisma)` and the hook's write runs directly on `tx` — no nested/reentrant `$transaction`, no deadlock, no double-wrap. Correct by construction.

Caveat (non-blocking): the rollback test's mock `client.$transaction` passes the FULL client (which still has `$transaction`) as `tx`, so the test exercises the mock's nested-transaction path rather than the real-Prisma flatten path. The test still proves atomicity/rollback (the `audit` `context.db` write is rolled back), but it does not exercise the real-Prisma reentrancy-avoidance branch. Suggest a follow-up test with a tx mock that OMITS `$transaction` to pin the real contract.

### Regression / safety
- `excludeIds` fully removed; connect (no creates) and update (known-id recovery) unaffected. #578 connect/connectOrCreate read-access + reachability untouched; `create` and `connectOrCreate` correctly share one `preExistingIds` snapshot per field.
- Nested delete pipeline, MAX_DEPTH (5), Field Visibility, and the rollback guarantee unchanged by the refactor.
- `id`-as-PK assumption (`r.id`, `where: { id }`) holds for stack-generated schemas (id is a system field); pre-existing constraint, not new.
- to-one creates handled via `includedRows` normalization (single → `[obj]`).
- The intentional `index.ts` ↔ `write-pipeline.ts` cycle is runtime-only (invoked at write time, not module-eval); `tsc` build passes.
- Only one new `as any` (dynamic Prisma model access in `capturePreExistingIds`), internal, eslint-disabled with justification, mirroring existing `getModel`. No `any`/unsafe casts on public surfaces — `populateDbDelegate`/`buildDbDelegate` are properly generic over `TPrisma`.

### Verification run
- `vitest run` core: **641 passed** (incl. 13 in `nested-write-hooks.test.ts`, 4 new).
- `tsc` build: clean. `pnpm lint`: 0 errors (changed files clean; warnings all pre-existing/unrelated).
- Changeset present and accurate (`minor`, documents the `context.db`→tx behavior change).

**Verdict: APPROVE.** Blockers resolved, no regressions. Only a non-blocking suggestion to add a tx-mock-without-`$transaction` test to lock in the real-Prisma reentrancy contract.

---

Issue #579, opened 2026-06-27
SF-14 residual: generated .opensaas client is bundler-only — not loadable by plain-Node consumers of a shared live module (blocks better-auth adapter swap)


borisno2:
Follow-up to #548 / #553 (which made the generated `.opensaas` bundle loadable by a BUNDLER via explicit `.ts` import extensions + static-import recipe). That fix works for Next/webpack-bundled consumers (e.g. an App-Router admin server component). But a residual remains for **live modules consumed in BOTH a bundled and a plain-Node runtime**.

## Symptom
Pointing `better-auth`'s Prisma adapter at the stack-generated client (to drop a dependency on the old ORM's client during a Keystone→stack migration) routes through a runtime `await import('<app>/.opensaas/context.ts')`. In a plain-Node runtime that import fails:
```
SyntaxError: Cannot use import statement outside a module
Warning: Failed to load the ES module: .opensaas/context.ts (set "type":"module" or use .mjs)
```
The auth module (`auth-server.ts`) is imported by:
- the Next server (bundled — OK), AND
- plain-Node consumers: Playwright e2e helpers that create users via better-auth, and (per the warnings) some server runtime paths.
The bundler-only `.ts` artifact can't be loaded in the plain-Node consumers, so the swap regresses every flow that creates a user through better-auth.

## Why it matters
There is no sanctioned **Node-loadable** form of the generated client for a module that must run outside the bundle. The previous ORM's generated client was plain-Node-loadable (statically transpiled), so it worked in every runtime; the stack `.opensaas` client does not. This blocks removing the old ORM entirely (the live auth path still needs a Node-loadable client).

## Ask
A sanctioned Node-loadable build of the generated client/context — e.g. a generator option to emit a compiled-JS (`.cjs`/`.mjs`) `.opensaas/**` alongside the `.ts`, OR a documented `package.json` `exports`/`type:module` setup so `import('.opensaas/context')` resolves in plain Node — so a live module (auth) can use the stack client in ALL its runtimes (Next bundle, plain Node, build-time). Until then, hosts must keep the old ORM's client for the auth path.

## Context
Production music-school app, Keystone→stack migration. This is the exact residual that made us DEFER the better-auth live-adapter swap. Happy to provide the failing e2e trace. Relates to #548/#553.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief (ready-for-human)

**Category:** enhancement
[...]

PR #594, opened 2026-06-27
feat(cli,core): opt-in Node build of the generated bundle (#579)

borisno2:
Implements #579. Part of #580.

Carries **ADR-0010** (`docs/adr/0010-node-build-makes-the-bundle-loadable-without-a-bundler.md`) and the **"Node build"** term in `CONTEXT.md` (both authored on the base branch — untouched here).

## Problem (#579)
The default `.opensaas/` bundle ships as TypeScript with explicit `.ts` import extensions so the host's bundler traces and transpiles it. That form cannot execute under plain Node (`Cannot use import statement outside a module`; `.ts` specifiers don't resolve). A live module that must run in BOTH a bundled and a bundler-less runtime — better-auth's Prisma adapter, imported by the Next server AND by a Playwright e2e helper / build-time script — therefore had no Node-loadable entry to point at, which blocked the better-auth live-adapter swap.

## What this adds (architecture per ADR-0010 — implemented exactly)
An **opt-in, additive Node build**: a compiled, plain-Node-loadable form of the bundle emitted *alongside* the default `.ts` form.

### 1. Config surface (`@opensaas/stack-core`)
- New `output?: { buildTarget?: 'node' }` on `OpenSaasConfig['output']` (`OutputConfig` in `packages/core/src/config/types.ts`). Strongly-typed string-literal union (no `any`/casts), documented. Default (omitted) = today's behaviour.

### 2. Emit step (`@opensaas/stack-cli`)
- New `packages/cli/src/generator/node-build.ts`. In `generate.ts`, after the `.ts` bundle is written **and** after `prisma generate` produced `.opensaas/prisma-client/**`, when `config.output?.buildTarget === 'node'` it compiles the bundle to `.opensaas/dist/` via the **TypeScript compiler API**:
  - `rewriteRelativeImportExtensions: true` (rewrites the bundle's `.ts`-extension imports to runnable `.js`), `allowImportingTsExtensions`, `declaration` (`.d.ts`), `module: esnext` + `moduleResolution: bundler` (ESM out), `target: es2022`, `skipLibCheck`, `noEmitOnError: false`.
  - Writes `.opensaas/dist/package.json` = `{"type":"module"}`.
  - **Compiled entry is `.opensaas/dist/context.js`**; the Prisma client subtree compiles to `.opensaas/dist/prisma-client/**`; the project `opensaas.config` is compiled in as a sibling so the build is self-contained under plain Node.
  - To keep the entry at `.opensaas/dist/context.js` while the source `context.ts` imports the config one level *above* `.opensaas/`, the emit stages the bundle + a copy of the config into a hidden `.opensaas/.node-build/`, rewrites the `../opensaas.config.ts` imports to the sibling `./opensaas.config.ts`, compiles to `dist/`, then removes the staging dir.
- The default `.ts` form emission is completely unchanged when the option is absent.

### 3. Real plain-Node verification consumer (the DoD)
- `examples/starter-auth` opts in via `output: { buildTarget: 'node' }`.
- `examples/starter-auth/scripts/node-build-create-user.mjs` is run by the **`node` binary** (a `.mjs`, NO bundler, NO tsx): it `import`s the compiled `.opensaas/dist/context.js`, constructs better-auth via `createAuth(config, rawOpensaasContext)`, and creates a user — mirroring the failing #579 trace.
- `e2e/starter-auth/04-node-build.spec.ts` asserts the `dist/` shape and **spawns `node` on that script**, asserting it prints `NODE_BUILD_OK` and exits 0. It reuses the existing `pnpm generate && pnpm db:push` global setup, so the compiled bundle is produced in CI.

## Tests
- `packages/cli/src/generator/node-build.test.ts` (vitest): option default-off (no `dist/`), option-on emits `dist/context.js` + `.d.ts` + `package.json {"type":"module"}`, and emitted imports carry `.js` (not `.ts`) extensions and self-contain the config import inside `dist/`.
- Changeset: MINOR for `@opensaas/stack-cli` and `@opensaas/stack-core` with a usage example.

https://claude.ai/code/session_01RyPQQD4zvAiBFJJyKYhvCH

---
_Generated by [Claude Code](https://claude.ai/code/session_01RyPQQD4zvAiBFJJyKYhvCH)_


borisno2 (commented):
## Review — verdict: REQUEST CHANGES

(Submitted as a COMMENT review because the PR was authored by this integration identity; the explicit verdict is **REQUEST CHANGES** for the one blocking issue below.)

The PR is well-scoped and the design matches ADR-0010 closely. The default-off gating is clean, the staging/rewrite/compile mechanics are correct, the entry/`package.json` shape is right, and the test coverage (unit + e2e spawning the real `node` binary on a `.mjs`) genuinely verifies the bundler-less path. One blocking dependency issue, plus a couple of lower-severity notes.

### 1. (Blocking) `typescript` is only a devDependency but `node-build.ts` imports it at runtime
`packages/cli/src/generator/node-build.ts:3` does `import ts from 'typescript'`, and this is the **only** runtime `typescript` import in the CLI's shipped `src/`. In `packages/cli/package.json`, `typescript` sits in `devDependencies` (`^5.9.3`), and none of the runtime `dependencies` (chalk, chokidar, commander, fs-extra, glob, jiti, ora, prompts, zod) guarantee it is installed.

Failure: an end user who installs `@opensaas/stack-cli` from npm and runs `opensaas generate` with `output: { buildTarget: 'node' }` can hit `Cannot find package 'typescript'` — the Node build fails for exactly the bundler-less consumers it exists to serve. The monorepo (and CI/e2e) hides this because TypeScript is hoisted into the workspace root. Move `typescript` to `dependencies` (or declare it as a `peerDependency` with a clear install note).

### 2. (Low) Config-import rewrite is hardcoded to `../opensaas.config.ts` and will silently miss a customised path
`stageBundleFile` rewrites the literal `'../opensaas.config.ts'` → `'./opensaas.config.ts'` (`node-build.ts`, `CONFIG_IMPORT_FROM/TO`). But the generator emits the config specifier from `configImport ?? '../opensaas.config'` (`packages/cli/src/generator/context.ts:21`, `prisma-extensions.ts:20`), i.e. it is configurable. If a project sets a custom `opensaasDir`/`opensaasPath` that changes the relative depth (e.g. `../../opensaas.config.ts`), the rewrite no-ops, the staged file keeps reaching outside `dist/`, and the compiled entry's config import won't resolve under plain Node — with no error surfaced (`noEmitOnError: false`). Fine for the default and starter-auth layout shipped here, but worth either deriving the from-specifier from the resolved paths or asserting the default layout. At minimum, document the constraint.

### 3. (Info) The e2e/starter-auth compile will emit non-fatal type diagnostics by design
`examples/starter-auth/opensaas.config.ts:5` imports `import type { Lists } from '@/.opensaas/lists'`, which uses the host `@/*` alias that the Node-build tsc options don't define. Because it's `import type` it is elided from JS (so `opensaas.config.js` is clean) and `noEmitOnError: false` lets emit proceed — this is handled and explicitly anticipated in the `node-build.ts` header comment. Calling it out only so the yellow "completed with type diagnostic(s)" warning in `generate.ts` isn't mistaken for a regression.

### What I verified as correct
- **Default-off:** the entire Node build is gated behind `config.output?.buildTarget === 'node'` in `generate.ts`; with `output` absent, no `dist/` is emitted and behaviour is byte-for-byte unchanged.
- **Staging cleanup on failure:** `buildNodeBundle` wraps the compile in `try { … } finally { fs.rmSync(stagingDir, …) }`, so the hidden `.opensaas/.node-build/` is removed even if compile throws.
- **Real ESM that runs in plain Node:** `rewriteRelativeImportExtensions` (TS 5.7+, satisfied by `^5.9.3`) rewrites `.ts` specifiers to `.js`; `module: ESNext`, `dist/package.json = {"type":"module"}`, `.d.ts` emitted; `rootDir`/`outDir` map `context.ts → dist/context.js` and `prisma-client/client.ts → dist/prisma-client/client.js` cleanly. Unit + e2e assert `.js` (not `.ts`) specifiers and `not.toContain('.ts.js')`.
- **Verification consumer truly bypasses bundler+tsx:** `04-node-build.spec.ts` spawns `process.execPath` (the `node` binary) on `scripts/node-build-create-user.mjs`, which `import`s the compiled `.opensaas/dist/context.js` and creates a user via `createAuth(...).api.signUpEmail`. No tsx/Next in the load path. `setupDatabase` runs `pnpm generate && pnpm db:push`, so `dist/` exists before the spec.
- **Type safety:** `buildTarget?: 'node'` is a clean string-literal union (no `any`/casts); `node-build.ts` uses concrete `ts.*` types throughout.
- **Changeset:** present, MINOR for both `@opensaas/stack-cli` and `@opensaas/stack-core`, with a usage example.
- **Carried docs:** ADR-0010 and the CONTEXT.md "Node build" term are additive and consistent with the implemented code.

Address #1 (move `typescript` to a runtime dependency) and this is good to merge.

---

borisno2:
Both review findings addressed in `0f71860`:

**1. `typescript` is now a runtime dependency of the CLI.** `node-build.ts` does `import ts from 'typescript'` at runtime, but it was only in `devDependencies` — a user installing `@opensaas/stack-cli` from npm and running `opensaas generate` with `output: { buildTarget: 'node' }` would hit `Cannot find package 'typescript'` (the monorepo hides this via root hoisting). Moved it to `dependencies` at the same `^5.9.3` range used everywhere else so versions stay aligned; ran `pnpm install` (lockfile updated) and `pnpm manypkg fix`.

**2. Hardened the config-import rewrite + added a loud guard.** The rewrite was a literal `'../opensaas.config.ts'` → `'./opensaas.config.ts'`, which silently no-ops for a non-default layout (e.g. a deeper `'../../opensaas.config.ts'`) — and with `noEmitOnError: false` the broken Node build would emit with no error surfaced. Now:
- The rewrite matches a relative `opensaas.config` specifier of **any** `../` depth (bounded regex, quote-style preserved) and rewrites it to the sibling `./opensaas.config.ts`.
- After staging, if any staged bundle file still contains a `../`-rooted `opensaas.config` import, it **throws** a clear error (commented as a silent-failure guard, since `noEmitOnError: false`).

Tests: added a real-compile test asserting the deeper `../../opensaas.config.ts` specifier self-contains inside `dist/`, plus direct coverage of the rewrite/guard helpers. Also raised the timeout (60s, block-level) on the real-`tsc`-compile node-build tests so they don't flake on slow CI runners (the test that timed out at 5000ms). `pnpm --filter @opensaas/stack-cli build` + `test` pass (308 tests); `pnpm lint` is clean (0 errors). Reviewer note #3 (the `@/.opensaas/lists` type-only diagnostic) left as-is, by design.

---

Issue #582, opened 2026-06-27
[PRD] Validation & field-type input parity (zod 4.4 + calendarDay)


borisno2:
> *This was generated by AI during triage.*

## PRD — Validation & field-type input parity

**Goal:** Make field input validation and generated input types correct and consistent, so normal partial-update idioms work and field types don't contradict their own validation/storage. Both issues surface in the built-in field builders' `getZodSchema` / type generation.

**Why these are grouped:** both touch the built-in field builders' update-schema and input-type generation; #571 explicitly coordinates with #570's `union → optional` change to avoid conflict.

### Child issues
- **#570** (bug, ready-for-agent) — zod 4.4 rejects omitted keys on update (`z.union([schema, z.undefined()])`) → use key-optionality (`.optional()`) across all field builders.
- **#571** (bug, ready-for-agent) — `calendarDay` type/validation/storage contradiction → string-only (`YYYY-MM-DD`) end-to-end, including read output (behavioral change, flag in changeset).

### Suggested sequencing
- Do **#570** first (it touches every builder's required-on-update branch, including calendarDay's `union`), then **#571** (calendarDay-specific type + read transform) on top.

### Definition of done
- Partial updates that omit required-on-create fields validate under the resolved zod 4.4.
- `calendarDay` is a `YYYY-MM-DD` string in input types, validation, and read output; passing a `Date` is a compile error.

borisno2:
## QA Sign-off — PRD #582 (Validation & field-type input parity: zod 4.4 + calendarDay)

**Verdict: ✅ PASS (with one accepted, tracked deviation)**

Independently verified against the merged code on integration branch `claude/deliver-prd-582-aadkkh` at HEAD `618ab09` (commits `4a0ec24` #570 and `618ab09` #571). Verified against the spec, not the PR descriptions.

### Acceptance criteria

| Criterion | Result | Evidence |
|---|---|---|
| #570 — no `z.union([schema, z.undefined()])` update-shapes remain | ✅ PASS | `grep z.undefined()` in `packages/core/src/fields/index.ts` → **zero matches**. The two remaining `z.union` (L294 decimal `string\|number`, L459 timestamp `date\|iso`) are legitimate value-type unions, not key-optionality. |
[...]

Issue #570, opened 2026-06-11
Update-input validation rejects OMITTED fields under zod 4.4 — union-with-undefined no longer matches missing keys


borisno2:
## Summary

`@opensaas/stack-core@0.24.0` declares `zod ^4.3.6`, which resolves to 4.4.x in host apps. The validation layer's required-on-update shape — `z.union([schema, z.undefined()])` — accepts an explicitly-`undefined` value but, as of zod 4.4, REJECTS a *missing* key. Net effect: any `context.db.<list>.update` whose `data` omits a required-on-create field throws `ValidationError` **before the DB write** (and before any hooks fire), even though partial updates are the normal Prisma idiom.

Concrete casualty in our migration: an admin "term details" action that updates `{ termStatus }` without re-sending `name` — worked on zod 4.3, throws on 4.4. Because the failure happens pre-write, hook-driven side effects are also silently lost for any caller that catches errors loosely.

## Repro

zod 4.4.x resolved in the host; any list with a required text field; `db.<list>.update({ where, data: { <someOtherField>: x } })` → ValidationError on the omitted required field.

## Ask

Use `.optional()` (key-optionality) rather than `z.union([schema, z.undefined()])` for update-shapes — partial updates should validate only the keys present. Until then, constrain the declared range to `<4.4` so resolution can't silently break hosts.

## Context

Found while regression-testing through the real write pipeline (mock-free) during a Keystone→stack migration. Happy to PR the `.optional()` change if useful.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Update-input validation uses `z.union([schema, z.undefined()])` for required-on-create fields, which under zod 4.4 rejects an omitted key — so partial updates that don't re-send a required field throw `ValidationError` before the write. Use key-optionality (`.optional()`) instead.

**Current behavior:**
[...]

PR #596, opened 2026-06-27
fix(core): update validation no longer rejects omitted required fields under zod 4.4

borisno2:
## Summary

Implements #570. Part of #582.

The update-input validation layer generated the required-on-update field shape with `z.union([fieldSchema, z.undefined()])`. Under zod 4.4 (which the declared `zod ^4.3.6` resolves to in host apps) a union-with-`undefined` accepts an explicit `undefined` value but **rejects a missing key**. As a result, any `context.db.<list>.update({ where, data })` whose `data` omitted a required-on-create field threw a `ValidationError` before the DB write and before any hooks fired — breaking the normal partial-update idiom.

### Fix

Replaced the union-with-undefined update shape with key-optionality (`.optional()`) on the present-value schema, so update validation only checks keys that are actually present. The present-value rules (non-empty / length / format) are preserved; the field is **not** made nullable (omitting is allowed, but passing `null` to clear a required field is not newly introduced). The `create` branch and the non-required `.optional().nullable()` branch are unchanged.

Four built-in field builders in `packages/core/src/fields/index.ts` used this pattern and were all converted:

- `text` — `z.union([withMax, z.undefined()])` → `withMax.optional()`
- `calendarDay` — `z.union([dateSchema, z.undefined()])` → `dateSchema.optional()`
- `password` — multi-line `z.union([ z.string().min(1, …), z.undefined() ])` → `z.string().min(1, …).optional()`
- `json` — `z.union([baseSchema, z.undefined()])` → `baseSchema.optional()`

A grep confirms zero remaining `z.undefined()` occurrences in the file.

### Tests

Added a regression `describe` block in `packages/core/src/validation/schema.test.ts` that exercises the issue through the **real** validation path (`validateWithZod` → `generateZodSchema`, no mocks):

- Omitted required text field on update with another field present → passes.
- Required text field present-but-invalid on update (empty string / too short) → still rejected.
- Required text field on **create** still required (unchanged).
- Omitted required `calendarDay` / `json` / `password` on update → passes; invalid/empty present values still rejected.

### Verification

- `pnpm --filter @opensaas/stack-core build` — passes, no type errors.
- `pnpm --filter @opensaas/stack-core test run` — 650 passed (30 files), including the new regression tests.
- `pnpm lint` — 0 errors (4 pre-existing warnings in unrelated files); `pnpm manypkg fix`; `pnpm format` — clean.

A `patch` changeset for `@opensaas/stack-core` is included.

---

### DIFF
--- a/.changeset/quiet-otters-fix.md
+++ b/.changeset/quiet-otters-fix.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix update validation rejecting omitted required fields under zod 4.4 by using key-optionality (`.optional()`) instead of `z.union([schema, z.undefined()])`. Partial updates that omit a required-on-create field now validate; present values still enforce their rules.
--- a/packages/core/src/fields/index.ts
+++ b/packages/core/src/fields/index.ts
@@ -84,7 +84,7 @@ export function text<
           : withMin
 
       if (isRequired && operation === 'update') {
-        return z.union([withMax, z.undefined()])
+        return withMax.optional()
       }
 
       return !isRequired ? withMax.optional().nullable() : withMax
@@ -574,8 +574,8 @@ export function calendarDay<
       if (isRequired && operation === 'create') {
         return dateSchema
       } else if (isRequired && operation === 'update') {
-        // Required in update mode: can be undefined for partial updates
-        return z.union([dateSchema, z.undefined()])
+        // Required in update mode: omitted keys pass; present values must be valid
+        return dateSchema.optional()
       } else {
         return dateSchema.optional().nullable()
       }
@@ -752,13 +752,13 @@ export function password<TTypeInfo extends import('../config/types.js').TypeInfo
             message: `${formatFieldName(fieldName)} is required`,
           })
       } else if (isRequired && operation === 'update') {
-        // Required in update mode: if provided, reject empty strings
-        return z.union([
-          z.string().min(1, {
+        // Required in update mode: omitted keys pass; if provided, reject empty strings
+        return z
+          .string()
+          .min(1, {
             message: `${formatFieldName(fieldName)} is required`,
-          }),
-          z.undefined(),
-        ])
+          })
+          .optional()
       } else {
         // Not required: can be undefined or any string
         return z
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review: APPROVE (verdict stated as COMMENT)

Reviewed the full diff and the enclosing field builders + the validation layer (`generateZodSchema` / `validateWithZod`) at `refs/pull/596/head`. This is a tight, correct fix for #570. **Verdict: APPROVE** — no blocking or non-blocking findings.

### What it does
Replaces the required-on-update Zod shape `z.union([schema, z.undefined()])` with `schema.optional()` across the four built-in field builders that used it (`text`, `calendarDay`, `password`, `json`). Under zod 4.4 the union-with-`undefined` form treats the key as required-but-undefined-accepting, so an omitted key on a partial `update` was wrongly rejected before any DB write or hooks ran. `.optional()` marks the key itself optional, so only present keys are validated.

### Verification against the risk areas
- **Present-value rules preserved on every converted site.**
  - `text` → `withMax.optional()` keeps `.min(minLength)` / `.max(length.max)`.
  - `calendarDay` → `dateSchema.optional()` keeps the `^\d{4}-\d{2}-\d{2}$` regex.
  - `password` → `z.string().min(1, …).optional()` keeps the non-empty rule.
  - `json` → `z.unknown().optional()` (no present-value rules to lose).
- **No new nullability introduced.** None of the four conversions added `.nullable()`. `null` is still rejected by `text`/`calendarDay`/`password` exactly as before. For `json`, `z.unknown()` already accepted `null` under the old union too, so behavior is unchanged. Omitting a required field on update passes; passing `null` to clear it is *not* newly allowed.
- **create branches and `.optional().nullable()` branches unchanged** — confirmed line-by-line; only the `isRequired && operation === 'update'` arms were touched.
- **Tests exercise the real validation layer.** The new `describe` block calls `validateWithZod(...)` → `generateZodSchema(...)` with real field builders, no mocks/stubs below the validation boundary. Coverage includes omitted-required-passes, present-but-empty/too-short-rejected, and create-still-required across text/calendarDay/json/password.
- **Conventions.** No new `any` or type casts (the pre-existing `as any` in `password` hooks is untouched). ESM `.js` import extension preserved in the test import. Changeset `.changeset/quiet-otters-fix.md` present, scoped `@opensaas/stack-core: patch` — correct for a bug fix.
- **Completeness.** Grep of the PR-head `fields/index.ts` shows zero remaining `z.undefined()`. Other field types (`integer`, `decimal`, `select`, `timestamp`, `checkbox`) already use `.optional().nullable()` for update, which handles missing keys correctly, so no further sites need conversion.

### Notes (pre-existing, not introduced by this PR, no action required)
- `json` required-on-**create** returns bare `z.unknown()`, which also accepts a missing key / `undefined` — so a required json field is not actually enforced on create. This predates the PR and is out of scope here; worth a separate follow-up if create-enforcement for required json is desired.

No changes requested.

🤖 Reviewed with Claude Code

---

Issue #571, opened 2026-06-11
calendarDay validation accepts only 'YYYY-MM-DD' strings — but the Prisma @db.Date boundary wants Date; Date inputs are rejected pre-write


borisno2:
## Summary

The `calendarDay()` field's zod validation in `@opensaas/stack-core@0.24.0` accepts only a `YYYY-MM-DD` string (no Date instance, no string→Date transform). But the value ultimately lands in a Prisma `@db.Date` column, where the natural application type is `Date`. A caller passing `new Date(...)` — the obvious move for code that previously talked to Prisma or Keystone — gets a `ValidationError` before the write.

So neither representation round-trips cleanly: pass a `Date` → validation rejects it; pass the documented string → fine for validation, but app code converting at the boundary has to know which layer wants which type, and nothing in the types/docs says so.

## Ask

Accept both: `z.union([z.date(), z.string().regex(YYYY_MM_DD)])` with a normalising transform to whatever the engine layer expects — or at minimum document the wire format for calendarDay writes through `context.db.*` and make the generated input types say `string` so it's a compile-time error to pass a Date.

## Context

Found in a Keystone→stack migration: converted admin actions send `Date` objects (correct for the old Keystone/Prisma boundary) and now fail validation pre-write. Mock-based tests don't catch it because they stub below the validation layer. Related: the zod-4.4 omitted-key regression filed separately.

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** `calendarDay` is internally contradictory — validation accepts only a `YYYY-MM-DD` string, but the generated TypeScript type says `Date`, so a typed caller passes a `Date` and gets a runtime `ValidationError`. Make `calendarDay` a `YYYY-MM-DD` string end-to-end (Keystone-accurate), so the type, validation, and runtime agree.

**Current behavior:**
The `calendarDay` field validates input as a `z.string()` matching `^\d{4}-\d{2}-\d{2}$` (no `z.date()`, no transform), stores as `DateTime @db.Date`, but its `getTypeScriptType()` returns `Date`. Because the create/update input types are generated from that same `getTypeScriptType()`, the generated input types say `Date` — so a caller following the types passes `new Date(...)` and hits a `ValidationError` before the write. The type, the validator, and the storage boundary disagree.

**Desired behavior:**
[...]

PR #598, opened 2026-06-27
fix(core): make calendarDay a YYYY-MM-DD string end-to-end

borisno2:
Implements #571. Part of #582.

## Problem

`calendarDay()` was internally contradictory: its zod validation accepts only a `YYYY-MM-DD` string, it stores as `DateTime @db.Date`, but `getTypeScriptType()` returned `Date`. Since the generated `CreateInput`/`UpdateInput` types come from `getTypeScriptType()`, a typed caller passes `new Date(...)` and hits a runtime `ValidationError` before the write — type, validator, and storage boundary disagreed.

## Change

Make `calendarDay` a `YYYY-MM-DD` **string** end-to-end (Keystone's `CalendarDay` scalar) so type, validation, and runtime agree.

- `getTypeScriptType()` now returns `{ type: 'string', optional: isNullable }` (was `'Date'`). This makes the entity/read type **and** the standalone generated `CreateInput`/`UpdateInput` types `string`. At the `context.db` write path a `Date` is rejected at runtime by validation (a `ValidationError`); the generated db method `data` type still derives from Prisma's `Date | string` input, so a `Date` is not a compile-time error at the call site. Making it genuinely compile-time is a separate cross-cutting generator change, tracked in #599.
- `getZodSchema()` keeps the `YYYY-MM-DD` string validation (string + `^\d{4}-\d{2}-\d{2}$` regex with a clear message). The `.optional()` update branch from #570 is left intact.
- Added a field-level `resolveOutput` hook (merged with any user hooks, same pattern as `password()`) that normalises the stored date-only value to a `YYYY-MM-DD` string on read. It handles the value being a `Date` (Postgres/MySQL `@db.Date`), an already-formatted string (SQLite TEXT fallback), or null/undefined. Formatting is **UTC-safe** — a `Date` is formatted via `toISOString().slice(0, 10)` so it never drifts a day in non-UTC timezones; a string takes its date-only prefix.
- `getPrismaType()` unchanged: `DateTime`, with `@db.Date` for non-sqlite providers.
- Updated the JSDoc to document the `YYYY-MM-DD` string wire format in both directions.

## ⚠️ Behavioral change (reads)

Reading a `calendarDay` now returns a `YYYY-MM-DD` **string** instead of a `Date`. Flagged in the changeset (minor bump) with a usage note. Consumers relying on a `Date` on read should switch to the string form.

## Write-time Date rejection (runtime, not compile-time)

Writes accept only a `YYYY-MM-DD` string. A `Date` (or a malformed string) is rejected at runtime by validation (a `ValidationError`) — it is **not** a compile-time error at the `context.db.<list>.create()/update()` call site, because that method's `data` type derives from Prisma's `Date | string` input for a `@db.Date` column. The standalone generated `CreateInput`/`UpdateInput` types are now `string`. Genuine compile-time enforcement at the db call site is tracked in #599.

## Tests

New `packages/core/src/fields/calendar-day.test.ts` covers:
- `getTypeScriptType()` returns `string` (optional per nullability)
- type-level assertion: the declared type literal is exactly `string` (the standalone input/entity types are generated from it) plus a `@ts-expect-error` guard that the type is never `'Date'`
- `getPrismaType()` stays `DateTime @db.Date` (and omits `@db.Date` on sqlite)
- valid `YYYY-MM-DD` write succeeds; malformed string rejected with a `YYYY-MM-DD` message; a `Date` instance rejected at runtime
- read transform returns a `YYYY-MM-DD` string for `Date` and string inputs, is timezone-safe (late-UTC `Date` does not drift), and passes null/undefined through
- user-provided `resolveOutput` hooks are preserved (merge order)

## Verification

- `pnpm --filter @opensaas/stack-core build` — no type errors
- `pnpm --filter @opensaas/stack-core test run` — 665 pass
- `pnpm --filter @opensaas/stack-cli test run` — 308 pass (type/snapshot generator path unaffected; emitted input type for calendarDay is now `string`)
- `pnpm lint` — 0 errors (only pre-existing warnings), `pnpm manypkg fix`, `pnpm format` clean

---

### DIFF
--- a/.changeset/calm-otters-glide.md
+++ b/.changeset/calm-otters-glide.md
@@ -9,10 +9,10 @@ both directions, so its type, validation, and runtime value finally agree.
 Previously the field validated a `YYYY-MM-DD` string but its TypeScript type was
 `Date`, so a typed caller passing `new Date(...)` hit a runtime `ValidationError`.
 
-- The field type (entity, `CreateInput`, `UpdateInput`) is now `string`, so
-  passing a `Date` is a compile error.
-- Writes still accept only a `YYYY-MM-DD` string (malformed strings are rejected
-  with a clear message).
+- The field/read type and the generated `CreateInput`/`UpdateInput` input types
+  are now `string`.
+- Writes accept only a `YYYY-MM-DD` string; a malformed string or a `Date` is
+  rejected at runtime by validation (a `ValidationError`).
 - Storage is unchanged: `DateTime @db.Date` on Postgres/MySQL, the SQLite TEXT
   fallback as before.
 
--- a/.changeset/calm-otters-glide.md
+++ b/.changeset/calm-otters-glide.md
@@ -1,0 +1,31 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Make `calendarDay` a `YYYY-MM-DD` string end-to-end (Keystone's CalendarDay scalar)
+
+`calendarDay` is now a `YYYY-MM-DD` **string** at the `context.db` boundary in
+both directions, so its type, validation, and runtime value finally agree.
+Previously the field validated a `YYYY-MM-DD` string but its TypeScript type was
+`Date`, so a typed caller passing `new Date(...)` hit a runtime `ValidationError`.
+
+- The field type (entity, `CreateInput`, `UpdateInput`) is now `string`, so
+  passing a `Date` is a compile error.
+- Writes still accept only a `YYYY-MM-DD` string (malformed strings are rejected
+  with a clear message).
+- Storage is unchanged: `DateTime @db.Date` on Postgres/MySQL, the SQLite TEXT
+  fallback as before.
+
+**Behavioral change (reads):** reading a `calendarDay` now returns a
+`YYYY-MM-DD` string instead of a `Date`. A field `resolveOutput` transform
+normalises the value Prisma returns from the `@db.Date` column, using UTC
+components to avoid timezone off-by-one. Consumers that previously relied on a
+`Date` on read should update to the string form:
+
+```typescript
+const event = await context.db.event.findUnique({ where: { id } })
+event?.startDate // => '2025-01-15' (string, not Date)
+
+// Writes: pass YYYY-MM-DD strings, not Date objects
+await context.db.event.create({ data: { startDate: '2025-01-15' } })
+```
--- a/packages/core/src/fields/calendar-day.test.ts
+++ b/packages/core/src/fields/calendar-day.test.ts
@@ -1,0 +1,139 @@
+import { describe, it, expect, expectTypeOf } from 'vitest'
+import { calendarDay } from './index.js'
+import { generateZodSchema, validateWithZod } from '../validation/schema.js'
+import type { FieldConfig } from '../config/types.js'
+
+/**
+ * calendarDay is a YYYY-MM-DD string end-to-end (Keystone's CalendarDay
+ * scalar). Type, validation, and runtime read value must all agree on `string`.
+ * See issue #571.
+ */
+describe('calendarDay field (YYYY-MM-DD string end-to-end)', () => {
+  describe('getTypeScriptType', () => {
+    it('returns string (not Date), driving the entity + input types', () => {
+      const field = calendarDay()
+      expect(field.getTypeScriptType?.()).toEqual({ type: 'string', optional: true })
+    })
+
+    it('is non-optional when required and not nullable', () => {
+      const field = calendarDay({ validation: { isRequired: true } })
+      expect(field.getTypeScriptType?.()).toEqual({ type: 'string', optional: false })
+    })
+
+    it('type-level: the declared type is the literal "string", so a Date is rejected', () => {
+      const field = calendarDay()
+      const tsType = field.getTypeScriptType?.()
+      // The generated entity/CreateInput/UpdateInput types are emitted from this
+      // literal. Asserting it is exactly 'string' is the type-level guarantee
+      // that a Date input is a compile error (the input field type is `string`).
+      expectTypeOf(tsType).toEqualTypeOf<{ type: string; optional: boolean } | undefined>()
+      if (tsType) {
+        expectTypeOf(tsType.type).toEqualTypeOf<string>()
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Review — `fix(core): make calendarDay a YYYY-MM-DD string end-to-end`

**Verdict: REQUEST CHANGES (one substantive correctness/claim issue; everything else is solid).**

The PR resolves the core contradiction well: `getTypeScriptType()` now returns `string`, the `resolveOutput` read transform is correctly wired and timezone-safe, zod validation is unchanged (string-only, `.optional()` update branch from #570 intact), and the changeset is a correct minor bump that flags the read behavioral change. Good test coverage on the field unit. My one real concern is that the PR's headline claim — "passing a `Date` is a **compile error** on `CreateInput`/`UpdateInput`" — does not hold for the actual `context.db.*` write path.

### Findings (most severe first)

**1. `packages/cli/src/generator/types.ts` (`generateCustomDBType` / `generateCreateArgsType` / `generateUpdateArgsType`) — the "Date is a compile error on input" claim is FALSE for the real `db.create()/update()` path.**

`getTypeScriptType()` returning `string` only flips two surfaces:
- `{List}Output` (the entity/read type) — ✅ correct, this is the meaningful half of the fix (read value now matches declared type).
- The standalone `{List}CreateInput` / `{List}UpdateInput` exports — these are emitted from `getTypeScriptType()`, but they are **not** referenced by the generated `db` method signatures.

The actual write methods are typed as:
```
create: <T extends {List}CreateArgs>(args: Prisma.SelectSubset<T, {List}CreateArgs>) => ...
{List}CreateArgs = Omit<Prisma.{List}CreateArgs, 'select' | 'include'> & { ... }
```
So `data` flows from **Prisma's** `Prisma.{List}CreateArgs`. Since the column is generated as `DateTime @db.Date`, Prisma's input type for that field is `Date | string`. Therefore:
```ts
context.db.event.create({ data: { startDate: new Date() } }) // type-checks — NOT a compile error
```
The compile-error guarantee only holds if a caller explicitly annotates with the unused standalone `{List}CreateInput`. The in-code comment in `calendarDay`'s `getTypeScriptType` ("flips the entity type AND the generated CreateInput/UpdateInput types, so passing a Date is a compile error"), the PR description, and the changeset all overstate this. The runtime zod validator is still a correct backstop (a `Date` is rejected at write time with a `ValidationError`), so this is not a data-safety bug — but the stated type-level guarantee is misleading and the test does not actually exercise the `db.create()` call signature (it only asserts the `getTypeScriptType()` literal). Please either (a) soften the claim to "the entity/read type is `string`; writes are enforced by runtime validation" or (b) make the generated write `data` type actually use the OpenSaaS input type so the Date rejection is genuinely compile-time.

**2. `packages/core/src/fields/index.ts` — `formatCalendarDay` silently returns `undefined` for unexpected non-Date/non-string/non-null values (low severity).**

The final fall-through returns `undefined` for any other shape, dropping the field. For a `@db.Date` column this realistically only yields `Date | string | null`, so the trigger is unlikely, but a malformed/unexpected stored value would be silently nulled out on read rather than surfaced. Acceptable as a defensive default; worth a one-line note that this is intentional data-hiding (it already has a comment, so minor).

### Items I verified as CORRECT (not issues)

- **Read pipeline:** field-level `resolveOutput` is invoked on every read (`findUnique`/`findFirst`/`findMany`/singleton `get`) via `filterReadableFields` → `resolveReadableFieldValue` in `packages/core/src/access/field-visibility.ts`; the returned value is substituted back; the call shape is `{ value, operation:'query', fieldName, listKey, item, context }`, so `({ value }) => formatCalendarDay(value)` receives the raw DB value. The hook is `await Promise.resolve(...)`-wrapped, so the synchronous formatter is fine. It runs for `null`/`undefined` values too — and `formatCalendarDay` correctly passes those through.
- **Timezone safety:** `Date` formatted via `toISOString().slice(0,10)` (UTC) — no local-time off-by-one. All three input shapes (Date, string, null/undefined) handled, with tests including a late-UTC (`23:59:59.999Z`) drift case.
- **Validation unchanged:** zod remains string + `^\d{4}-\d{2}-\d{2}$`; required-update branch is `dateSchema.optional()` (#570 preserved, not reverted to a union). A `Date` is rejected at runtime.
- **`as any` cast (the merged-hooks object) is acceptable.** It mirrors the pre-existing `password()` pattern exactly, is confined inside the field builder, and `any` does not leak into the external return type (`CalendarDayField<TTypeInfo>`). The CLAUDE.md rule prohibits exposing `any` as an *external* type — not violated here. Both eslint-disable lines are scoped and justified.
- **Changeset** (`.changeset/calm-otters-glide.md`): correct `@opensaas/stack-core` **minor** bump, explicitly documents the read-now-returns-string behavioral change with a migration snippet.

### Note
This PR targets `claude/deliver-prd-582-aadkkh`, not `main` (expected — part of #582).

Net: the behavioral fix (reads return `YYYY-MM-DD` strings matching the declared type; writes validated string-only) is correct and well-tested. Requesting changes only to correct the overstated compile-time-Date-rejection claim (comment/description/changeset wording, or make it actually true at the `db.create` signature). 🤖 Generated with Claude Code

---

borisno2:
Thanks for the careful review. Addressed the REQUEST CHANGES in `e842c67`:

- **Finding #1 (overstated compile-time `Date` claim):** corrected the wording in the `calendarDay` `getTypeScriptType()` comment, the JSDoc, the changeset, and the PR description. They now accurately state that the entity/read type and the standalone generated `CreateInput`/`UpdateInput` types are `string`, while at the `context.db.*.create()/update()` call site a `Date` is rejected **at runtime** by validation (the method's `data` type derives from Prisma's `Date | string` for a `@db.Date` column) — not at compile time. Genuine compile-time enforcement at the write signature is a cross-cutting generator change, tracked separately in #599.
- **Finding #2 (`formatCalendarDay` fall-through):** acknowledged as intentional defensive behavior; left as-is per your note.

No runtime behavior, zod validation, the `resolveOutput` transform, or `getTypeScriptType()`'s `string` return changed — wording only. Build + 665 core / 308 CLI tests remain green.

---

Issue #588, opened 2026-06-27
Nested connect: apply owning-side relationship field-level access gate + document query deny-by-default for connect


borisno2:
Follow-up from #578 (part of #581). #578 corrected nested `connect` to use **read/query** access on the target list plus a **DB-reachability** check, fixing the false-denial bug. Two smaller items were intentionally deferred during that PR and are captured here so they are not lost.

## 1. Owning-side relationship field-level access gate (deferred)

Keystone required, for a nested `connect`, **read access on the target list AND the relationship field's field-level access on the owning side**. #578 implemented the target read-access + DB-reachability half, but the nested-op handler args (`NestedOpHandlerArgs` in `packages/core/src/context/nested-operations.ts`) do not currently carry the **owning** relationship field's config/access, so the field-level gate could not be applied without new plumbing. There is a `// TODO(#578)` marker in `verifyConnectReachable`.

**Ask:** thread the owning relationship field's `access` (create/update field-level access for the owning list's relationship field) into the nested-connect handler and gate `connect`/`connectOrCreate`-connect by it, in addition to the target read-access + reachability check already in place. Add tests for: owning field-level access denies connect even when target is readable/reachable; sudo bypasses.

## 2. Document the query deny-by-default behaviour change for connect (docs/changeset)

Switching the connect check from target `operation.update` to `operation.query` combines with the access engine's **deny-by-default** for an undefined access rule (`access/engine.ts`). Consequence: a related list with `query: undefined` but a permissive `update` now **denies** nested connect, where it was previously allowed. This is the correct (read-gated) direction and is consistent with how normal reads of such a list already behave (they return empty), but it is a behaviour change worth an explicit note in the access-control docs / migration notes.

**Ask:** add a short note to the access-control docs (and mention in the relevant changeset/release notes) that nested `connect` now requires **read/query** access on the target list, so lists relying on an undefined `query` + permissive `update` must define `query` access to permit connects.

## References
- Parent PRD: #581
- Implemented in: #578

PR #600, opened 2026-06-27
feat(core): gate nested connect by owning relationship field access

borisno2:
Implements #588. Part of #581. Follows #578.

## Summary

#578 made nested `connect` use the **target** list's read/query access plus a DB-reachability check. This PR adds the other half Keystone required: the **owning relationship field's field-level access** gate.

A nested `connect` (and the connect branch of `connectOrCreate`) is now denied unless **both**:
1. the target list's read/query access permits reading the row (existing #578 check, evaluated in the DB), **and**
2. the owning relationship field's create/update field-level access permits the write (new, #588) — evaluated for the enclosing write's operation (`create`/`update`) via the shared `checkFieldAccess` evaluator.

`sudo` bypasses the entire check, as before.

## How the owning field access is threaded

- `NestedOpHandlerArgs` now carries `owningFieldAccess: FieldAccess | undefined` and `enclosingOperation: 'create' | 'update'`.
- `processNestedOperations` already iterates `fieldConfigs` per relationship field, so it reads `fieldConfig.access` (the owning field's field-level access, e.g. `Post.author`) and passes it — together with its own `operation` — into the per-field `args` handed to `processFieldNestedOps`. Those flow through `...args` to the `connect` / `connectOrCreate` handlers unchanged.
- `verifyConnectReachable` (used by `processNestedConnect`) and the connect branch of `processNestedConnectOrCreate` call `checkFieldAccess(owningFieldAccess, enclosingOperation, { session, context })` first; a deny throws the existing connect-denied error. The #578 target read-access + DB-reachability behaviour is unchanged and still runs.

No `// TODO(#578)` breadcrumb existed in `verifyConnectReachable` (already removed on a prior PR), so no cleanup was needed there.

### Note on layering (finding)

The owning relationship field's field-level access is **also** enforced one step earlier by the parent write's `filterWritableFields` gate (#568), which fires before nested-op processing whenever the write names the relationship key. For the standard nested-connect payload that gate short-circuits first (with a `Cannot create/update <field>` error). The connect-site gate added here is the same check applied at the connect site (defense-in-depth and exactly what #588 asked to thread through the handler). The deny tests therefore assert the write is denied fail-loud and never persisted; the allow/sudo tests exercise the connect-site gate's pass-through directly.

## Docs

Added a "Nested `connect` requires read/query access on the target" section to `docs/content/core-concepts/access-control.md`, including the deny-by-default migration note: a related list with an undefined `query` but permissive `update` now denies nested connect and must define a `query` rule to permit it.

## Tests

New describe block in `packages/core/tests/nested-access-and-hooks.test.ts`:
- owning field DENIES connect for enclosing **update** and **create** (target readable/reachable) → write denied, not persisted;
- owning field ALLOWS → connect succeeds for a reachable row;
- **sudo** bypasses the gate;
- `connectOrCreate` connect branch gated identically (deny + allow);
- existing #578 nested-filter reachability tests still pass.

## Verification

- `pnpm lint`: 0 errors (4 pre-existing warnings, none in changed files).
- `pnpm manypkg fix`, `pnpm format`: clean.
- Build: all 11 packages.
- Full unit suites green: core 647, cli 308, auth 133, create-opensaas-app 41 (+2 skipped), rag 356, storage 123, storage-s3 43, storage-vercel 30, ui 190.
- Changeset: `@opensaas/stack-core` minor.

---

### DIFF
--- a/.changeset/silver-foxes-gather.md
+++ b/.changeset/silver-foxes-gather.md
@@ -1,0 +1,24 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Gate nested `connect` by the owning relationship field's field-level access
+
+Nested `connect` (and the connect branch of `connectOrCreate`) is now gated by
+the owning relationship field's create/update field-level access, in addition to
+the target list's read/query access and DB-reachability check. This completes
+the Keystone-parity rule that a connect requires both read access on the target
+AND write access on the owning relationship field. `sudo` bypasses the check.
+
+```typescript
+Post: list({
+  fields: {
+    // A non-sudo caller can only connect an author when this field's
+    // update access permits it (and the target User is readable/reachable).
+    author: relationship({
+      ref: 'User.posts',
+      access: { update: ({ session }) => session?.role === 'editor' },
+    }),
+  },
+})
+```
--- a/docs/content/core-concepts/access-control.md
+++ b/docs/content/core-concepts/access-control.md
@@ -176,6 +176,53 @@ Fields `id`, `createdAt`, `updatedAt` are automatically:
 
 You cannot override access control for system fields.
 
+## Nested `connect` requires read/query access on the target
+
+When a write uses a nested `connect` (or the connect branch of `connectOrCreate`)
+to link an existing related row, the connect is gated by **read/query access on
+the target list**, evaluated against the database. The caller must be able to
+_read_ the row to connect it — a connect references an existing row but does not
+modify its data, so it requires read access on the target, not `update`. The
+owning relationship field's own create/update field-level access (e.g. the
+`access` on `Post.author`) must also permit the write, as with any field.
+
+```typescript
+lists: {
+  Author: list({
+    fields: { name: text() },
+    access: {
+      operation: {
+        // Without a `query` rule, read access is DENY-BY-DEFAULT...
+        update: () => true, // ...even though update is permissive.
+      },
+    },
+  }),
+  Post: list({
+    fields: {
+      title: text(),
+      author: relationship({ ref: 'Author' }),
+    },
+    access: { operation: { query: () => true, update: () => true } },
+  }),
+}
+
+// This nested connect is DENIED, because Author has no `query` rule
+// (deny-by-default), even though Author's `update` is permissive:
+await context.db.post.update({
+  where: { id },
+  data: { author: { connect: { id: authorId } } },
+})
+```
+
+> **Behaviour change / migration note.** Because the access engine is
+> **deny-by-default** for an undefined access rule, a related list that defines a
+> permissive `update` but leaves `query` **undefined** now **denies** nested
+> `connect` (previously such connects were allowed). This is the correct,
+> read-gated direction and is consistent with how normal reads of that list
+> already behave (they return empty). If you rely on connecting to such a list,
+> **define a `query` rule** that permits the connect (for example a permissive
+> `query: () => true` or a scoped filter). `sudo` bypasses the check entirely.
+
 ## Access Control Execution Order
 
 For **write operations** (create/update):
--- a/docs/content/core-concepts/access-control.md
+++ b/docs/content/core-concepts/access-control.md
@@ -176,15 +176,22 @@ Fields `id`, `createdAt`, `updatedAt` are automatically:
 
 You cannot override access control for system fields.
 
-## Nested `connect` requires read/query access on the target
+## Nested `connect` is gated by the owning relationship field's access
 
 When a write uses a nested `connect` (or the connect branch of `connectOrCreate`)
-to link an existing related row, the connect is gated by **read/query access on
-the target list**, evaluated against the database. The caller must be able to
-_read_ the row to connect it — a connect references an existing row but does not
-modify its data, so it requires read access on the target, not `update`. The
-owning relationship field's own create/update field-level access (e.g. the
-`access` on `Post.author`) must also permit the write, as with any field.
+to link an existing related row, the connect is gated by the **owning relationship
+field's create/update field-level access** (e.g. the `access` on `Post.author`),
+evaluated for the enclosing write's operation. If that field's field-level access
+denies the write, the connect is denied — exactly as for any other field on the
+write. This gate receives the same `item` (the row being updated) and `inputData`
+(the write payload) the parent write's field-access check uses, so a rule that
+depends on either evaluates identically wherever the field is enforced.
+
+For context, the connect is **also** gated by **read/query access on the target
+list** (evaluated against the database): the caller must be able to _read_ the row
+to connect it, because a connect references an existing row but does not modify its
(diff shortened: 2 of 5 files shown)


borisno2 (commented):
## Review verdict: APPROVE (posted as COMMENT — cannot self-approve via API). Minor non-blocking notes only.

Reviewed strictly from the PR diff + the code it touches. The connect-site owning-field gate is implemented correctly and the redundancy the author flagged is harmless defense-in-depth that matches #588's explicit ask. No correctness bugs; nothing blocking.

### Correctness — verified sound
- **Right field access**: `processNestedOperations` sets `owningFieldAccess = fieldConfig.access` where `fieldConfig` is the relationship field on the *parent* list being written (e.g. `Post.author`), not the target list. Correct.
- **Right operation**: `enclosingOperation: operation` is the enclosing `create`/`update` (typed `'create' | 'update'`), never `'read'`. Correct.
- **Sudo preserved, doubly**: in `processNestedConnect` the `verifyConnectReachable` loop is inside `if (!context._isSudo)`; the `connectOrCreate` connect-branch gate is inside its `if (!context._isSudo)` block; and `checkFieldAccess` itself early-returns `true` under sudo. Correct.
- **#578 preserved & additive**: the new gate is inserted at the *top* of `verifyConnectReachable` / before the existing `queryAccess` block in `processNestedConnectOrCreate`. The target read-access + DB-reachability logic is untouched and still runs as an AND, not a replacement.
- `NestedOpHandlerArgs` field types are sound; the two registry handlers destructure/forward correctly; `await` present throughout; no `any`/unsafe casts beyond the pre-existing localized `// eslint-disable` pattern for dynamic Prisma model access.

### The redundancy claim — confirmed, and acceptable
Traced all three `processNestedOperations` callers: every one is fed post-`filterWritableFields` data. For a payload like `data: { author: { connect: {...} } }`, the `author` key is in `data`, so #568's `filterWritableFields` calls `checkFieldAccess(Post.author.access, operation, …)` in write-pipeline **Phase 5** and throws `Cannot <op> "author": field-level access denied.` *before* nested processing (Phase 5.5). Critically, **deeply-nested** writes are NOT a bypass: `processNestedCreate`/`processNestedUpdate` each call `filterWritableFields` on their sub-payload before recursing, so a nested `{ create: { organisation: { connect } } }` still gates `User.organisation` via #568 first.

So the connect-site gate never adds a denial #568 wouldn't already throw — it is functionally redundant for every payload shape I could find. Per the issue's explicit ask to thread the gate through the handler, this is acceptable defense-in-depth and **not a reason to block**.

### Minor notes (non-blocking)

1. **Narrowed `checkFieldAccess` invocation (latent inconsistency).** The connect-site call is `checkFieldAccess(owningFieldAccess, enclosingOperation, { session, context })` — it omits `item` and `inputData`, which the canonical Phase-5 call (`filterWritableFields`) does pass (`item: originalItem`, `inputData`). For a relationship `access.update`/`create` rule that dereferences `item`/`inputData`, the two evaluations can diverge. Because the connect-site gate only runs *after* #568 already passed, the practical effect is at most a spurious extra denial (never a missed denial). Still, `field-access.ts` documents `checkFieldAccess`'s signature as deliberate ("field access can depend on … `item` … and `inputData`"); if this gate is kept, consider passing the enclosing `originalItem`/`inputData` so it cannot diverge from the canonical gate.

2. **Deny tests don't exercise the new deny branches.** The three DENY tests name the relationship key, so #568's `filterWritableFields` throws `field-level access denied` in Phase 5 first — and `OWNING_FIELD_DENIES = /field-level access denied|Cannot connect/` matches that alternative. The new `Access denied: Cannot connect…` throws in `processNestedConnect` / the `connectOrCreate` connect branch therefore have **zero deny-path coverage**. The in-test NOTE honestly acknowledges #568 short-circuits. The ALLOW tests (assert flow reaches the #578 `findUnique` after the gate) and the sudo test (`owningFieldUpdate` spy never called) *do* validly exercise the new pass-through/bypass. Optional: add a deny test that reaches the connect-site gate directly to cover the new throw.

3. **Doc is accurate but mis-scoped to this PR.** The new access-control section's headline migration note ("permissive `update` + undefined `query` ⇒ nested connect now denied") describes the **#578** target-read change, which already shipped — not #600's owning-field gate. #600's actual change (owning relationship field's create/update access must also permit) appears only as a single prose sentence. The content is correct and complementary, but a reader may attribute the #578 behaviour change to this PR. Consider expanding the owning-field paragraph or re-attributing the migration note.

### Test soundness
`beforeEach` reassigns a fresh `mockPrisma` and runs `vi.clearAllMocks()`, so dynamically-added `account`/`student` mocks don't leak and "not called" assertions are reliable. `mockPrisma.user.findFirst` exists in the factory, so the sudo test's assertion is clean (not a throw-on-undefined).

Net: solid, correct, well-documented in code. The redundancy is harmless and intended; the three notes above are polish, not blockers.

---

borisno2:
Addressed the review finding (commit `0c93edc`).

**Connect-site gate now receives `item`/`inputData` (no divergence from #568).** The connect-site `checkFieldAccess` previously passed only `{ session, context }`, omitting the `item`/`inputData` that the canonical Phase-5 `filterWritableFields` call passes. A relationship field whose field-level access depends on `item` or `inputData` (e.g. `({ item }) => item.status === 'draft'`) could be ALLOWED by Phase 5 yet DENIED at the connect site — a spurious denial of a legitimate connect.

Fix: threaded the enclosing write's `originalItem` (`item`) and `inputData` from `processNestedOperations` (invoked in the write pipeline with the same `originalItem`/`input` Phase 5 uses) through new `enclosingItem`/`enclosingInputData` fields on `NestedOpHandlerArgs` into the connect-site `checkFieldAccess({ session, item, inputData, context })` call, for both `connect` and the `connectOrCreate` connect branch. The two gates now evaluate the identical rule with identical values and cannot diverge. #578's target read-access + DB-reachability check and the `sudo` bypass are unchanged; #569 in-transaction behavior is untouched.

**Test added.** A no-spurious-denial test asserts an `item`/`inputData`-dependent owning-field rule that should ALLOW the connect (enclosing UPDATE) does allow it, and verifies the gate was evaluated with the enclosing `item`/`inputData`. Because #568 (Phase 5) is the first line of defense and evaluates the same rule with the same values, a sole-connect-site DENY isn't separately constructible end-to-end, so a focused unit test pins the shared evaluator's deny/allow branches directly with `item`/`inputData`.

**Docs re-scoped.** `access-control.md` now leads with the owning relationship field's field-level access gate (this change) and keeps the #578 target-read requirement as context, noting both checks must pass and that the gate sees the same `item`/`inputData` as the parent write.

Verification: `pnpm lint` (0 errors, 4 pre-existing warnings), `manypkg fix`, `format` clean; core + cli built. Full suites green — core 650, cli 308, auth 133, create-opensaas-app 41 (+2 skipped), rag 178, storage 123, storage-s3 43, storage-vercel 30, ui 190.

---

PR #601, opened 2026-06-27
feat(core): beforeTransaction / afterTransaction transaction-boundary hooks

borisno2:
Implements #590. Builds on #569 / ADR-0010. Part of #581.

## What

Adds list- and field-level **`beforeTransaction`** / **`afterTransaction`** hooks that run **outside** the write's `$transaction` — the compensation bracket the ADR-0010 "deferred to a follow-up" section described. Unlike the in-transaction `beforeOperation`/`afterOperation` (which roll back with the transaction), these are for non-transactional side effects (external API calls) that must not hold a DB transaction open and cannot be rolled back.

- **`beforeTransaction`** runs before the transaction opens.
- **`afterTransaction`** runs after it settles and **always runs** (when its paired `beforeTransaction` ran), receiving the outcome:
  - `status: 'committed' | 'rolled-back'`, plus `error` on rollback.
  - On commit: the persisted `item` (and `originalItem` for update/delete), like `afterOperation`.
  - On rollback: **no `item`** — `inputData`/`originalItem` + `error`, so it can compensate.
- **Per list involved** in the write (top-level list + each nested create/update/delete list).
- **Symmetric-bracket "always-run" rule:** `afterTransaction[L]` runs iff `beforeTransaction[L]` ran. A throwing `beforeTransaction` aborts the write (the transaction never opens) and compensates only the already-run lists. A throwing `afterTransaction` does not stop the remaining compensators; errors are surfaced afterward.
- **Sudo** is unaffected (these always run; sudo only bypasses access).

## How

- **Enumerating involved lists before the transaction:** a new `enumerateInvolvedLists` walks the input tree purely (no DB reads), reusing `getRelatedListConfig` and the same nested-op kinds (`create`/`update`/`delete`/`connectOrCreate`) that #569 processes, deduped by `(listKey, operation)`. The top-level list is always first.
- **Symmetric bracket:** `runWithTransactionBoundary` tracks which lists' `beforeTransaction` began (a list is marked "ran" the moment its `beforeTransaction` starts, so a throwing list is still compensated), runs all `beforeTransaction`, opens the existing #569 transaction, then runs `afterTransaction` only for the ran-lists with the settled outcome, collecting (not throwing) errors so all compensators fire.
- **Single target read:** the top-level target + access is resolved once **before** the transaction (gating silent-failures so a denied write fires neither hook, and supplying `beforeTransaction`'s `item`); that resolution is reused inside the transaction via a new internal `preResolvedTarget`, so the target is still read exactly once and **#569 in-transaction before/afterOperation semantics are unchanged.**

`afterTransaction` arg shape (discriminated on `status`), list-level example:
```
{ listKey, operation: 'create', status: 'committed', inputData, item, context }
{ listKey, operation: 'create', status: 'rolled-back', inputData, error, context }
{ listKey, operation: 'update', status: 'committed', inputData, originalItem, item, context }
{ listKey, operation: 'update', status: 'rolled-back', inputData, originalItem, error, context }
{ listKey, operation: 'delete', status: 'committed', originalItem, context }
{ listKey, operation: 'delete', status: 'rolled-back', originalItem, error, context }
```
(field-level mirrors this with an added `fieldKey`.)

## Not in scope
- No change to #569 in-transaction `beforeOperation`/`afterOperation` semantics.
- No explicit `rollback()` helper.

## Tests
New `packages/core/tests/transaction-boundary-hooks.test.ts`: commit path fires `afterTransaction` with the item; rollback path (thrown in-tx `afterOperation`) fires `rolled-back` + error + no item; thrown `beforeTransaction` aborts and compensates only already-run lists; per-list firing across a nested write; an `afterTransaction` throw doesn't block the others; field-level variants; sudo path; access-denied fires neither hook.

## Verification
- `pnpm lint` (0 errors), `pnpm manypkg fix`, `pnpm format`, full `pnpm build` — all green.
- `@opensaas/stack-core` 650 tests, `@opensaas/stack-cli` 308, plus auth/ui/rag/tiptap/storage all green.
- Changeset added (`@opensaas/stack-core`, minor) with usage example. Docs: new in-transaction vs transaction-boundary section in `docs/content/core-concepts/hooks.md`; ADR-0010 follow-up note updated to "shipped".

---


borisno2 (commented):
## Review: transaction-boundary hooks (#590) — REQUEST CHANGES

(Posting as a COMMENT because GitHub won't let me formally request changes on my own PR, but the verdict is **REQUEST CHANGES** — item 1 below is a must-fix.)

Solid, well-documented feature. The symmetric bracket is implemented correctly: `ran` is appended the moment a `beforeTransaction` *begins* (transaction-boundary.ts:825), so a hook that throws mid-action is still compensated; a `beforeTransaction` throw aborts before the tx opens and compensates only ran lists (836-843); `afterTransaction` runs for every ran list on both commit and rollback, errors are collected (not thrown) so remaining compensators fire, and the tx's own error takes precedence over `AfterTransactionError` (859-867). Enumeration correctly excludes `connect`/`disconnect`/`set`/`updateMany`/`deleteMany` (only create/update/delete/connectOrCreate count), depth/recursion is correct with no off-by-one, dedupe is by (listKey, operation), top-level first. Access-denied gating (no boundary hooks for a denied top-level write; nested denial rolls back inside the tx → `rolled-back` fires) is the right and consistent semantics. #569 in-transaction semantics are untouched and the boundary hooks are additive. Tests cover the important paths.

### 1. (must-fix) Type-vs-runtime unsoundness: nested committed `afterTransaction` gets the wrong list's `item`
`packages/core/src/context/transaction-boundary.ts` (~696 / ~702-712, and the field variant in `hooks/index.ts` `executeFieldAfterTransactionHooks` committed branch)

On a committed write, `runAfterTransactionForList` is called for **every** involved list with the single top-level `outcome.item`. For a *nested* list (e.g. a `User` created via `author: { create }` under a top-level `Post.update`), the committed args are built as `{ operation:'create', status:'committed', item: outcome.item }` — i.e. the persisted **Post** row, not the nested User row. But `AfterTransactionHookArgs` declares committed `item: TOutput` (the list's *own* item type). So a nested list's `afterTransaction({ item })` is typed as that list's row while at runtime receiving a different list's row, with **no type error** — a hook reading `args.item.<ownField>` silently gets wrong/undefined data. The inline comment honestly acknowledges "we cannot recover the persisted row outside the transaction, so the committed item we surface is the top-level persisted row" — but the type contract promises otherwise. Please either drop/optional-ize `item` for nested involvements (so the type matches what's delivered), add a discriminator distinguishing top-level from nested, or actually recover the nested row. The honest type is better than a `TOutput` that lies. (`update`'s `originalItem ?? outcome.item` has the same issue.)

### 2. Inaccurate comment + read moved off the transaction client
`packages/core/src/context/write-pipeline.ts:240-242` (comment) and the gate at ~235 vs reuse at ~343

The pre-transaction gate calls `strategy.resolveTarget(getModel(prisma, listName))` against the **non-transactional** request-time client, then passes `preResolvedTarget: gate`, so the in-tx `resolution = args.preResolvedTarget ?? (await strategy.resolveTarget(model))` **never re-resolves**. The comment "The authoritative in-transaction resolveTarget still runs again inside the transaction (#569 semantics unchanged)" is therefore inaccurate — the in-tx resolve is *skipped*, and the target row + singleton `count()` now read outside the tx against the base client. The "read exactly once" / #569 call-count claim is correct (count stays at 1), but the read both moved out of the transaction and switched clients. Isolation impact is low in practice (short interactive tx; singleton's real guard is the fixed `id:1` PK, not the count), but please correct the comment so it matches the code, and note the read now uses the request-time client.

### 3. (doc) `connectOrCreate` is enumerated as an unconditional create involvement
`transaction-boundary.ts` `NESTED_OP_OPERATIONS`

Defensible (over-enumerate is the safer failure mode for a compensator, and pure `connect` is correctly excluded), but a `connectOrCreate` that resolves to *connect* fires the list's bracket with no write. For charge/reserve side effects this can double-act. Worth a one-line note in the hook contract that a `create` involvement is best-effort *intent* and compensators should be idempotent.

### 4. (doc) "per-list" framing is really per-(list, operation), and only the first nested record's `inputData` is surfaced
The bracket fires once per distinct (list, operation) — so a list reached under two operations in one write (e.g. self-relation) fires its `before`/`afterTransaction` twice; and for multiple nested records of the same (list, op), only `entries[0]`'s payload is surfaced as `inputData` (the others are dropped). Both are reasonable per-list-granularity choices, but the repeated "per-LIST compensation bracket" prose is misleading. Please document that it's per-(list, operation), that handlers should use the `operation` discriminant if they want once-per-list semantics, and that per-record nested compensation must use the in-transaction `afterOperation` (which surfaces only `entries[0]` here).

None of these are crashes or data-corruption; #1 is the only one I'd consider a must-fix (unsound public type). Items 2-4 are comment/doc accuracy. Happy to re-review promptly.

---

borisno2:
Addressed the review findings in `7173f9f`.

## MUST-FIX — `afterTransaction.item` was unsound for nested lists
The committed `afterTransaction` handed the **top-level** persisted row as `item`/`originalItem` to **nested** lists' hooks, even though the type declared the nested list's own item type. A hook reading `args.item.<ownField>` silently got the wrong record with no type error.

**Type made honest (optional, not split):** on the committed `afterTransaction` args (both list- and field-level), `item`/`originalItem` are now `TOutput | undefined`. They are guaranteed only for the **top-level** record; for nested lists they are `undefined`. The discriminated `status` shape is unchanged, and rolled-back still carries no `item`. JSDoc on both arg types and both hook declarations now states this.

**Runtime made honest:** added an `isTopLevel` flag to `InvolvedList`. In `runAfterTransactionForList` and `executeFieldAfterTransactionHooks`, the persisted/pre-write rows are passed **only** when `isTopLevel`; nested lists now receive `undefined` instead of the top-level row (removed the `originalItem ?? outcome.item` fallback). Per-record nested compensation should use the in-transaction `afterOperation`, which already receives the correct nested item.

## ALSO — inaccurate write-pipeline comment
The ~line 240 comment claimed the target "still runs again inside the transaction." It does not: `preResolvedTarget` makes the in-tx resolve reuse the pre-tx result (single read preserved). Corrected the comment to describe actual behavior and noted the pre-tx read uses the **non-tx** client. (Moving that read onto the tx client isn't safe — it gates the silent-failure short-circuit *before* the tx opens — so only the comment changed.)

## ALSO — doc accuracy (hooks docs)
Added caveats to the transaction-boundary section: (1) `item`/`originalItem` are top-level-only; (2) granularity is per-`(list, operation)` — a list reached under two operations fires twice, and only the first nested record's `inputData` is surfaced; (3) `connectOrCreate` is a best-effort create involvement (resolve-to-connect still fires the bracket with no write), so compensators should be **idempotent**. Updated the changeset wording to match the new `item` contract.

## Verification — full suite, all green
- `pnpm lint` 0 errors, `pnpm manypkg fix`, `pnpm format` clean; core + cli built (core first).
- `@opensaas/stack-core` **650**, `@opensaas/stack-cli` **308**, `auth` 133, `ui` 190, `rag` 178, `storage` 123, `storage-s3` 43, `storage-vercel` 30 — all passing.
- Tests changed: extended the nested-write test in `transaction-boundary-hooks.test.ts` to assert the corrected contract — top-level Post gets the persisted `item` + its `originalItem`, while the nested User's committed `afterTransaction` gets `item: undefined` and compensates off `inputData`. No existing test asserted the old (wrong) nested-`item` behavior, so none needed reverting.

---

PR #605, opened 2026-06-27
fix(core): required json fields reject present null in validation (#604)

### LINKED ISSUE #604: Required json field accepts present `null` in validation but emits a NOT NULL column (late DB error)
## Summary

Surfaced as a non-blocking note during review of #603 (which fixed #597 — required json now enforced on create).

A required `json` field (`json({ validation: { isRequired: true } })`):
- emits a **NOT NULL** Prisma column (`isNullable = !isRequired = false` in `getPrismaType`), but
- its create/update validation now explicitly **accepts a present `null`** as a legitimate JSON value.

So `context.db.<list>.create({ data: { meta: null } })` passes validation, then fails at the database with a NOT NULL constraint violation — a late, less-friendly error than a validation rejection. The update branch already accepts present `null` too, so this tension is pre-existing, not introduced by #603.

## Decision needed

Clarify what "required json" should mean and make validation and storage agree:
- **Option A — required = non-null:** reject present `null` in validation for a required json field (so the error surfaces at the validation layer, matching the NOT NULL column).
- **Option B — required = key present, null allowed:** keep accepting `null` in validation but emit a nullable column for required json (storage allows null), so the accepted value can actually persist.

Whichever is chosen, add a regression test through the real validation layer and confirm the Prisma column nullability matches.

## Context

- `packages/core/src/fields/index.ts` — `json()` `getZodSchema` (create/update branches) and `getPrismaType` (`isNullable`).
- Related: #597 (merged via #603), #570.

borisno2:
Closes #604.

## Problem

A required `json` field emits a `NOT NULL` Prisma column, but its validation accepted a present `null` (a legitimate JSON value). So `context.db.<list>.create({ data: { meta: null } })` passed validation and then failed at the database with a `NOT NULL` constraint violation — a late, unfriendly error.

## Decision

Per maintainer decision on #604: **required json = non-null.** Validation now rejects a present `null` for a required json field so the error surfaces at the validation layer. The Prisma column stays `NOT NULL` (storage unchanged).

## Change (`packages/core/src/fields/index.ts`, `json()` `getZodSchema`)

- **create + required:** `z.unknown().refine(v => v !== undefined && v !== null, { message: '<Field> is required' })` — rejects absent key and present `null`; accepts other present values.
- **update + required:** `z.unknown().refine(v => v !== null, { message: '<Field> is required' }).optional()` — still allows omission (partial-update idiom, #570), rejects present `null`, accepts present non-null.
- **non-required:** unchanged.
- `getPrismaType()` / nullability untouched — required json stays `NOT NULL`.

## Verified through the real validation layer (`validateWithZod`)

- required create `{}` (absent) → rejected; `{ meta: null }` → rejected; `{ meta: {} }`/`[]`/`0`/`""`/`false` → accepted
- required update `{}` (absent) → accepted (#570 preserved); `{ meta: null }` → rejected; `{ meta: { a: 1 } }` → accepted
- non-required create/update `{}` and `{ meta: null }` → accepted

## Tests & changeset

- Regression block added to `packages/core/src/validation/schema.test.ts`; field-level tests in `packages/core/tests/field-types.test.ts` updated.
- `.changeset/quiet-foxes-jump.md` — patch bump for `@opensaas/stack-core`.

Build clean, `@opensaas/stack-core` 678/678 tests pass, lint clean.

---

### DIFF
--- a/.changeset/quiet-foxes-jump.md
+++ b/.changeset/quiet-foxes-jump.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Required json fields now reject a present `null` during validation rather than failing later as a DB NOT NULL violation. Omitted keys on update are still allowed; the Prisma column nullability is unchanged.
--- a/packages/core/src/fields/index.ts
+++ b/packages/core/src/fields/index.ts
@@ -1376,17 +1376,27 @@ export function json<
       const baseSchema = z.unknown()
 
       if (isRequired && operation === 'create') {
-        // Required in create mode: value must be provided (any present JSON
-        // value is accepted, including null). A bare z.unknown() is treated as
-        // optional inside z.object(), so an omitted key would silently pass;
-        // the refinement makes the key genuinely required by rejecting
-        // undefined (which also covers an absent key).
-        return baseSchema.refine((value) => value !== undefined, {
+        // Required in create mode: a value must be provided and it must be
+        // non-null (issue #604 — a required json field means non-null). A bare
+        // z.unknown() is treated as optional inside z.object(), so an omitted
+        // key would silently pass; the refinement makes the key genuinely
+        // required by rejecting undefined (which also covers an absent key) and
+        // rejecting a present null, while still accepting any other present
+        // JSON value (object, array, primitive, including falsy 0/""/false).
+        return baseSchema.refine((value) => value !== undefined && value !== null, {
           message: `${formatFieldName(fieldName)} is required`,
         })
       } else if (isRequired && operation === 'update') {
-        // Required in update mode: omitted keys pass; present values must be valid
-        return baseSchema.optional()
+        // Required in update mode: omitted keys still pass (issue #570 — partial
+        // updates may leave the field untouched), but a present null is rejected
+        // (issue #604 — required json means non-null). The `.refine()` runs
+        // before `.optional()` short-circuits on undefined: absent/undefined
+        // passes, a present null is rejected, and other present values pass.
+        return baseSchema
+          .refine((value) => value !== null, {
+            message: `${formatFieldName(fieldName)} is required`,
+          })
+          .optional()
       } else {
         // Not required: can be undefined or null
         return baseSchema.optional().nullable()
--- a/packages/core/src/validation/schema.test.ts
+++ b/packages/core/src/validation/schema.test.ts
@@ -332,7 +332,8 @@ describe('Zod Schema Generation', () => {
   // A bare `z.unknown()` is treated as optional inside `z.object(...)`, so a
   // required-on-create json field used to pass when the key was omitted. The
   // create branch now refines the schema to reject undefined/absent keys while
-  // still accepting any present JSON value (object, array, primitive, null).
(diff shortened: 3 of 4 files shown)


borisno2 (commented):
## Review — `/review` (medium effort)

**Verdict: APPROVE** (posted as COMMENT since the review identity may match the PR author; treat this as an approving review.)

### What this PR does
Tightens validation of required `json` fields so a present `null` is rejected at the validation layer (issue #604) instead of surfacing later as a Prisma `NOT NULL` constraint violation. Create-required rejects absent + `null`; update-required still allows an omitted key (#570) but rejects a present `null`. Prisma column nullability is unchanged. Validation-only change, plus regression tests and a patch changeset.

### Crux verified empirically (not taken on the description's word)
I reconstructed the exact PR schemas with the real zod (v4, the version vendored in `packages/core`) and ran them inside `z.object({...}).safeParse(...)` — the same path `validateWithZod → generateZodSchema` takes. Results:

**UPDATE required (`z.unknown().refine(v => v !== null).optional()`):**
- absent `{}` → **PASS** ✅ (the #570 partial-update idiom is preserved)
- `{ meta: null }` → **REJECT** ✅
- `{ meta: { a: 1 } }`, `{ meta: 0 }`, `{ meta: '' }`, `{ meta: false }` → **PASS** ✅

Both load-bearing cases are genuinely covered: the `.refine().optional()` ordering works because inside an object an absent key resolves to `undefined`, `.optional()` short-circuits `undefined` before the refine predicate runs, so absent passes; a present `null` reaches and fails the refine. If it were `.optional().refine(...)` the absent case would still pass but the semantics would be fragile — the chosen ordering is correct.

**CREATE required (`z.unknown().refine(v => v !== undefined && v !== null)`):**
- absent `{}` → **REJECT** ✅
- `{ meta: null }` → **REJECT** ✅
- `{ meta: 0 }`, `{ meta: '' }`, `{ meta: false }`, `{ meta: [] }`, `{ meta: {} }` → **PASS** ✅

The predicate is the explicit `!== undefined && !== null`, not a truthiness check, so falsy-but-valid JSON (`0`/`''`/`false`) is correctly accepted.

### Other checks
- **Tests through the real path:** `schema.test.ts` exercises both absent-on-update→accepted *and* null-on-update→rejected via `validateWithZod`, plus the create matrix and the non-required cases. `field-types.test.ts` covers the same at the `getZodSchema` level. Both directions of the crux are tested.
- **Non-required json unchanged:** `z.unknown().optional().nullable()` branch untouched; omission and `null` both still accepted (tested).
- **Column nullability unchanged:** `getPrismaType()` derives `isNullable = db?.isNullable ?? !isRequired`, so required json (`isRequired: true`) → `isNullable = false` → no `?` → stays `NOT NULL`. Storage behaviour is genuinely unchanged; the fix is validation-only.
- **Conventions:** No `any` / casts introduced; no new imports (ESM `.js` extensions unaffected). Changeset `.changeset/quiet-foxes-jump.md` is a **patch** bump for `@opensaas/stack-core` — correct for a bug fix.

### Findings
None. All 8 review angles (correctness ×3, cleanup ×3, altitude, conventions) came back clean after verification. CI status is green.

Recommend merge.

---

PR #603, opened 2026-06-27
fix(core): enforce required json fields on create (#597)

borisno2:
## Summary

Closes #597.

`json()`'s `getZodSchema` returned a bare `z.unknown()` for the required-on-**create** branch. Because `z.unknown()` is treated as optional inside a `z.object(...)`, a `json({ validation: { isRequired: true } })` field was **not** enforced on create — a create that omitted the key passed validation silently.

This PR refines the create-branch schema so a required json key is genuinely required, while still accepting any present JSON value.

## The fix

```ts
// required + create
return baseSchema.refine((value) => value !== undefined, {
  message: `${formatFieldName(fieldName)} is required`,
})
```

### Verified zod mechanism

Empirically tested against the project's zod (`^4.3.6`). In zod v4, wrapping `z.unknown()` with `.refine(...)` produces a schema that is **no longer treated as optional** inside `z.object(...)`, so an **absent key (`{}`)** is genuinely rejected — not just an explicit `undefined`. Confirmed through the real validation layer (`validateWithZod` → `generateZodSchema`):

- Required json, key **absent** (`{}`) → REJECTED with "Meta is required"
- Required json, explicit `undefined` → REJECTED
- Required json with `{ a: 1 }`, `[]`, `null`, `0` → ACCEPTED
- Non-required json omitted on create → ACCEPTED
- Required json **omitted on update** (#570 behavior) → still ACCEPTED (preserved, untouched)

The required-on-**update** branch from #570 (`.optional()`) is left unchanged.

## Tests

- Added a `required json on create (issue #597)` regression block in `packages/core/src/validation/schema.test.ts`, mirroring the #570 block and exercising the real validation path.
- Updated the field-level unit test in `tests/field-types.test.ts` that previously documented the buggy behaviour (`undefined` accepted) to assert the corrected semantics.

All 672 core tests pass; `tsc` build is clean; lint passes (0 errors).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017JiVKJwhQbG4AfTy65auS4

---
_Generated by [Claude Code](https://claude.ai/code/session_017JiVKJwhQbG4AfTy65auS4)_

PR #606, opened 2026-06-27
Enforce field-level scalar narrowing at the write call site (#599)

### LINKED ISSUE #599: Generated db.create()/update() data type uses Prisma input types, so field-level TS narrowing isn't enforced at compile time
## Summary

Surfaced during review of #598 (PRD #582, follow-up to #571).

A field builder's `getTypeScriptType()` flips the `{List}Output` (entity/read) type and the standalone `{List}CreateInput`/`{List}UpdateInput` exports. But the generated `context.db.<list>.create()/update()` method signatures type their `data` from **Prisma's** `Prisma.{List}CreateArgs` / `UpdateArgs`, not from the OpenSaaS input types:

```ts
create: <T extends {List}CreateArgs>(args: Prisma.SelectSubset<T, {List}CreateArgs>) => ...
{List}CreateArgs = Omit<Prisma.{List}CreateArgs, 'select' | 'include'> & { ... }
```

So any field-level TypeScript narrowing that is *stricter* than the Prisma column type is **not** enforced at the call site. Concrete case: `calendarDay` is generated as `DateTime @db.Date`, whose Prisma input type is `Date | string`. After #571 the OpenSaaS `CreateInput` says `string`, but:

```ts
context.db.event.create({ data: { startDate: new Date() } }) // still type-checks — Date NOT a compile error
```

Runtime zod validation rejects the `Date` (so this is not a data-safety bug), but the compile-time guarantee that #571/#582 aimed for does not hold at the real write path.

## Ask

Make the generated `db.create()/update()` `data` type use the OpenSaaS-generated `{List}CreateInput`/`{List}UpdateInput` (which derive from each field's `getTypeScriptType()`) instead of Prisma's input types — so field-level narrowing (e.g. `calendarDay` → `string`) is a compile error to violate, while keeping relationship/nested-write ergonomics intact.

## Risk / why this is broader than one field

This is a **cross-cutting generator change** (`packages/cli/src/generator/types.ts`) affecting the write-path typing of **every list and field**, including relationship nested writes where the OpenSaaS input shape may not match Prisma's expected nested-write types. Needs design before implementation — likely a careful mapping (or an override of only scalar fields' `data` types) rather than a wholesale swap.

## Context

- Reviewer analysis on #598.
- Related: #571 (calendarDay string end-to-end), PRD #582.
- The #598 PR description/changeset were corrected to stop claiming compile-time Date rejection at `db.create/update`; this issue tracks making it genuinely true.

borisno2:
Closes #599

## Problem

After PRD #582, `calendarDay`'s `getTypeScriptType()` returns `string`, so the entity type and the standalone `{List}CreateInput`/`{List}UpdateInput` exports are `string`. But the generated `context.db.<list>.create()/update()` method signatures typed their `data` from **Prisma's** `Prisma.{List}CreateArgs['data']` (where a `@db.Date` column is `Date | string`), so `context.db.event.create({ data: { startDate: new Date() } })` still type-checked. The field-level narrowing was only enforced at runtime (zod), never at the real write call site.

## Approach (B — scalar-only override)

In `packages/cli/src/generator/types.ts`, the `data` member of the call-site `CreateArgs`/`UpdateArgs` (and the `createMany`/`updateMany` variants) now overrides scalar fields with their OpenSaaS `getTypeScriptType()` types while leaving relationship nested-writes on Prisma's shape:

```ts
data: Omit<Prisma.{List}CreateArgs['data'], scalarKeys | backingColumns>
      & { /* each narrowed scalar: its OpenSaaS type, optional per nullability/default */ }
```

For `createMany` the element type wraps `Prisma.{List}CreateInput`; for `updateMany`, `Prisma.{List}UpdateInput`. This flows through `generateCustomDBType` to all four write methods.

### Verified by `tsc` against the real generated Prisma client (blog example)

```
string -> calendarDay   ✅ compiles
Date   -> calendarDay   ❌ error TS2322: Type 'Date' is not assignable to type 'string'.
author.connect / authorId (relationship + unchecked FK)   ✅ compiles
```

## Edge cases

| Field | Handling |
| --- | --- |
| **decimal** | Left on Prisma's input type — still accepts `Decimal | number | string`. A strict override to `Decimal` would wrongly reject valid `number`/`string` writes. It is excluded from both the `Omit` key set and the narrowed members. |
| **json** | Left on Prisma's input type — keeps the `JsonNull`/`DbNull` sentinels (and doesn't regress the #603/#604 json work). Narrowing to `unknown` would drop the sentinels. Excluded like decimal. |
| **multi-column / storage fields** (`getColumnNames`) | Their raw backing columns are added to the `Omit` (so no dangling Prisma keys remain — mirroring `generateGetPayload`), but they are not narrowed to a single scalar. |
| **createMany / updateMany** | Same override applied — createMany over an array element, updateMany over the single partial input. |
| **optionality / nullability** | Each narrowed scalar preserves required/optional and `| null`, from the same source of truth as the standalone input generators. |

## Bonus bug fix

`checkbox({ defaultValue: false })` was generated as a **required** create field because the optionality check used `!field.defaultValue` (truthy for `false`). Fixed to `field.defaultValue === undefined`, in both the call-site override and the standalone `CreateInput` generator. Covered by a test asserting the field is optional on create.

## Tests

New `packages/cli/src/generator/types-write-narrowing.test.ts` compiles the generated write methods (exactly as `generateCustomDBType` wires them, `<T extends Args>(args: Prisma.SelectSubset<T, Args>) => …`) against a faithful `Prisma` stub via `ts.createProgram`, asserting:
- `Date -> calendarDay` errors (create, update, createMany); `string` compiles
- `number`/`string -> decimal` compiles; `json` value compiles
- relationship nested write + unchecked FK compile
- `checkbox({ defaultValue: false })` is optional on create

A mutation check confirmed the Date assertion is load-bearing (removing the `@ts-expect-error` surfaces `Type 'Date' is not assignable to type 'string'.`). Full CLI suite: 310 passed (308 prior + 2 new; 13 snapshots updated for the new `data` shape).

## Known limitation (out of scope)

The generated method generic uses `Prisma.SelectSubset<T, …>`, which weakens excess-property checking. So the guarantee is **"a known scalar with the wrong type is rejected"** (the `calendarDay`-`Date` case), **not** "no unknown keys are allowed".

---

### DIFF
--- a/.changeset/calm-otters-narrow.md
+++ b/.changeset/calm-otters-narrow.md
@@ -1,0 +1,29 @@
+---
+'@opensaas/stack-cli': minor
+---
+
+Enforce field-level scalar narrowing at the write call site, and fix `checkbox({ defaultValue: false })` optionality
+
+The generated `context.db.<list>.create()/update()/createMany()/updateMany()` `data`
+type now narrows scalar fields to their OpenSaaS `getTypeScriptType()` types instead of
+inheriting Prisma's wider input types. Field-level narrowing (e.g. `calendarDay` → `string`)
+is now a genuine compile-time error to violate, not just a runtime validation failure.
+
+```ts
+// calendarDay is a `string` end-to-end:
+await context.db.event.create({ data: { startDate: new Date() } })
+//                                                  ^^^^^^^^^^ Type 'Date' is not assignable to type 'string'.
+await context.db.event.create({ data: { startDate: '2026-01-01' } }) // ✅ compiles
+```
+
+Relationship nested writes (`connect`/`create`/`connectOrCreate`), unchecked foreign keys
+(e.g. `authorId`), and `decimal`/`json` writes are unaffected: `decimal` still accepts
+`Decimal | number | string` and `json` still accepts Prisma's `JsonNull`/`DbNull` sentinels.
+
+Also fixes a latent bug where `checkbox({ defaultValue: false })` (and any field with a
+falsy-but-present default) was generated as a required field on create — it is now correctly
+optional.
+
+Note: this may surface pre-existing type errors in consumer code that passed a `Date` to a
+`calendarDay` field. Such code already failed at runtime; it now fails at compile time. Pass a
+`YYYY-MM-DD` string instead.
--- a/.changeset/swift-foxes-clean.md
+++ b/.changeset/swift-foxes-clean.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+Remove unused getRelatedListName helper from the types generator (dead code, no behavior change)
--- a/examples/composable-dashboard/lib/actions.ts
+++ b/examples/composable-dashboard/lib/actions.ts
@@ -19,7 +19,7 @@ export async function createPost(data: PostCreateInput) {
     const post = await context.db.post.create({
       data: {
         ...postData,
-        status: postData.status || 'DRAFT',
+        status: postData.status || 'draft',
         publishedAt: postData.publishedAt ? new Date(postData.publishedAt) : undefined,
       },
     })
@@ -47,7 +47,7 @@ export async function updatePost(id: string, data: PostUpdateInput) {
       title: data.title,
       slug: data.slug,
       content: data.content,
-      status: data.status || 'DRAFT',
+      status: data.status || 'draft',
       internalNotes: data.internalNotes,
       publishedAt: data.publishedAt ? new Date(data.publishedAt) : undefined,
     },
--- a/examples/rag-openai-chatbot/scripts/seed.ts
+++ b/examples/rag-openai-chatbot/scripts/seed.ts
@@ -1,6 +1,7 @@
 import { getContext } from '@/.opensaas/context'
+import type { KnowledgeBaseCreateInput } from '@/.opensaas/types'
 
-const sampleArticles = [
+const sampleArticles: KnowledgeBaseCreateInput[] = [
   {
     title: 'What is OpenSaas Stack?',
     content: `OpenSaas Stack is a Next.js-based framework for building admin-heavy applications with built-in access control. It uses a config-first approach similar to KeystoneJS but modernized for Next.js App Router and designed to be AI-agent-friendly with automatic security guardrails. The stack includes core packages for configuration, authentication, UI components, and specialized integrations like RAG (Retrieval-Augmented Generation). It's built as a pnpm monorepo with packages for core functionality, CLI tools, admin UI, authentication, and various integrations. OpenSaas Stack emphasizes type safety, automatic code generation, and developer experience.`,
--- a/packages/cli/src/commands/__snapshots__/generate.test.ts.snap
+++ b/packages/cli/src/commands/__snapshots__/generate.test.ts.snap
@@ -292,16 +292,26 @@ export type UserFindFirstArgs = Omit<Prisma.UserFindFirstArgs, 'select'> & {
 
 /**
  * Custom CreateArgs for User with virtual field support
+ * The \`data\` member narrows scalar fields to their OpenSaaS types (#599).
  */
-export type UserCreateArgs = Omit<Prisma.UserCreateArgs, 'select'> & {
+export type UserCreateArgs = Omit<Prisma.UserCreateArgs, 'select' | 'data'> & {
   select?: UserSelect | null
+  data: Omit<Prisma.UserCreateArgs['data'], 'name' | 'email'> & {
+    name: string
+    email: string
+  }
 }
 
 /**
  * Custom UpdateArgs for User with virtual field support
+ * The \`data\` member narrows scalar fields to their OpenSaaS types (#599).
  */
-export type UserUpdateArgs = Omit<Prisma.UserUpdateArgs, 'select'> & {
+export type UserUpdateArgs = Omit<Prisma.UserUpdateArgs, 'select' | 'data'> & {
   select?: UserSelect | null
+  data: Omit<Prisma.UserUpdateArgs['data'], 'name' | 'email'> & {
+    name?: string
+    email?: string
+  }
 }
 
 /**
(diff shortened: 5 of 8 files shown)


borisno2 (commented):
## Review: APPROVE (posted as COMMENT — review identity is the PR author, so a formal Approve is blocked)

**Verdict: APPROVE.** This is a well-scoped, correctly-implemented generator change. I verified the central claims empirically rather than trusting the description (checked out the PR head, ran the new test, and ran two mutation experiments). No blocking issues found.

### What it does
Rewrites the `data` type of the generated call-site `db.<list>.create()/update()/createMany()/updateMany()` so scalar fields are narrowed to OpenSaaS `getTypeScriptType()` types via `Omit<Prisma.*['data'], keys> & { …narrowed scalars }`, while relationship nested writes, unchecked FKs, and `decimal`/`json` stay on Prisma's input shape. Also fixes a latent optionality bug where `checkbox({ defaultValue: false })` was generated as a required create field.

### Verified (empirical)

**(a) The Date→calendarDay compile error is genuinely tested — CONFIRMED.**
Ran `types-write-narrowing.test.ts` against the PR head: passes. Then removed the `@ts-expect-error` on the `day: new Date()` create line — the suite failed with exactly `Type 'Date' is not assignable to type 'string'.`, proving the assertion is load-bearing, not vacuous. `day: '2026-01-01'` (string) compiles; relationship `owner.connect` and unchecked `ownerId` writes compile.

**(b) decimal and json are NOT regressed — CONFIRMED.**
Mutated `shouldNarrowScalarWrite` to stop excluding decimal/json (i.e. forced them to narrow). The protective tests immediately failed: the structural test caught `'price'`/`'meta'` leaking into the Omit key set, and decimal narrowed to `import('decimal.js').Decimal` (which would reject the `12.5`/`'12.50'` writes). The exclusion is real and tested. Restored after.

**Other checks:**
- **Omit correctness:** `getScalarOverrideOmitKeys` omits each narrowed scalar plus multi-column raw backing columns (`getColumnNames`), and does NOT omit relationship keys or the logical multi-column field key — so no dangling Prisma keys and no dropped relationship writes. Mirrors `generateGetPayload`'s existing `multiColumnRawNames` handling. Correct.
- **createMany/updateMany:** got the same treatment (`Array<Omit<…CreateInput, keys> & {…}>` and `Omit<…UpdateInput, keys> & {…}>`); snapshots and the tsc test cover both.
- **Snapshots:** spot-checked the generate.test and types.test snapshots across single-field, multi-field, required-vs-optional, and relationship lists. New `data` shape is correct; no `any` leaked, no relationship field wrongly narrowed. Full `src/generator` + `src/commands` suite passes (216 tests).
- **Checkbox fix:** `!field.defaultValue` → `field.defaultValue === undefined` applied in both the standalone `generateCreateInputType` and the new `buildScalarOverrideMembers`; the tsc test asserts `active` (a `checkbox({ defaultValue: false })`) is optional on create (`db.event.create({ data: { title: 't' } })` compiles). Latent bug fixed in both code paths.
- **Conventions:** `.js` ESM import extensions intact; no `any`/casts in generated output. Changeset `calm-otters-narrow.md` exists and is a correct `minor` bump for `@opensaas/stack-cli` with a clear consumer-facing migration note.

### Non-blocking observations (no action required)
1. **`| null` divergence from the standalone `*CreateInput`.** The override emits `name?: string | null` for a nullable field, whereas the standalone `{List}CreateInput` emits `name?: string` (no `| null`). This is *more* correct (it matches Prisma's nullable-column input), not a bug — but the two generators now disagree on nullability representation. Worth a future consolidation so there's one source of truth.
2. **Known limitation is honestly documented.** The `Prisma.SelectSubset` generic weakens excess-property checking, so the guarantee is "a known scalar with the wrong type is rejected," not "unknown keys are rejected." The PR states this explicitly; fine for this scope.
3. **Stub fidelity.** The test's `SelectSubset`/`*Input` stub is a faithful model rather than the real Prisma client, but the mutation test confirms it rejects `Date` for the right reason. Reasonable trade-off for a fast, hermetic compile test.

Nice work — the mutation-checked tests make the narrowing guarantee trustworthy, and the deliberate decimal/json carve-out is exactly right.

---

borisno2:
## CI fix: example build failures were latent example bugs (root cause B)

The failing `test` job builds every example (`pnpm generate && next build`, which runs a full TypeScript typecheck). The #599 write-narrowing tightened **select** fields to their literal unions at the `create()`/`update()` call site, which surfaced two pre-existing latent type errors in example app code (both files unchanged by this PR — they relied on the old loose Prisma `data` type and compiled on `main`).

**Root cause: B (example app code), not a generator bug.** No generator change was needed.

### Errors found (after a clean per-example regenerate)
1. `examples/composable-dashboard/lib/actions.ts` — passed `'DRAFT'` to a select that only accepts `'draft' | 'published'`:
   ```
   error TS2322: Type '"DRAFT"' is not assignable to type '"draft" | "published" | null | undefined'.
   ```
   This is the one CI named (turbo fails fast, so it stopped here).
2. `examples/rag-openai-chatbot/scripts/seed.ts` — the `sampleArticles` array literal widened `category` to `string`, rejected by the narrowed `category: "ai-ml" | "web-dev" | ...` union (all the actual values were already valid).

### Fixes
- composable-dashboard: `'DRAFT'` → `'draft'` (2 call sites).
- rag-openai-chatbot: annotated the seed array as `KnowledgeBaseCreateInput[]` (imported from `@/.opensaas/types`) so the literal values narrow correctly — no cast, fully type-safe.

### #599 guarantee preserved
Generator (`types.ts`) untouched. The full CLI suite still passes **310/310**, including the write-narrowing tests asserting `Date → calendarDay` errors and `string` compiles.

### Verification
- All **12 examples** now pass a real `next build` (incl. blog — calendarDay + relationships, and both fixed examples).
- `pnpm --filter @opensaas/stack-cli build` + `@opensaas/stack-core build`: clean.
- `pnpm --filter @opensaas/stack-cli test run`: 310 passed.
- `pnpm lint` (0 errors), `pnpm format:check`, `pnpm manypkg check`: all green.

No `packages/*` changes, so no changeset needed.

---

PR #609, opened 2026-06-27
Consolidate create/update input nullability into one shared helper (#608)

### LINKED ISSUE #608: Consolidate nullability representation between write-data override and standalone {List}CreateInput
## Summary

Non-blocking observation from the review of #606 (#599). The generator now has two places that describe a list's create input shape, and they disagree on how a nullable scalar is represented:

- The new call-site write-`data` override (in `db.create/update`) emits `name?: string | null` for a nullable field (matches Prisma's nullable-column input — more correct).
- The standalone `{List}CreateInput` export emits `name?: string` (no `| null`).

Both compile and neither is a bug today, but having two sources of truth for the input shape is a maintenance hazard — a future field change has to be mirrored in both, and they can silently drift further.

## Ask

Consolidate so there is a single source of truth for the generated create/update input member shape (including nullability), used by both the standalone `{List}CreateInput`/`UpdateInput` and the call-site `data` override. Align on the more-correct `| null` representation for nullable scalars (verify it doesn't regress any consumer that imports the standalone input types). Add/extend generator tests so the two stay in sync.

## Context

- `packages/cli/src/generator/types.ts` — `generateCreateInputType` / `generateUpdateInputType` (standalone) vs the new `buildScalarOverrideMembers` / override path.
- Related: #599 (merged via #606).

borisno2:
Closes #608

## Problem

After #599 (merged via #606), the generator described a list's create/update input shape in **two** places that disagreed on how a nullable scalar is represented:

- The call-site write-`data` override (`buildScalarOverrideMembers`) emitted `name?: string | null` — matching Prisma's nullable-column input (more correct).
- The standalone `{List}CreateInput` / `{List}UpdateInput` exports (`generateCreateInputType` / `generateUpdateInputType`) emitted `name?: string` (no `| null`).

Two sources of truth for the same shape is a maintenance hazard — a future field change has to be mirrored in both and they can silently drift.

## Change

Introduced a single shared helper, `renderScalarInputMember(fieldName, fieldConfig, forCreate, indent)`, that is now the **one** source of truth for a scalar input member's shape: name, `?` optionality, TypeScript type (from the field's own `getTypeScriptType()`), and `| null` nullability.

Both call paths render through it:
- `generateCreateInputType` and `generateUpdateInputType` (standalone `{List}CreateInput` / `{List}UpdateInput`), and
- `buildScalarOverrideMembers` (the call-site `data` override used by `create`/`update`/`createMany`/`updateMany`).

They now **agree** on the more-correct `| null` representation for nullable scalars — the standalone generators were brought up to the override's representation; the override was not weakened.

### Preserved behavior

- Required scalars stay required: no `?`, no `| null` (e.g. `title: string`).
- Nullable scalars: `?:` and `| null` consistently in both outputs.
- Create vs update optionality semantics unchanged (update members already optional; create optional when nullable or has a default — the `defaultValue === undefined` check from #599 is preserved).
- `calendarDay` stays `string`; decimal/json keep Prisma's input type; relationship and multi-column fields unchanged.
- Runtime/zod layer untouched — this is purely the generated TypeScript type shape.

## Tests

- Updated the `types.test.ts` snapshots: every diff is exactly a nullable scalar `?: T` → `?: T | null` (verified — no required field, relationship, or `Date`/`number` regressions).
- Added a `create/update input nullability consolidation (#608)` describe block that pins: standalone `CreateInput`/`UpdateInput` and the `data` override emit the **same** `content?: string | null` for a nullable scalar, required `title` stays required (never gains `| null`), and the two stay in lock-step — so they can't drift again.
- The existing #599 compile-time `tsc` write-narrowing test continues to pass.

`pnpm --filter @opensaas/stack-cli test run`: 316 passed.

## Examples build (the #599 lesson)

Because this changes the public generated `CreateInput`/`UpdateInput` types, I locally ran `pnpm generate` + `prisma generate` + `next build` (full app typecheck) for **all 12 examples**: blog, auth-demo, custom-field, json-demo, composable-dashboard, mcp-demo, tiptap-demo, file-upload-demo, starter, starter-auth, rag-ollama-demo, rag-openai-chatbot. **All build and typecheck clean** — the change surfaced no genuine type errors in example app code, so no example fixes were needed.

## Versioning

`minor` for `@opensaas/stack-cli`: the standalone `{List}CreateInput`/`{List}UpdateInput` exports now include `| null` on nullable scalars, which is a meaningful (non-breaking but visible) refinement of the public generated types.

---

### DIFF
--- a/.changeset/clever-foxes-consolidate.md
+++ b/.changeset/clever-foxes-consolidate.md
@@ -1,0 +1,17 @@
+---
+'@opensaas/stack-cli': minor
+---
+
+Consolidate nullability between the standalone `{List}CreateInput`/`{List}UpdateInput` exports and the call-site write-`data` override into a single source of truth (#608).
+
+The generated types previously described a list's create/update input shape in two places that disagreed on how a nullable scalar was represented: the write-`data` override emitted `name?: string | null` (matching Prisma's nullable-column input) while the standalone `{List}CreateInput`/`{List}UpdateInput` emitted `name?: string`. Both paths now render each scalar member through one shared helper, so a nullable scalar is consistently `name?: T | null` in every input representation. Required scalars stay required, and `decimal`/`json`/relationship/multi-column handling is unchanged.
+
+This is a non-breaking type refinement, but if you assigned the standalone `{List}CreateInput`/`{List}UpdateInput` types into a stricter local type, a nullable scalar may now be inferred as `T | null`:
+
+```typescript
+// A nullable text() field on Post now generates:
+export type PostCreateInput = {
+  title: string // required scalar — unchanged
+  content?: string | null // nullable scalar — now includes `| null`
+}
+```
--- a/packages/cli/src/generator/__snapshots__/types.test.ts.snap
+++ b/packages/cli/src/generator/__snapshots__/types.test.ts.snap
@@ -43,11 +43,11 @@ export type UserOutput = {
 export type User = UserOutput
 
 export type UserCreateInput = {
-  name?: string
+  name?: string | null
 }
 
 export type UserUpdateInput = {
-  name?: string
+  name?: string | null
 }
 
 export type UserWhereInput = Prisma.UserWhereInput
@@ -302,12 +302,12 @@ export type Post = PostOutput
 
 export type PostCreateInput = {
   title: string
-  content?: string
+  content?: string | null
 }
 
 export type PostUpdateInput = {
   title?: string
-  content?: string
+  content?: string | null
 }
 
 export type PostWhereInput = Prisma.PostWhereInput
@@ -565,12 +565,12 @@ export type UserOutput = {
 export type User = UserOutput
 
 export type UserCreateInput = {
-  name?: string
+  name?: string | null
   posts?: { connect: Array<{ id: string }> }
 }
 
 export type UserUpdateInput = {
-  name?: string
+  name?: string | null
   posts?: { connect: Array<{ id: string }>, disconnect: Array<{ id: string }> }
 }
 
@@ -835,12 +835,12 @@ export type PostOutput = {
 export type Post = PostOutput
 
 export type PostCreateInput = {
-  title?: string
+  title?: string | null
   author?: { connect: { id: string } }
 }
 
 export type PostUpdateInput = {
-  title?: string
+  title?: string | null
   author?: { connect: { id: string } } | { disconnect: true }
 }
 
@@ -1482,12 +1482,12 @@ export type Post = PostOutput
 
 export type PostCreateInput = {
   title: string
-  content?: string
+  content?: string | null
 }
 
 export type PostUpdateInput = {
   title?: string
-  content?: string
+  content?: string | null
 }
 
 export type PostWhereInput = Prisma.PostWhereInput
@@ -1744,11 +1744,11 @@ export type UserOutput = {
 export type User = UserOutput
 
 export type UserCreateInput = {
-  name?: string
+  name?: string | null
 }
 
 export type UserUpdateInput = {
-  name?: string
+  name?: string | null
 }
 
 export type UserWhereInput = Prisma.UserWhereInput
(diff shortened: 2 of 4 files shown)


borisno2 (commented):
## Review verdict: APPROVE (posted as COMMENT — GitHub identity matches the PR author, so a formal Approve is blocked)

Thorough review of the generator refactor in `packages/cli/src/generator/types.ts`. The change does exactly what it claims: consolidates scalar create/update input member rendering (name, `?`, type, `| null`) into one helper, `renderScalarInputMember`, used by both the standalone `{List}CreateInput`/`{List}UpdateInput` exports and the call-site write-`data` override, and aligns the standalone exports onto the override's `| null` representation. No correctness regressions found.

### Risk-area confirmations

**(a) #599's Date→calendarDay compile error still holds — CONFIRMED.**
- `shouldNarrowScalarWrite` and `mapFieldTypeToTypeScript` are unchanged; the helper still derives the scalar type from the field's own `getTypeScriptType().type`, so `calendarDay` still narrows to `string` in the override.
- The #599 guard, `types-write-narrowing.test.ts`, is **not in this diff** (untouched). It runs the real generator and a real `tsc` compile of a fixture with `@ts-expect-error` markers, asserting `day: new Date()` is rejected on create/update/createMany while `string` and decimal `number`/`string` compile.
- That test's structural assertion (`day?: string | null`, line 280) shows the override **already** emitted `| null` for nullable scalars before this PR — so the override was not weakened; only the standalone exports were brought up to match. The narrowing is intact.

**(b) Standalone + override nullability provably in sync — CONFIRMED, drift-guard test exists.**
- `renderScalarInputMember` is a genuine single source of truth: `generateCreateInputType`, `generateUpdateInputType`, and `buildScalarOverrideMembers` all delegate to it (no duplicated rendering logic remains — the old inline blocks were removed).
- New `create/update input nullability consolidation (#608)` describe block pins: nullable scalar is `content?: string | null` in standalone CreateInput, standalone UpdateInput, and both `data` overrides; required `title` stays `title: string` (create) / `title?: string` (update) and asserts `not.toContain('title?: string | null')`; plus a `>= 4` lock-step count so the two paths can't silently drift again.
- In the helper, `nullability` is `''` whenever `!nullable`, so a required scalar can never gain `| null`. The create-vs-update `?` optionality and the `defaultValue === undefined` check from #599 are preserved verbatim.

### Other checks (all pass)
- **No behavior change beyond nullability representation:** decimal/json stay on Prisma's input type (`shouldNarrowScalarWrite` returns false for them, unchanged); relationship and multi-column/`getColumnNames` fields untouched in both the code and the snapshot diff.
- **Snapshot churn is correct and intentional:** spot-checked ~10 hunks — every changed line is exactly a nullable scalar `?: T` → `?: T | null`. No required field gained `| null`; every relationship line (`author?: { connect... }`, `posts?: { connect..., disconnect... }`) is unchanged.
- **No `any` / no new casts in generated output or new code; ESM `.js` import extensions intact** (`./prisma.js`).
- **Changeset bump:** `minor` for `@opensaas/stack-cli` is the right call — nullable scalars in the public standalone `{List}CreateInput`/`{List}UpdateInput` exports now include `| null`, a visible (non-breaking) public-type refinement. Patch would understate it.

### Minor note (non-blocking, not introduced by this PR)
The two paths share *rendering* but not the *field-skip predicate*: the standalone exports render every non-relationship scalar (including `decimal`/`json`/multi-column), while the override skips those via `shouldNarrowScalarWrite`. So the "lock-step" guarantee is exact only for fields the override also narrows (the common scalar case the new test covers with `text()`); for a nullable `decimal`/`json` field the standalone export emits the OpenSaaS type + `| null` while the override leaves it on Prisma's type. This asymmetry pre-dates this PR and is consistent with the documented decimal/json intent — flagging only so a future reader doesn't over-read the drift-guard as covering decimal/json.

**Recommendation: approve and merge.** Clean, well-tested consolidation; the #599 guarantee is preserved and the two sources of truth are now provably reconciled.

---

PR #610, opened 2026-06-27
ci: fix Release changelog generation (changelog-git, no GitHub API)

borisno2:
## Problem

The **Release** workflow (`changesets/action@v1` → `changeset version`) is failing **deterministically** on push to `main` — reruns don't help:

```
🦋 error Failed to parse data from GitHub
🦋 error Invalid response body while trying to fetch https://api.github.com/graphql: Premature close
   at @changesets/get-github-info/.../changesets-get-github-info.cjs.js:185
```

`.changeset/config.json` used `@changesets/changelog-github`, which calls the GitHub GraphQL API **once per changeset** to enrich each entry with PR/author links. With the batch of accumulated changesets, the request is dropped every time, so `changeset version` fails and the **Version Packages** PR is never created — blocking all releases. (Changesets reports it "escaped applying the changesets," so nothing is corrupted — it just can't proceed.)

## Fix

Switch the changelog generator to **`@changesets/changelog-git`**, which builds entries from local git history with **no network calls** — removing the failure mode entirely.

- `.changeset/config.json`: `changelog` → `"@changesets/changelog-git"`.
- `package.json`: swap the devDependency `@changesets/changelog-github@^0.7.0` → `@changesets/changelog-git@^0.2.1` (already present in the lockfile).
- `pnpm-lock.yaml`: synced.

Verified locally: `@changesets/changelog-git` resolves; `@changesets/changelog-github` is removed from root resolution; `prettier --check` and `manypkg check` pass.

## Trade-off

Published changelog entries become **commit/changeset-summary based** and lose the `Thanks [@author]! — [#PR]` GitHub enrichment. This is the cost of removing the live-API dependency that's currently breaking every release.

If keeping GitHub attribution matters more than this robustness, the alternative is to leave `changelog-github` in place and make the release resilient another way (e.g. self-hosted retry around the version step) — happy to take that route instead.

No changeset is included: this changes root tooling/config only, not a published `packages/*`.

---

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://unpkg.com/@changesets/config@3.1.1/schema.json",
-  "changelog": ["@changesets/changelog-github", { "repo": "OpenSaasAU/stack" }],
+  "changelog": "@changesets/changelog-git",
   "commit": false,
   "fixed": [["@opensaas/stack-*"]],
   "linked": [],
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
   "author": "",
   "license": "MIT",
   "devDependencies": {
-    "@changesets/changelog-github": "^0.7.0",
+    "@changesets/changelog-git": "^0.2.1",
     "@changesets/cli": "^2.30.0",
     "@eslint/js": "^9.39.4",
     "@manypkg/cli": "^0.25.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,9 +8,9 @@ importers:
 
   .:
     devDependencies:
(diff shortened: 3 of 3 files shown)


borisno2:
Closing without merging — per maintainer preference we're keeping `@changesets/changelog-github` (to retain the `Thanks @author — #PR` changelog enrichment) and treating the release failure as a transient GitHub GraphQL outage to wait out, rather than switching the changelog generator to `changelog-git`.

The diagnosis stands if it recurs persistently: the failure is the per-changeset `api.github.com/graphql` enrichment call (`Premature close`), independent of changeset content. If waiting doesn't clear it, the alternative is to keep `changelog-github` and add resilience around the version step rather than swapping the generator.

---

PR #616, opened 2026-06-28
feat(core): add context.transaction interactive hook-firing transaction

### LINKED ISSUE #614: Context API has no interactive (hook-firing) transaction — can't atomically enforce concurrency-sensitive constraints
### Summary

The stack `Context` exposes no interactive transaction that runs `context.db.*` writes (with their access checks + list hooks) atomically. Keystone 6 provided `context.transaction(async (tx) => { … })`, where the `tx` ran the access-scoped `db` API **inside a single DB transaction with a chosen isolation level**, and list hooks still fired. The stack's only transaction is the raw underlying `prisma.$transaction`, which **bypasses the stack's access filters + list `validateInput`/`resolveInput` hooks** — i.e. it drops the security/validation boundary.

This makes it impossible to atomically enforce a **concurrency-sensitive invariant** (e.g. a capacity/quota gate: "re-read count, check `< capacity`, then create — atomically, serializable") while preserving the hook-based access boundary.

### Concrete use case (where it blocks us)

A callback-slot booking flow must guarantee **exactly N concurrent bookings succeed against a capacity-N slot**. On Keystone 6 this was a `context.transaction` with `Serializable` isolation + a serialization-failure retry loop: re-read the live booking count *inside* the tx, gate on `count >= capacity`, create the `CallbackBooking`, all atomic. Two racing tx that both read `count = N-1` and both insert → one hits a serialization failure and retries → exactly N commit.

On the stack there is no equivalent. Replacing it with `context.db` writes + a post-create recount + compensating-delete cannot hold the guarantee: under true contention every racer creates, every racer then sees `count > capacity`, and every racer rolls itself back → **0 succeed** (or, with naive ordering, over-books). Using raw `prisma.$transaction` would restore atomicity but bypass the list hooks that are the access/validation boundary.

### Ask

An interactive transaction on the stack `Context` — e.g. `context.transaction(async (txContext) => { … }, { isolationLevel })` — where `txContext.db.*` performs access-checked, hook-firing writes within one DB transaction, with selectable isolation (at least `Serializable`) and a way to detect/propagate serialization failures for retry. (Equivalent to Keystone 6's `context.transaction`.)

### Workarounds considered (and why they're insufficient)
- `prisma.$transaction` (raw): atomic, but bypasses access filters + list hooks.
- create + recount + compensating-delete via `context.db`: not atomic; cannot guarantee exactly-N under contention.

### Environment
- `@opensaas/stack-core@0.24.0`

borisno2:
## Summary

Adds `context.transaction(fn, options)` to the stack `Context` — an interactive, **hook-firing** transaction. The `txContext.db.*` operations are access-checked and run list/field hooks exactly like the request context, but persist against **one** Prisma interactive transaction, so every write in the callback is atomic (a throw anywhere rolls it all back). This is the secured alternative to raw `prisma.$transaction`, which bypasses access control and hooks.

It enables concurrency-sensitive invariants (e.g. a capacity/quota gate) to be enforced atomically without dropping the access/hook boundary — the gap reported in #614 (Keystone 6's `context.transaction` had no stack equivalent).

```typescript
const result = await context.transaction(
  async (tx) => {
    const count = await tx.db.booking.count({ where: { slotId } })
    if (count >= capacity) return { booked: false }
    return { booked: true, item: await tx.db.booking.create({ data: { slotId } }) }
  },
  { isolationLevel: 'Serializable' },
)
```

## What changed

- **`context.transaction(fn, options?)`** added to `getContext`'s returned context.
  - Reuses the ADR-0010 rebind primitive (`buildDbDelegate` via the `getContext` factory) to bind the secured context to the transaction client — no new persistence logic.
  - `options` (`isolationLevel` incl. `'Serializable'`, `maxWait`, `timeout`) pass through to Prisma.
  - **Serialization failures (`P2034`) propagate** to the caller — retry is caller-owned (matching Keystone). Built-in retry is intentionally out of scope.
  - **Nested `context.db` writes join** the outer transaction via the Write Pipeline's existing "no `$transaction` on a tx client → run directly" fallback; calling `transaction()` on a client with no `$transaction` runs `fn` directly with identical hook/access semantics.
  - Plugin runtimes are **reused, not re-run**, for the tx context (new internal `_sharedPlugins` path).
- Exports new types `StackContext`, `TransactionOptions`, `TransactionIsolationLevel` from `@opensaas/stack-core`.
- Adds **ADR-0012** (cross-referencing ADR-0010), core + root `CLAUDE.md` docs, and a minor changeset.

## Test plan

- [x] New `tests/interactive-transaction.test.ts` (10 tests): method presence, returns callback result, **atomic rollback on throw**, **isolationLevel passthrough**, **access control enforced** (denied create → `null`), **list hooks fire**, **serialization-failure propagation**, session/`sudo()` on the tx context, and the no-`$transaction` fallback.
- [x] A **Serializable capacity-gate concurrency test**: 4 racers against capacity-2 commit **exactly 2** (deterministic predicate-conflict emulation + caller retry).
- [x] Full core suite green (706 + 10 tests), `pnpm build` (all 11 tasks), `pnpm lint` (0 errors), `pnpm format`, `pnpm manypkg fix`.

Closes #614

---
_Generated by [Claude Code](https://claude.ai/code/session_01UHLArNoEjHjGZnHpSxmu5V)_


borisno2:
## Code review (subagent) — verdict: approve-with-nits → nits addressed

An independent review pass checked the implementation adversarially across atomicity, nesting, TDZ, error propagation, sudo binding, and type safety. Summary:

| Concern | Result |
| --- | --- |
| **Atomicity & nesting** | ✅ Correct. Inside `prisma.$transaction`, the rebuilt `txContext.db.*` writes reach `runInTransaction(tx, …)`; a Prisma tx client has no `$transaction`, so the existing fallback runs them directly on `tx` — they **join** the outer transaction. No double-transaction, no lost rollback. |
| **No-`$transaction` fallback (`return fn(returned)`)** | ✅ Safe — `transaction` is a hoisted decl, `returned` is only dereferenced at call time (after `getContext` returns). No TDZ. Plugin runtimes receive the internal `AccessContext`, which has no `transaction` method. |
| **Serialization-failure propagation** | ✅ Confirmed. Nothing in the tx path converts a throw to `null`; only access-denial short-circuits return `null`. |
| **`sudo()` inside a tx** | ✅ Stays bound to `tx` (sudo writes commit/roll back with the transaction). |
| **Type safety** | ✅ Clean. The two internal casts are localized; exported types (`StackContext`/`TransactionOptions`/`TransactionIsolationLevel`) contain no `any`. |

### Nits raised (both addressed in `4608965`)

1. **Plugin services are not tx-bound** (the one substantive caveat). Plugin `runtime()` services stay bound to the top-level client by design (ADR-0012). The reviewer noted the *write-escape* consequence was understated: a plugin service that **writes** would escape the transaction and survive a rollback. → Added an explicit caveat to the `transaction()` docstring **and** ADR-0012: use `txContext.db` for writes that must be atomic. (No shipped plugin writes from `runtime()` today; this is forward-looking.)
2. **Sudo-write coverage was thin** — the sudo test only checked `_isSudo`. → Added a test asserting a `sudo()` write inside a transaction **commits on success and rolls back on throw**.

Core suite now 11/11 in the new file (707 total green locally).

---

Issue #618, opened 2026-06-28
file()/image() fields are required on create — getZodSchema union with z.undefined() doesn't make the key optional in Zod 4


borisno2:
### Summary

`file()` and `image()` (`@opensaas/stack-storage@0.24.0`) return their create/update validation schema as:

```ts
return z.union([fileMetadataSchema, z.null(), z.undefined()])   // file()
return z.union([imageMetadataSchema, z.null(), z.undefined()])  // image()
```

In **Zod 4**, a union that merely *accepts* `undefined` as a value does NOT make the object key optional. So when stack-core builds the create-input via `generateZodSchema` (`z.object({ <field>: getZodSchema(...) })`), an **omitted** file/image field fails validation:

```
ValidationError: Validation failed: image: Invalid input: expected nonoptional, received undefined
```

even though the field is conceptually optional (its `getTypeScriptType()` returns `{ type: 'ImageMetadata | null', optional: true }`, and Keystone's equivalent `image({})`/file fields were nullable/optional on create).

### Impact

Any list with a file/image field cannot be created without supplying that field — `createMany` (which drops the multi-field) and `create` (omitting it) both throw. This breaks seeds and any create path that legitimately has no file/image (a teacher with no photo, etc.). Passing `image: null` doesn't help for multi-column (`db: { columns: 'keystone' }`) fields — the logical key is stripped before validation, so it reads back as `undefined`.

### Fix

Return `.nullish()` (= `.nullable().optional()`) instead of the bare union, so the object key is genuinely optional in Zod 4:

```ts
return fileMetadataSchema.nullish()
return imageMetadataSchema.nullish()
```

This validates an omitted/null/undefined value AND marks the key optional. Verified as a local pnpm patch — omitted image now validates and the record is created with a null image.

### Environment
- `@opensaas/stack-storage@0.24.0`, `zod@4`

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** `file()` and `image()` fields are erroneously required on create/update because their `getZodSchema()` returns a union containing `z.undefined()`, which does not make the object key optional in Zod 4.

**Current behavior:**
[...]

PR #619, opened 2026-06-28
fix(storage): make file()/image() fields optional on create/update

borisno2:
Implements #618.

## Problem

`file()` and `image()` (`@opensaas/stack-storage`) returned their create/update validation schema as:

```ts
return z.union([fileMetadataSchema, z.null(), z.undefined()])   // file()
return z.union([imageMetadataSchema, z.null(), z.undefined()])  // image()
```

In **Zod 4**, a union that merely *accepts* an `undefined` value does **not** make the object key optional. Core composes create/update input via `z.object({ <field>: getZodSchema(...) })` (`generateZodSchema` in `packages/core/src/validation/schema.ts`) with no extra wrapping, so an **omitted** file/image field failed validation:

```
ValidationError: Validation failed: image: Invalid input: expected nonoptional, received undefined
```

This broke seeds and any create/update path that legitimately has no file/image. It's the same class of bug already fixed for core fields under #570.

## Fix

Return `metadataSchema.nullish()` (= `.nullable().optional()`) instead of the bare union, so the object **key** is genuinely optional — matching the established idiom every optional core field (`text`, `integer`, `json`, `checkbox`, `timestamp`, `select`) uses. Out of scope: no "required file/image" capability, no metadata-shape or multi-column-layout changes, no other field types touched.

## Acceptance criteria

- [x] Creating a record with a `file()` field, omitting it, succeeds and stores null.
- [x] Creating a record with an `image()` field, omitting it, succeeds and stores null.
- [x] Same holds for UPDATE operations that omit the field.
- [x] Explicit `null` / `undefined` still validate on create and update, including multi-column (`db: { columns: 'keystone' }`) fields.
- [x] A correctly-shaped metadata object still validates.
- [x] A malformed metadata object still fails validation.
- [x] Regression test covers the omitted-field create case for both `file()` and `image()`.

## Tests

New `packages/storage/tests/field-zod-schema.test.ts` reproduces core's `z.object({ <field>: getZodSchema() })` composition and covers omitted / null / undefined / valid / malformed across create and update, single- and multi-column modes.

- `pnpm --filter @opensaas/stack-storage test`: 147 passed
- `pnpm --filter @opensaas/stack-core test`: 707 passed
- `pnpm lint`: 0 errors (3 pre-existing unrelated warnings)
- Builds: storage + core green

Changeset added (`@opensaas/stack-storage` patch).

https://claude.ai/code/session_01SigyUcdxbBsdeLVWiFbYtS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SigyUcdxbBsdeLVWiFbYtS)_

### DIFF
--- a/.changeset/file-image-nullish.md
+++ b/.changeset/file-image-nullish.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-storage': patch
+---
+
+Fix file()/image() fields being required on create/update: their Zod schema now uses key-optionality (.nullish()) so an omitted field validates and stores null (Zod 4).
--- a/packages/storage/src/fields/index.ts
+++ b/packages/storage/src/fields/index.ts
@@ -302,8 +302,11 @@ export function file<TTypeInfo extends TypeInfo = TypeInfo>(
         metadata: z.record(z.string(), z.unknown()).optional(),
       })
 
-      // Allow null or undefined values
-      return z.union([fileMetadataSchema, z.null(), z.undefined()])
+      // Allow null or undefined values. `.nullish()` (= `.nullable().optional()`)
+      // makes the object KEY optional in Zod 4 — a bare union that merely accepts
+      // an `undefined` value does NOT (see issue #618, and the #570 precedent in
+      // core's validation tests).
+      return fileMetadataSchema.nullish()
     },
 
     getPrismaType: (_fieldName: string) => {
@@ -506,8 +509,11 @@ export function image<TTypeInfo extends TypeInfo = TypeInfo>(
           .optional(),
       })
 
-      // Allow null or undefined values
-      return z.union([imageMetadataSchema, z.null(), z.undefined()])
+      // Allow null or undefined values. `.nullish()` (= `.nullable().optional()`)
+      // makes the object KEY optional in Zod 4 — a bare union that merely accepts
+      // an `undefined` value does NOT (see issue #618, and the #570 precedent in
+      // core's validation tests).
+      return imageMetadataSchema.nullish()
     },
 
     getPrismaType: (_fieldName: string) => {
--- a/packages/storage/tests/field-zod-schema.test.ts
+++ b/packages/storage/tests/field-zod-schema.test.ts
@@ -1,0 +1,133 @@
+import { describe, it, expect } from 'vitest'
+import { z } from 'zod'
+import { image, file } from '../src/fields/index.js'
+import type { ImageMetadata, FileMetadata } from '../src/config/types.js'
+
+/**
+ * Regression tests for issue #618: file()/image() fields were erroneously
+ * required on create/update because getZodSchema() returned
+ * `z.union([metadataSchema, z.null(), z.undefined()])`. In Zod 4, a union that
+ * merely accepts an `undefined` *value* does NOT make the object KEY optional —
+ * an omitted key fails with "expected nonoptional, received undefined". The fix
+ * returns `metadataSchema.nullish()` so the key is genuinely optional.
+ *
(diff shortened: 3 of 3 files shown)


borisno2 (commented):
## Review — Verdict: APPROVE (posted as Comment; formal Approve blocked because the integration identity authored the PR)

Tightly-scoped fix: `file()`/`image()` `getZodSchema()` now returns `metadataSchema.nullish()` instead of `z.union([metadataSchema, z.null(), z.undefined()])`, making the **object key** genuinely optional so an omitted file/image field validates on create/update. Plus a new regression test that reproduces core's `z.object({ <field>: getZodSchema() })` composition. The change matches the idiom every optional core field already uses and is strictly safer than the bare union.

### Verification performed
I reproduced the exact composition core uses — `validateFieldRules` → `validateWithZod` → `generateZodSchema` builds `z.object({ [fieldName]: getZodSchema(fieldName, op) })` (`packages/core/src/validation/schema.ts`, `packages/core/src/hooks/index.ts:760`) — and ran the real `file()`/`image()` builders against zod 4.3.6:

- **Omitted key validates** — single-column AND multi-column (`db: { columns: 'keystone' }`), create AND update. ✅
- **Explicit `null` / `undefined` validate.** ✅
- **A correctly-shaped metadata object validates.** ✅
- **A malformed metadata object still FAILS** (e.g. `{ avatar: { filename: 'x.png' } }` → `success: false`). The fix does NOT loosen validation. ✅
- **`.nullish()` makes the key optional** (`z.object({ avatar: meta.nullish() }).safeParse({})` → success), which a bare `z.union([…, z.undefined()])` is the wrong tool for. ✅

### Risk areas (all clear)
- **Zod 4 key-optionality:** Confirmed `.nullish()` makes the composed object key optional — an omitted key passes, not merely an `undefined` value. ✅
- **Multi-column path:** In the real write pipeline the logical key is split into per-part columns and deleted in field `resolveInput` (`hooks/index.ts` ~L555–558) *before* `validateFieldRules` runs (`hook-pipeline.ts:158`), so by validation time the key is omitted — exactly the case the fix now permits. The test asserts against the schema directly (logical key present) rather than the post-split payload; that's a touch artificial but not wrong, and the schema is correct either way. ✅
- **No `any` / no type casting introduced:** the source change is a single return-expression swap; the test types the field param structurally (`{ getZodSchema?: (...) => z.ZodTypeAny }`) with no `any` and no casts. ✅
- **Test fidelity:** the new test reproduces `z.object({ <field>: getZodSchema() })` (not a bare schema call) and covers both `file()` and `image()` for the omitted-on-create case (and update, null, undefined, valid, malformed, multi-column). ✅
- **Changeset present** (`@opensaas/stack-storage` patch). ✅

### Non-blocking notes
1. **Could not reproduce the original "expected nonoptional, received undefined" failure in isolation.** With zod **4.3.6**, even the *old* `z.union([meta, z.null(), z.undefined()])` wrapped in a plain `z.object` already accepted an omitted key in my repro. The reported failure presumably depended on a slightly different zod patch or extra wrapping. This does not undermine the fix — `.nullish()` is the correct, conventional idiom and is provably safer — but the PR's "expected nonoptional" quote may not reproduce on the currently-pinned zod. Worth a one-line confirmation that the repro was seen on the pinned version.
2. **Multi-column test is schema-level, not pipeline-level.** Since the logical key is stripped before validation in keystone-columns mode, an end-to-end test (record create with `db: { columns: 'keystone' }` omitting the field) would more faithfully guard the real path. Optional follow-up; the unit assertion is still valid.

No correctness, regression, convention, or type-safety issues found. Recommend merge.

---

Issue #615, opened 2026-06-28
select() defaultValue is not applied before validateInput on create (diverges from Keystone)


borisno2:
### Summary

On a `select()` field with both `validation: { isRequired: true }` and a `defaultValue`, the stack runs the `isRequired` validation **before** the field's `defaultValue` is applied, so a create that OMITS the field fails with "<Field> is required" — even though a default exists. Keystone 6 applies the select `defaultValue` first (default-then-validate), so an omitted value resolves to the default and passes.

### Repro

```ts
kind: select({
  validation: { isRequired: true },
  options: [{ value: 'STANDARD', label: 'Standard' }, { value: 'TRIAL', label: 'Trial' }],
  defaultValue: 'STANDARD',
})
```

Create the row WITHOUT `kind`:
- Keystone 6: `kind` resolves to `STANDARD`, create succeeds.
- Stack `@opensaas/stack-core@0.24.0`: `ValidationError: Validation failed: Kind must be one of: STANDARD, TRIAL` (thrown from the write pipeline's `validateInput`/hook pipeline before the default is applied). Reproduces on both top-level and nested-relation creates.

### Impact

Any list that models a required-with-default select (very common — e.g. a `kind`/`status`/`type` enum-ish column with a default) cannot be created by callers that rely on the default, which all Keystone callers do. Forces every call site to pass the value explicitly, or a per-list `resolveInput` workaround (below).

### Workaround

A list-level `resolveInput` hook that fills the field before `validateInput`:

```ts
hooks: {
  resolveInput: (args) =>
    args.operation === 'create' && args.resolvedData.kind == null
      ? { ...args.resolvedData, kind: 'STANDARD' }
      : args.resolvedData,
}
```

(Pipeline order is `resolveInput → validateInput`, so this works — but it duplicates the field's own `defaultValue`.)

### Ask

Apply a field's `defaultValue` to omitted inputs BEFORE `validateInput` on create (matching Keystone 6's resolve-then-validate ordering), so `isRequired` + `defaultValue` behaves as it does in Keystone.

### Environment
- `@opensaas/stack-core@0.24.0`

borisno2:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** A field's `defaultValue` is not applied to omitted inputs before validation runs on create, so a required field with a default fails `isRequired` validation instead of resolving to its default (diverges from Keystone's resolve-then-validate ordering).

**Current behavior:**
[...]

PR #620, opened 2026-06-28
fix(core): apply field defaultValue before create validation (#615)

borisno2:
Implements #615.

## Problem

A field's `defaultValue` was only realised as a Prisma `@default(...)` applied by the DB at write time — **after** the write pipeline's validation phase. Because validation never consulted `defaultValue`, a create that **omitted** a required-with-default field (e.g. `select({ validation: { isRequired: true }, defaultValue: 'STANDARD' })`) failed `isRequired` validation instead of resolving to its default. This diverges from Keystone 6's resolve-then-validate ordering and affected **every** field type that supports `defaultValue`, not just `select`.

## Fix (resolve-then-validate, Keystone parity)

A single shared helper `applyCreateDefaults()` (`packages/core/src/context/apply-defaults.ts`) fills omitted fields with their declared `defaultValue` during the **resolve** phase — after `resolveInput` hooks, before validation:

- **Top-level create**: wired into the Hook Pipeline (`hook-pipeline.ts`) between field `resolveInput` and list `validate`.
- **Nested-relation create**: wired into the nested create path (`nested-operations.ts`) at the same point.

### Guard rails
- **Create only** — update never injects defaults for omitted fields (helper invoked only when `operation === 'create'`).
- **Explicit values preserved** — only keys whose value is `undefined` (omitted) are filled; an explicitly-provided value, **including explicit `null`**, is left untouched.
- **Skips** virtual, system (`id`/`createdAt`/`updatedAt`) and relationship fields.
- **Timestamp `{ kind: 'now' }` sentinel** is *not* injected (it is a DB-level `@default(now())` request, not a literal); a concrete `Date` default *is* injected.
- **Prisma `@default(...)` generation is unchanged.**

## Tests
- New `tests/default-value-create.test.ts`: Hook-Pipeline unit coverage (omitted `select`/`text`/`integer`/`checkbox` resolve to defaults; explicit value and explicit `null` preserved; update does not inject; required-without-default still throws) plus full `context.db` create through a tx-aware mock for **top-level and nested-relation** creates.
- Updated two `tests/sudo.test.ts` assertions to reflect the now-correct injection of a field's `defaultValue: 0` into the create payload.
- Changeset added (patch, `@opensaas/stack-core`).

All 714 core tests pass; build + lint clean.

https://claude.ai/code/session_01SigyUcdxbBsdeLVWiFbYtS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SigyUcdxbBsdeLVWiFbYtS)_

### DIFF
--- a/.changeset/spotty-llamas-cheer.md
+++ b/.changeset/spotty-llamas-cheer.md
@@ -3,3 +3,5 @@
 ---
 
 Apply a field's defaultValue to omitted inputs before create validation (resolve-then-validate, matching Keystone), so isRequired + defaultValue no longer fails on create.
+
+Note: because an omitted-but-defaulted field is now filled into `resolvedData` before validation, that field's create-side field-level `beforeOperation`/`afterOperation` hooks (gated on the field key being present in `resolvedData`) now fire for defaulted fields where they previously would not.
--- a/.changeset/spotty-llamas-cheer.md
+++ b/.changeset/spotty-llamas-cheer.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Apply a field's defaultValue to omitted inputs before create validation (resolve-then-validate, matching Keystone), so isRequired + defaultValue no longer fails on create.
--- a/packages/core/src/context/apply-defaults.ts
+++ b/packages/core/src/context/apply-defaults.ts
@@ -1,0 +1,79 @@
+import type { FieldConfig } from '../config/types.js'
+
+/**
+ * Apply field `defaultValue`s to omitted inputs on CREATE — the runtime half of
+ * the resolve-then-validate ordering (Keystone 6 parity, issue #615).
+ *
+ * A field's `defaultValue` is otherwise only realised as a Prisma `@default(...)`
+ * applied by the database at write time, which is AFTER the write pipeline's
+ * validation phase has already run. That ordering means a required-with-default
+ * field (e.g. `select({ validation: { isRequired: true }, defaultValue: 'X' })`)
+ * fails `isRequired` validation on an omitted input even though a default exists.
+ *
+ * This helper closes that gap: in the resolve phase (after `resolveInput` hooks,
+ * before validation) it fills `resolvedData[field]` with the field's
+ * `defaultValue` ONLY when the field was OMITTED (value is `undefined`). It is a
+ * SINGLE shared mechanism used by both the top-level create path (Hook Pipeline)
+ * and the nested-relation create path.
+ *
+ * Guard rails (each acceptance-criteria-driven):
+ *   - CREATE only. Update never injects defaults for omitted fields (the caller
+ *     only invokes this for `operation === 'create'`).
+ *   - Explicitly-provided values are preserved. A key present in `resolvedData`
+ *     — INCLUDING an explicit `null` — is left untouched; only `undefined`
+ *     (omitted) keys are filled.
+ *   - Virtual, system (`id`/`createdAt`/`updatedAt`) and relationship fields are
+ *     skipped — they have no scalar `defaultValue` to inject and relationships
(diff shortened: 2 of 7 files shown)


borisno2 (commented):
## Review — `/review` (verdict: APPROVE)

I reviewed this strictly from the PR diff and the code it touches (no prior context). The fix is well-scoped, correctly placed in the resolve-then-validate span, and the guard rails hold up under scrutiny. **Verdict: APPROVE.** (Posting as a Comment review rather than a formal Approve since the change originates from the integration identity; the verdict is an explicit approve.)

### What it does
Adds a single shared helper `applyCreateDefaults()` that fills omitted (`undefined`) fields with their declared `defaultValue` during the resolve phase — after `resolveInput`, before validation — wired into both the top-level Hook Pipeline (Phase 1.75) and the nested-relation create path (`processNestedCreate` step 3.5). This fixes #615 (a required-with-default field failing `isRequired` on an omitted create input), matching Keystone's ordering.

### Risk-area verification (all pass)
- **Placement / create-only.** Injection runs after field `resolveInput` and before list/field `validate` and `validateFieldRules`, in both paths. Top-level is gated by `if (operation === 'create')`; the nested helper lives inside `processNestedCreate`, which only ever runs with `'create'`. Update never injects. Confirmed `validateFieldRules` validates `resolvedData`, so injected defaults satisfy `isRequired`. ✓
- **Explicit-value preservation incl. `null`.** `if (resolvedData[fieldKey] !== undefined) continue` skips any present key, so an explicit `null` is preserved and not overwritten. Falsy literals (`false`, `0`, `''`) are correctly injected since the guard tests `undefined`, not falsiness. ✓
- **Nested create covered.** Wired and tested (`profile: { create: {} }` resolves the nested `kind` default). ✓
- **Timestamp `{ kind: 'now' }` sentinel.** `isNowSentinel` matches `{ kind: 'now' }` exactly the shape `timestamp()` emits and skips it (left to Prisma `@default(now())`); a concrete `Date` default is still injected, which is a valid Prisma create value. ✓
- **Field skips.** Virtual fields are reliably skipped — `virtual()` sets `virtual: true` (required on `VirtualField`), so `if (fieldConfig.virtual)` matches. System (`id`/`createdAt`/`updatedAt`) and `relationship` fields are skipped. ✓
- **Prisma `@default(...)` generation unchanged.** No generator files are touched by this PR. ✓
- **House rules.** No `any` and no type casting introduced in the new helper; `defaultValue` is read via `'defaultValue' in fieldConfig` against the typed config and the lone `unknown` narrowing (`isNowSentinel`) is internal. Changeset present (patch, `@opensaas/stack-core`). ✓
- **All field-type defaults safe.** Audited every field type: `text`/`select`/`calendarDay` = string, `decimal` = string (Prisma coerces), `integer` = number, `checkbox` = boolean, `json` = literal, `timestamp` = the only object sentinel and it's guarded. No raw-object corruption path exists. ✓

### Test coverage
Exercises both the Hook-Pipeline unit surface and full `context.db` creates: omitted `select`/`text`/`integer`/`checkbox` resolve to defaults; explicit value and explicit `null` are preserved; update does not inject; required-without-default still throws; and **both top-level and nested-relation** creates store the default through the full pipeline. The two `sudo.test.ts` assertion updates correctly reflect the now-injected `views: 0`. Good coverage of every flagged case.

### Minor observation (non-blocking, no change required)
Injecting a default into `resolvedData` means a field that was *omitted but has a default* now appears in `resolvedData`, so its create-side field-level `beforeOperation`/`afterOperation` hooks (gated by `fieldKey in resolvedData`) will now fire where previously they would not. This is the correct Keystone-aligned behavior (the field genuinely has a resolved value now), but it is a subtle behavioral change for anyone relying on those hooks only firing for explicitly-provided fields. Worth a mention in the changeset/notes, not a code change.

No correctness, regression, or convention issues found.

---

PR #633, opened 2026-07-10
fix(core): strip virtual fields from include before Prisma call

### LINKED ISSUE #628: Virtual fields in `include` blocks cause runtime error: "Unknown field" from Prisma
### Summary

When a virtual field (a computed field defined in a list schema) is included in the `include` block of a `context.db.*` query, Prisma throws a runtime error:

```
Error: Unknown field '<fieldName>' for include statement on model '<ModelName>'. Available options are marked with ?.
```

However, if the virtual field is **not** included in the `include` block, TypeScript reports a type error when the field is accessed on the result, because the generated Prisma type doesn't include virtual fields.

This creates a no-win situation: include the virtual field → runtime error; don't include it → type error.

### Steps to Reproduce

1. Define a virtual field on a list, e.g.:
   ```ts
   // In a list definition
   name: virtual({
     field: graphql.field({
       type: graphql.String,
       resolve: (item) => `${item.firstName} ${item.surname}`,
     }),
   })
   ```

2. Query the list with the virtual field in `include`:
   ```ts
   const result = await context.db.myList.findMany({
     include: { name: true }
   })
   ```

3. Observe runtime error:
   ```
   Unknown field 'name' for include statement on model 'MyList'. Available options are marked with ?.
   ```

4. Remove `name` from `include` and access `result[0].name` → TypeScript error because `name` is not in the generated type.

### Expected Behaviour

Either:
- Virtual fields should be excluded from the `include` type signature (so TypeScript catches the mistake at compile time), **and** the result type should include virtual fields so they can be accessed without a type error; or
- The runtime should silently strip virtual field names from the `include` payload before passing it to Prisma.

### Workaround

Remove virtual fields from all `include` blocks and add `// @ts-expect-error` comments at every virtual field access site. This is error-prone and adds noise to the codebase.

### Version

- `@opensaas/stack-core`: 0.26.0

borisno2:
## Summary

- Virtual fields named in a caller-supplied `include` (e.g. `context.db.myList.findMany({ include: { myVirtualField: true } })`) used to be forwarded straight through to Prisma, which threw `Unknown field '<name>' for include statement` because virtual fields have no database column.
- Adds `stripVirtualFieldsFromInclude()` in `packages/core/src/access/access-filter.ts`, which recursively removes virtual-field keys from a Prisma `include` object while preserving real relationship includes (and their nested access-control filters) at every level.
- Applied as the final step in `context/index.ts` for both `findUnique` and `findMany`, after `include` is resolved via the fragment path, the access-controlled merge path, or the sudo passthrough path — so a virtual key never reaches the Prisma client regardless of which path produced it.
- The virtual field's value is unaffected: `filterReadableFields` already computes it unconditionally via `resolveOutput`, whether or not the field was named in `include`.

## Test plan

- [x] Regression tests added in `packages/core/tests/context.test.ts` (`virtual fields named in include no longer reach Prisma (#628)`):
  - `findMany`/`findUnique` with a virtual field alongside a real relationship in `include` — the virtual key is stripped from the Prisma call, the relationship include is untouched, and the virtual value is still returned.
  - The virtual value is populated even when omitted from `include`.
  - Sudo path also strips the virtual key (while leaving the real relationship passthrough unfiltered, matching existing sudo behaviour).
  - Read access control on a virtual field is still enforced (a field denied by read access is omitted from the result).
- [x] `pnpm test` — full `packages/core` suite passes (729 tests)
- [x] `pnpm build` (tsc) passes
- [x] `pnpm format` / `pnpm manypkg fix` — no changes needed beyond the diff
- [x] Changeset added (`@opensaas/stack-core`: patch)

Closes #628

---
_Generated by [Claude Code](https://claude.ai/code/session_01A6fWyGeMiexwmQLD2usMcn)_

### DIFF
--- a/.changeset/tame-plums-invite.md
+++ b/.changeset/tame-plums-invite.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix virtual fields named in `include`/`select` throwing a Prisma "Unknown field" error. Virtual field keys are now stripped from the query payload while their value is still computed via `resolveOutput`.
--- a/packages/core/src/access/access-filter.ts
+++ b/packages/core/src/access/access-filter.ts
@@ -251,3 +251,57 @@ export function mergeIncludeWithAccessControl(
 
   return merged
 }
+
+/**
+ * Remove keys that correspond to `virtual` fields from a Prisma `include`
+ * object, recursing into nested relationship includes using the related
+ * list's field configs.
+ *
+ * Virtual fields are computed in JavaScript (via `resolveOutput` in
+ * `field-visibility.ts`) and have no database column. Naming one in `include`
+ * type-checks — the generated `Include` type lists every field the config
+ * declares — but Prisma throws `Unknown field '<name>' for include statement`
+ * at runtime. This is applied as the final step on whatever `include` a read
+ * op ends up with (fragment-built, access-controlled merge, or sudo
+ * passthrough) so a virtual key never reaches the Prisma client, regardless
+ * of which path produced it. The virtual value is still computed
+ * unconditionally by `filterReadableFields`, so stripping the include key has
+ * no effect on whether the value appears in the result (#628).
+ */
+export function stripVirtualFieldsFromInclude(
+  include: Record<string, unknown> | undefined,
+  fieldConfigs: Record<string, FieldConfig>,
+  config: OpenSaasConfig,
+): Record<string, unknown> | undefined {
+  if (!include) return include
+
+  const result: Record<string, unknown> = {}
+
+  for (const [key, value] of Object.entries(include)) {
+    const fieldConfig = fieldConfigs[key]
+
+    // Virtual fields have no database column — drop them from the include.
+    if (fieldConfig?.virtual) continue
+
+    const isDeclaredRelationship =
+      fieldConfig?.type === 'relationship' && 'ref' in fieldConfig && !!fieldConfig.ref
+    const entry = asEntryObject(value)
+
+    if (isDeclaredRelationship && entry?.include) {
+      const relatedConfig = getRelatedListConfig(fieldConfig.ref as string, config)
+      if (relatedConfig) {
+        const nestedInclude = stripVirtualFieldsFromInclude(
+          entry.include,
+          relatedConfig.listConfig.fields,
+          config,
+        )
+        result[key] = { ...(value as Record<string, unknown>), include: nestedInclude }
+        continue
+      }
+    }
+
+    result[key] = value
+  }
+
+  return result
+}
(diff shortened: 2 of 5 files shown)


borisno2 (commented):
## Code Review

### Overview

Fixes #628: a virtual field named in a caller-supplied `include` (e.g. `context.db.myList.findMany({ include: { myVirtualField: true } })`) used to be forwarded straight to Prisma and throw `Unknown field` at runtime, since virtual fields have no database column. This PR adds `stripVirtualFieldsFromInclude()` (in `access-filter.ts`) and applies it as the final step before `model.findFirst`/`model.findMany` in `context/index.ts`, covering all three paths that can produce `include` (fragment-built, access-controlled merge, sudo passthrough).

### Correctness

- **Core logic is sound.** The strip check (`fieldConfig?.virtual`) runs before the relationship check, so a virtual field is dropped regardless of whether its value is bare `true` or an object — matches all the ways a caller could name it.
- **Nested relation includes preserve non-`include` keys correctly.** The recursive branch spreads the *original* `value` (not the narrowed `entry` from `asEntryObject`) and only overrides `include`, so `where`/`orderBy`/`take`/`skip`/`select` on a relation entry survive untouched.
- **`findFirst` is covered for free** since it delegates to the fixed `findMany`. `createGet` (singleton) doesn't accept a caller `include` and its auto-built include only ever contains relationship keys, so it's correctly left untouched.
- **The sudo passthrough fix is a real, distinct improvement** — previously sudo used `args.include` completely unfiltered, so this bug applied there too. Good that it's covered by both the fix and a dedicated test.

### Test coverage

Five new regression tests in `context.test.ts` under `virtual fields named in include no longer reach Prisma (#628)`:
- `findMany`/`findUnique` with a virtual field + real relationship in `include` → virtual key stripped, relationship include intact, virtual value still returned.
- Virtual value populated when omitted from `include` (pre-existing behavior, now guarded against regression).
- Sudo path strips the virtual key while leaving the real relationship passthrough unfiltered (matches documented sudo semantics).
- Read access control on a virtual field still applies.

This satisfies all the acceptance criteria from the issue's triage brief. Full suite (729 tests), `tsc` build, and lint all pass with no regressions.

### Minor/non-blocking observations

1. **Unresolvable related config skips nested stripping.** If a declared relationship's `ref` can't be resolved via `getRelatedListConfig` (unlikely — config validation should prevent a dangling ref), the nested `include` under that relation is passed through without recursing. This mirrors an existing fallback pattern elsewhere in the codebase (e.g. `filterReadableFields`'s "Related config not found, include the value as-is" branch), so it's consistent rather than a new gap.
2. **Nested `select` inside a relation include entry** (Prisma also supports `{ include: { posts: { select: {...} } } }`) isn't scanned for virtual keys. Out of scope per the issue — worth a follow-up only if that usage becomes common.

### Security

No concerns — this only narrows what reaches Prisma; it doesn't loosen any access control path. Field-level read access on virtual fields and relationship-level access filtering both remain intact and tested.

### Verdict

Solid, minimally-scoped fix with good test coverage matching the issue's acceptance criteria. No blocking issues found. Approving.

---

PR #635, opened 2026-07-10
feat(core): add label seam (getLabelFieldName / getItemLabel + ui.labelField)

### LINKED ISSUE #629: Label seam in stack-core (getLabelFieldName / getItemLabel + ui.labelField)
## What to build

A single **label seam** in `stack-core` that answers "what one field represents a row, and what text is its label" — used everywhere the admin UI shows a row as a reference (relationship cells, dropdown options, page headings).

Two entry points sharing one resolver, so the field chosen for *projection* can never drift from the field used for *rendering*:

- `getLabelFieldName(listConfig): string` — resolves `ui.labelField ?? 'name' ?? 'title' ?? 'id'` (first field that exists on the list).
- `getItemLabel(listConfig, item): string` — reads the resolved field off the row, falls back to `id` when the field is absent (e.g. stripped by field-level access).

Add `ui.labelField?: string` to the list config type, exported both functions from the `@opensaas/stack-core` root entry point. This is the **Label field** / **Item label** vocabulary now recorded in `CONTEXT.md`.

No consumer changes in this slice — this is the foundation two downstream slices build on.

## Acceptance criteria

- [ ] `ui.labelField` accepted on the list config and validated (references a declared scalar field when set)
- [ ] `getLabelFieldName` and `getItemLabel` exported from the root entry point
- [ ] Fallback order pinned: configured `labelField` → `name` → `title` → `id`
- [ ] `getItemLabel` falls back to `id` when the resolved field is missing from the row
- [ ] Unit tests table over `(ui.labelField, fields present, row)` asserting the projection⇄render agreement (the field `getLabelFieldName` returns is the field `getItemLabel` reads)
- [ ] Changeset added (minor: new config option + exports)

## Blocked by

None - can start immediately

borisno2:
## Summary

- Add `ui.labelField?: string` to `ListUIConfig` — the field used to represent a row as a single label (relationship cells, dropdown options, page headings)
- Add `getLabelFieldName(listConfig)`, which resolves `ui.labelField ?? 'name' ?? 'title' ?? 'id'` (first field that exists on the list), throwing if a configured `labelField` doesn't reference a declared, non-relationship field
- Add `getItemLabel(listConfig, item)`, which reads the field `getLabelFieldName` resolves off a row, falling back to `id` when that field is missing (e.g. stripped by field-level access)
- Both share one resolver so the field chosen for projection can never drift from the field used for rendering
- Export both from the `@opensaas/stack-core` root entry point
- Record the **Label field** / **Item label** vocabulary in `CONTEXT.md`
- No consumer changes in this slice — this is the foundation for downstream admin UI work

## Test plan

- [x] Unit tests table over `(ui.labelField, fields present, row)`, asserting the field `getLabelFieldName` returns is the field `getItemLabel` reads
- [x] `pnpm test` (packages/core) — all 744 tests pass
- [x] `pnpm build` (packages/core) — type-checks clean
- [x] `pnpm format` / `pnpm manypkg fix`
- [x] Changeset added (minor: new config option + exports)

Closes #629

---
_Generated by [Claude Code](https://claude.ai/code/session_01YaVyutPdHbv2Qp2wFa4bBj)_

### DIFF
--- a/.changeset/tame-otters-listen.md
+++ b/.changeset/tame-otters-listen.md
@@ -1,0 +1,17 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Add a label seam for the admin UI: `getLabelFieldName(listConfig)` resolves the field that represents a list's rows as a single label (`ui.labelField` → `name` → `title` → `id`), and `getItemLabel(listConfig, item)` reads that field off a row, falling back to `id` when it's missing. Both are exported from the root entry point.
+
+```typescript
+import { getLabelFieldName, getItemLabel } from '@opensaas/stack-core'
+
+Post: list({
+  fields: { title: text() },
+  ui: { labelField: 'title' },
+})
+
+getLabelFieldName(listConfig) // 'title'
+getItemLabel(listConfig, item) // item.title, or item.id if title is missing
+```
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -81,3 +81,13 @@ _Avoid_: storage backend, storage adapter, uploader
 **Provider registry**:
 The lookup from a configured provider `type` to the constructor that builds it, which the storage runtime consults instead of a closed `switch`. A host registers the (optional) provider packages it uses, so non-`local` and custom providers are constructable without the runtime depending on every provider's SDK.
 _Avoid_: provider switch, provider map, plugin registry
+
+### Admin UI
+
+**Label field**:
+The field a list uses to represent its rows as a single value — resolved by `getLabelFieldName` from `ui.labelField` (when configured and pointing at a declared, non-relationship field) or the fallback order `name` → `title` → `id`. The single source of truth so the field chosen for projection can never drift from the field used for rendering.
+_Avoid_: display field, title field, name field (the fallback happens to check a field called `name`, but the concept isn't tied to that key)
+
+**Item label**:
+The rendered text for one row, produced by `getItemLabel` reading the Label field off that row and falling back to `id` when the field is missing (e.g. stripped by field-level access). Used anywhere the admin UI shows a row as a reference — relationship cells, dropdown options, page headings.
+_Avoid_: display value, row label
--- a/packages/core/src/config/label.ts
+++ b/packages/core/src/config/label.ts
@@ -1,0 +1,58 @@
+import type { ListConfig } from './types.js'
+
+/**
+ * Resolve the field name that represents a list's rows as a single label —
+ * the projection half of the label seam. `getItemLabel` reads exactly the
+ * field this returns, so the field chosen for projection can never drift
+ * from the field used for rendering.
+ *
+ * Fallback order: configured `ui.labelField` → `name` → `title` → `id`
+ * (first field that exists on the list).
+ *
+ * @throws if `ui.labelField` is set but does not reference a declared,
+ * non-relationship field on the list.
+ */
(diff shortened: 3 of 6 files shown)


borisno2:
## Review: Label seam (`getLabelFieldName` / `getItemLabel` + `ui.labelField`)

### Overview
Adds a single label-resolution seam to `stack-core`: `getLabelFieldName(listConfig)` resolves which field represents a row (`ui.labelField` → `name` → `title` → `id`), and `getItemLabel(listConfig, item)` reads that same field off a row, falling back to `id` when it's absent. Both are exported from the root entry point, `ui.labelField` is added to `ListUIConfig`, and the "Label field" / "Item label" vocabulary is recorded in `CONTEXT.md`. Pure addition — no consumer wiring in this slice, matching the issue's stated scope.

### Code quality & style
- Implementation is small, well-contained, and matches the codebase's `ListConfig<any>` + `eslint-disable-next-line @typescript-eslint/no-explicit-any` convention used throughout `context/index.ts` / `write-pipeline.ts`.
- `getItemLabel` correctly delegates to `getLabelFieldName` rather than re-deriving the fallback order, which is exactly the guarantee the issue asks for (projection and render can't drift).
- `Object.prototype.hasOwnProperty.call(item, fieldName)` is the right way to distinguish "field stripped from row" from "field present but falsy" — more precise than an `in` check or truthiness check.
- CONTEXT.md entry follows the existing term/definition/`_Avoid_` format used by every other entry in the file.
- JSDoc on `getLabelFieldName`/`getItemLabel` and the `ui.labelField` type is clear and includes a usage example, consistent with the rest of `types.ts`.

### Test coverage
- Good table-driven coverage: fallback order (labelField → name → title → id), both throw paths (missing field, relationship field), null vs. missing field-on-row, and a dedicated `it.each` asserting the projection⇄render agreement across 5 scenarios — this directly satisfies the issue's acceptance criterion for that test shape.
- All 744 tests pass; `pnpm build`, `pnpm lint`, `pnpm format` are clean.

### Potential issues / minor suggestions (non-blocking)
1. **System fields (`id`, `createdAt`, `updatedAt`) can't be explicitly set as `ui.labelField`.** Since these are auto-injected by the generator rather than present in `listConfig.fields`, `ui: { labelField: 'createdAt' }` would throw "does not reference a field declared on this list" even though `createdAt` is a perfectly valid, always-present scalar on every row. `id` is already the terminal fallback so that specific case is low-impact, but `createdAt`/`updatedAt` are plausible label choices (e.g. "show record by creation date") that this rejects. Worth a follow-up if that's a use case downstream slices care about — not a blocker for this foundational slice.
2. **Per-call resolution in `getItemLabel`.** Each call re-derives `getLabelFieldName` (an `in` check plus optional throw), which is cheap in isolation but will run once per row when list views wire this up later. Not an issue at this slice's scope (no consumers yet), just worth keeping in mind if a future list-view render loop calls it per-row per-render — memoizing the resolved field name per list would be a trivial follow-up if it shows up in profiling.
3. **Minor:** the error thrown when `ui.labelField` is misconfigured is a plain `Error`, matching the existing `plugin-engine.ts` convention (no custom error class in this codebase for config-shape errors), so this is consistent rather than a gap.

### Security
No security-relevant surface here — pure config resolution over already-trusted config/row data, no user input parsing, no injection surface.

### Verdict
Solid, narrowly-scoped foundational change. No correctness bugs found; the two notes above are forward-looking observations for when this seam gets wired into the admin UI, not defects in this PR.


---

PR #602, opened 2026-06-27
Release: Validation &amp; field-type input parity (zod 4.4 + calendarDay) — PRD #582

borisno2:
Delivers PRD #582 — makes field input validation and generated input types correct and consistent so normal partial-update idioms work and `calendarDay` no longer contradicts its own validation/storage.

## Included changes

- **#570** (`@opensaas/stack-core`, patch) — Update-input validation rejected omitted required fields under zod 4.4. Replaced the required-on-update `z.union([schema, z.undefined()])` shape with `.optional()` across the `text`, `calendarDay`, `password`, and `json` builders, so partial updates validate only the keys present. No new nullability introduced; create-time requiredness unchanged.
- **#571** (`@opensaas/stack-core`, minor) — `calendarDay` is now a `YYYY-MM-DD` string end-to-end: `getTypeScriptType()` returns `string`, string-only zod validation retained, and a timezone-safe `resolveOutput` normalises reads to `YYYY-MM-DD`. Storage stays `DateTime @db.Date`.

## ⚠️ Behavioral change (reads)

Reading a `calendarDay` now returns a `YYYY-MM-DD` **string** instead of a `Date`. Flagged in the `@opensaas/stack-core` minor changeset; consumers relying on a `Date` on read should switch to the string form.

## Verification

QA acceptance pass on the integration branch (HEAD `618ab09`): build clean, `@opensaas/stack-core` 665/665 tests, `@opensaas/stack-cli` 308/308 tests. Both changesets present (`quiet-otters-fix.md` patch, `calm-otters-glide.md` minor). QA sign-off: #582.

## Known limitation / follow-ups

- **#599** — At the `context.db.*.create()/update()` call site a `Date` is rejected at **runtime** (zod `ValidationError`), but not at compile time, because that `data` type derives from Prisma's `Date | string` for a `@db.Date` column. Genuine compile-time narrowing is a cross-cutting generator change tracked there.
- **#597** — `json()` required-on-create is not enforced (bare `z.unknown()`); pre-existing, surfaced during this delivery.

Closes #582 (already closed). Implements #570, #571.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017JiVKJwhQbG4AfTy65auS4

---
_Generated by [Claude Code](https://claude.ai/code/session_017JiVKJwhQbG4AfTy65auS4)_

PR #637, opened 2026-07-10
fix(core): coerce calendarDay writes to Date before Prisma 7 validation

### LINKED ISSUE #621: calendarDay write to Prisma 7 @db.Date still 500s: #602 added the string type + read normaliser but no write-side string→Date coercion (#571 regression)
## Summary

Follow-up to #571 (closed COMPLETED via #602). #602 made `calendarDay` a `YYYY-MM-DD` string end-to-end at the `context.db` boundary and added a `resolveOutput` read normaliser — but it did **not** add a write-side transform. On Prisma 7 the documented write API still fails before the DB.

The changelog for #602 states:

> Writes: pass YYYY-MM-DD strings, not Date objects
> `await context.db.event.create({ data: { startDate: '2025-01-15' } })`

But Prisma 7's client validator rejects a bare `YYYY-MM-DD` string for a `DateTime @db.Date` / `@postgresql.Date` column:

```
PrismaClientValidationError:
Invalid value for argument `startDate`: premature end of input. Expected ISO-8601 DateTime.
```

So the documented string write path 500s, and (per #571) a `Date` is rejected by the field's zod validation — neither representation round-trips.

## Repro

- `@opensaas/stack-core@0.26.0`, Prisma `7.8.0`, Postgres (via `@prisma/adapter-neon` or `@prisma/adapter-pg`).
- A list with `startDate: calendarDay({ validation: { isRequired: true } })` → generated column `DateTime @db.Date`.
- `await context.db.event.create({ data: { startDate: '2025-10-10', ... } })`.

Result: `PrismaClientValidationError: Invalid value for argument 'startDate': premature end of input. Expected ISO-8601 DateTime.` thrown from the write pipeline before the DB write. Passing `new Date('2025-10-10')` instead is rejected by the field's zod schema (string-regex only).

## Root cause

The 0.26.0 `calendarDay()` field has `hooks.resolveOutput` (read) but no `hooks.resolveInput` (write). The zod schema validates the `YYYY-MM-DD` string and passes it through unchanged, so Prisma receives a bare date string and rejects it. #571 explicitly asked for `z.union([z.date(), z.string().regex(...)])` **with a normalising transform**; the union/transform on the write side appears to have been dropped when #602 reworked the type to string-only.

## Suggested fix

Re-add a field-level write transform on `calendarDay`, runs before field validation (so the string→Date conversion is in place when zod validates), and widen the schema to accept either shape:

```js
hooks: {
  resolveInput: ({ resolvedData, fieldKey }) => {
    const v = resolvedData?.[fieldKey]
    if (v == null || v instanceof Date) return v
    if (typeof v === 'string' && /^\d{4}-\d{2}-\d{2}$/.test(v)) return new Date(`${v}T00:00:00.000Z`)
    return v
  },
  resolveOutput: ({ value }) => formatCalendarDay(value),
},
// and: const dateSchema = z.union([stringSchema, z.date()])
```

(UTC midnight keeps `@db.Date` date-only with no TZ drift.) The hook should read `resolvedData[fieldKey]`, not raw `inputData`, so a LIST-level `resolveInput` that INJECTS a calendarDay default for an omitted key survives (otherwise the field hook reads `undefined` and overwrites the injected default with NULL).

Happy to open a PR if useful.

borisno2:
## Summary

- `calendarDay()` gained a field-level `resolveInput` hook that converts a valid `YYYY-MM-DD` string to a UTC-midnight `Date` before the write reaches Prisma. Prisma 7's client validator rejects a bare date string for a `DateTime @db.Date` column, so the documented string-write path was 500ing (#621, a regression from #602).
- The hook reads `resolvedData[fieldKey]` (not raw `inputData`), so a list-level `resolveInput` that injects a default for an omitted key still gets coerced instead of being overwritten with `null`.
- Malformed strings, `null`/`undefined`, and already-`Date` values pass through the hook unchanged so existing validation (isRequired, format errors) is unaffected.
- Widened the field's zod schema to `z.union([stringSchema, z.date()])` since a successful write now reaches validation as a `Date` (resolveInput runs before validation in the write pipeline).
- Reads are unchanged — `resolveOutput` still normalises back to a `YYYY-MM-DD` string.

## Test plan

- [x] Unit tests for the `resolveInput` hook (string→Date coercion, pass-through cases, reading `resolvedData` vs `inputData`)
- [x] Unit tests for the widened zod schema (string, Date, malformed string)
- [x] End-to-end tests via `context.db.event.create`/`update` against a mocked Prisma client asserting the value forwarded to Prisma is a `Date`, reproducing the issue's repro scenario
- [x] `pnpm test` (packages/core) — 752 passed
- [x] `tsc --noEmit`

Closes #621

---
_Generated by [Claude Code](https://claude.ai/code/session_01BZwNZE4vZD89KxeHsLb414)_

### DIFF
--- a/.changeset/tidy-plums-argue.md
+++ b/.changeset/tidy-plums-argue.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix `calendarDay` writes 500ing on Prisma 7 `@db.Date` columns — a `resolveInput` hook now coerces a `YYYY-MM-DD` string to a UTC-midnight `Date` before validation, and the field's zod schema accepts either shape.
--- a/packages/core/src/fields/calendar-day.test.ts
+++ b/packages/core/src/fields/calendar-day.test.ts
@@ -1,7 +1,9 @@
-import { describe, it, expect, expectTypeOf } from 'vitest'
+import { describe, it, expect, expectTypeOf, vi } from 'vitest'
 import { calendarDay } from './index.js'
 import { generateZodSchema, validateWithZod } from '../validation/schema.js'
+import { getContext } from '../context/index.js'
 import type { FieldConfig } from '../config/types.js'
+import type { OpenSaasConfig } from '../config/types.js'
 
 /**
  * calendarDay is a YYYY-MM-DD string end-to-end (Keystone's CalendarDay
@@ -54,7 +56,7 @@ describe('calendarDay field (YYYY-MM-DD string end-to-end)', () => {
     })
   })
 
-  describe('write validation (string-only)', () => {
+  describe('write validation (YYYY-MM-DD string, or a Date post-resolveInput)', () => {
     const fields: Record<string, FieldConfig> = {
       startsOn: calendarDay({ validation: { isRequired: true } }),
     }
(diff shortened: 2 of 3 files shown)


borisno2:
## Review

### Overview
Fixes #621: `calendarDay` writes were 500ing on Prisma 7 because the documented `YYYY-MM-DD` string write path reached the client's `@db.Date` validator unconverted. This PR adds a field-level `resolveInput` hook that coerces a valid `YYYY-MM-DD` string to a UTC-midnight `Date` before the write pipeline's validation phase, and widens `calendarDay`'s zod schema to `z.union([stringSchema, z.date()])` to accept the post-coercion shape.

### Correctness
- Hook ordering is right: `hook-pipeline.ts` runs field `resolveInput` (phase 1.5) strictly before `validateFieldRules`/zod (phase 3), so a valid string is a `Date` by the time the schema checks it. Confirmed by tracing `context/hook-pipeline.ts`.
- Reads `resolvedData[fieldKey]` rather than `inputData`, so a list-level `resolveInput` that injects a default for an omitted key is honored (loop in `executeFieldResolveInputHooks` only runs for keys present in `resolvedData`, which already reflects that injected default). Matches the issue's explicit ask and is covered by a test.
- `UTC` midnight construction (`${value}T00:00:00.000Z`) avoids the day-drift footgun the read-side `formatCalendarDay` already guards against.
- Malformed strings, `null`/`undefined`, and existing `Date` instances pass through the hook unchanged, so `isRequired` and format-error behavior is preserved.

### Minor observation (non-blocking)
For a value that's neither a valid `YYYY-MM-DD` string nor a `Date` (e.g. a stray number/object reaching `context.db` from an untyped JS caller), the error message quality regresses slightly. Previously the schema was a plain `z.string()` with a custom "must be a valid date in ISO8601 format" message; now `z.union([stringSchema, z.date()])` surfaces zod's generic `"Invalid input"` top-level message instead for non-string/non-date values (verified locally — a number input flips from the custom message to `"Invalid input"`). This is an edge case (the field's TS type is `string`, so a typed caller can't reach it), and it's not covered by a test either before or after this change, so it's not a regression this PR needs to fix — just worth a follow-up if better DX for that edge case matters.

### Test coverage
Strong — new tests cover the `resolveInput` hook directly (string→Date, Date/null/undefined pass-through, malformed-string pass-through, `resolvedData` vs `inputData`), the widened zod schema, and — notably — an end-to-end repro via `context.db.event.create/update` against a mocked Prisma client that asserts the exact value forwarded to the client is a `Date`, which is the most direct verification of the original bug. Read-side normalization back to `YYYY-MM-DD` is also re-verified end-to-end.

### Style / conventions
- Follows the existing field-self-containment pattern (no switch statements added elsewhere).
- Comments explain the non-obvious "why" (hook ordering constraint, UTC-midnight rationale) rather than restating code.
- Changeset present and correctly scoped as `patch`.

### Risk
Low. Change is scoped to a single field builder, additive to the hook object (still overridable by user-provided hooks per the existing merge pattern), and read behavior is untouched. All 752 core tests + typecheck + lint pass.

No blocking issues found.

---

PR #639, opened 2026-07-10
List page: relationship cell labels via the shared label seam

### LINKED ISSUE #632: List page: relationship cell labels via the shared label seam
## What to build

Route the list-page relationship cell label through the shared **label seam** so there is one label rule, not two drifted copies.

Today the list page derives a relationship cell's text with an inline `name → title → label → id` chain in the UI utils, while the item form used a different `name → title → id` chain — the two had already drifted. Replace the list-page inline rule with `getItemLabel` (from #629), computing the label for relationship columns so the cell renders through the same resolver as everywhere else.

Scope note: this unifies the **label rule** only. It does **not** re-project the list read — the list page's own depth-5 auto-include over-fetch is a separate, deferred piece of work (the read-projection seam) and is explicitly out of scope here.

## Acceptance criteria

- [ ] List-page relationship cells render their label via `getItemLabel`
- [ ] The inline `name → title → label → id` derivation is removed from the UI utils
- [ ] Cells honour a list's configured `ui.labelField`
- [ ] Existing list-view behaviour (empty value shows a dash, many-relationships join) preserved
- [ ] Changeset added (ui patch/minor)

## Blocked by

- #629

borisno2:
## Summary

- `ListView` (the built-in admin list page's server component) now resolves each relationship cell value to `{ id, label }` via the shared label seam (`getItemLabel`) before it crosses the server/client boundary to `ListViewClient` — honouring a related list's configured `ui.labelField`, and matching the same resolver already used for relationship-option labels in the item edit form (`relationship-options.ts`).
- Removed the drifted inline `name → title → label → id` derivation (`getRelationshipDisplayValue`) from `packages/ui/src/lib/utils.ts`. `getFieldDisplayValue` no longer special-cases `'relationship'` — that display logic now depends on a related list's config, which only the server component has.
- The standalone `ListTable` component (embedded directly by consumers with raw items and no list config) keeps its own local, unexported fallback for relationship display, since it never had access to any list config to resolve `ui.labelField` against. This keeps it working without depending on the now-removed shared chain in `utils.ts`.
- Existing list-view behaviour is preserved: empty relationship shows a dash, many-relationships render as comma-separated links.

## Test plan

- [x] Added `packages/ui/tests/components/ListView.test.tsx` covering: default `name` fallback, a configured `ui.labelField`, many-relationships resolving to an array of `{ id, label }`, falling back to `id` when the label field is stripped (e.g. by access control), and `null`/empty relationships resolving to `null`.
- [x] Updated `packages/ui/tests/components/ListViewClient.test.tsx` relationship fixtures to the new pre-resolved `{ id, label }` shape.
- [x] `pnpm test` in `packages/ui` — 215 tests passing.
- [x] `pnpm build` in `packages/ui` — typechecks cleanly.
- [x] `pnpm lint` — no new warnings/errors.
- [x] Changeset added (`@opensaas/stack-ui` patch).

Closes #632

---
_Generated by [Claude Code](https://claude.ai/code/session_01KEn73hWY1dK939k4eVZ1e3)_

### DIFF
--- a/.changeset/curly-donkeys-relax.md
+++ b/.changeset/curly-donkeys-relax.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+List-page relationship cells now render their label via the shared label seam (`getItemLabel`), honouring a related list's `ui.labelField` instead of an inline `name → title → label → id` guess that had drifted from the item form.
--- a/packages/ui/src/components/ListView.tsx
+++ b/packages/ui/src/components/ListView.tsx
@@ -1,7 +1,29 @@
 import Link from 'next/link.js'
 import { ListViewClient } from './ListViewClient.js'
 import { formatListName } from '../lib/utils.js'
-import { type AccessContext, getDbKey, getUrlKey, OpenSaasConfig } from '@opensaas/stack-core'
+import {
+  type AccessContext,
+  getDbKey,
+  getItemLabel,
+  getUrlKey,
+  OpenSaasConfig,
+} from '@opensaas/stack-core'
+
+/**
+ * Resolve a fetched relationship value (the full related record, or `null`)
+ * into the `{ id, label }` shape `ListViewClient` renders — computing the
+ * label via the shared label seam (`getItemLabel`) so list-page cells never
+ * drift from the item form's relationship-option labels.
+ */
+function toRelationshipLabel(
+  value: unknown,
+  relatedListConfig: OpenSaasConfig['lists'][string] | undefined,
+): { id: string; label: string } | null {
+  if (!value || typeof value !== 'object' || !relatedListConfig) return null
+  const row = value as Record<string, unknown>
+  if (!('id' in row)) return null
+  return { id: String(row.id), label: getItemLabel(relatedListConfig, row) }
+}
 
 /**
  * Default sort for the list table, mirroring Keystone's `ui.listView.initialSort`.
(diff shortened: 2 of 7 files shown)


borisno2:
## Self-review

Reviewed the diff for correctness and scope against the acceptance criteria in #632.

**Findings:** none blocking.

**Design notes for reviewers:**

- `ListView.tsx` (server component, has full `OpenSaasConfig`) now resolves each relationship value to `{ id, label }` via `getItemLabel(relatedListConfig, row)` *before* the props cross into the `'use client'` `ListViewClient` — a real `ListConfig` object can't be passed as a client-component prop (it carries functions), so resolution has to happen server-side, mirroring the existing pattern in `packages/core/src/query/relationship-options.ts`.
- The drifted inline `name → title → label → id` chain (`getRelationshipDisplayValue`) is removed from the shared `packages/ui/src/lib/utils.ts`, per the acceptance criteria.
- The standalone `ListTable` component (embedded directly by consumers with raw items and no list config at all) can't call `getItemLabel` — it has no `ListConfig` to resolve against. It keeps a local, unexported copy of the old fallback chain scoped to that one file, so it can't drift against anything else. This is intentionally out of scope of #632 (which is about the *list page*, i.e. `ListView`/`ListViewClient`), and its existing tests/behavior are unchanged.
- Edge case worth flagging: if a relationship `ref` points at a list key missing from `config.lists` (a config error), the cell now silently renders a dash rather than showing raw data — a defensive fallback, not a behavior a correctly configured app would hit.

**Verification:**
- `pnpm test` (packages/ui): 215/215 passing, including new coverage in `ListView.test.tsx` for the default `name` fallback, a configured `ui.labelField`, many-relationships, missing-label-field fallback to `id`, and null/empty relationships.
- `pnpm build` (packages/ui): typechecks cleanly.
- `pnpm lint` / `pnpm format` / `pnpm manypkg fix`: no new warnings, no diffs.


---

PR #689, opened 2026-07-15
fix(auth): route plugin schema extensions through the model-key remap

### LINKED ISSUE #677: Better-auth plugin schema extensions ignore the auth model-key remap and can collide with a same-named host list
## What to build

When the auth plugin adopts pre-existing **Auth lists**, it can remap the base better-auth `user` model onto a different list key (e.g. `AuthUser`) so it does not collide with an application's own domain list that happens to be keyed `User`. Separately, a better-auth provider plugin can ship its own `schema` that extends the base `user` model with extra fields (e.g. an anonymous-session flag).

Today the step that converts a provider plugin's `schema` into a list extension derives the target list key straight from the plugin's declared table name (pascal-cased), **without consulting the configured model-key remap**. So an extension targeting `user` resolves to the list key `User`. When the host application has its own unrelated `User` list, the extension is applied to *that* list: it injects the plugin's field(s) into it, and — because the extension also carries the plugin's permissive operation-level access defaults — it overwrites that list's existing access rules (e.g. an admin-only update silently becomes "anyone can update themselves").

Fix the key resolution so that a plugin schema extension targeting a base auth model is converted against the **same remapped list key** the Auth lists themselves use, so it lands on the adopted Auth list and never on a same-named but unrelated host list.

This is the same class of "silently extend an app's own `User`" blocker that ADR-0007 was written to prevent, resurfacing through the plugin-schema code path. A broader, complementary guardrail — that a list-extension should not silently overwrite an existing list's access at all — is tracked as a separate issue.

**Where:** the plugin-schema loop in `packages/auth/src/config/plugin.ts` calls `convertBetterAuthSchema(pluginSchema)`; key derivation is `tableSchema.modelName || toPascalCase(tableName)` in `packages/auth/src/server/schema-converter.ts`, which never receives the normalized model remap.

## Acceptance criteria

- [ ] With the base user model remapped (e.g. to `AuthUser`), a provider plugin whose `schema` extends `user` applies its field(s) to the adopted Auth list, not to a re-derived `User` key.
- [ ] An unrelated host list keyed `User` is left untouched by such a plugin extension — no injected column, no change to its operation-level access.
- [ ] Key resolution reads the actual configured model-key remap rather than re-deriving from the plugin's own table name.
- [ ] Regression test: remapped user model + an unrelated host `User` list + a provider plugin extending `user`; assert the extension lands on the Auth list and the host `User` keeps its own fields and access.

## Blocked by

borisno2:
## Summary

- A better-auth provider plugin's `schema` extension of a base model (e.g. `user`) was converted to an OpenSaaS list key by re-deriving it from the plugin's own table name (`tableSchema.modelName || toPascalCase(tableName)`), ignoring the auth plugin's configured model-key remap (e.g. `user: { modelName: 'AuthUser' }`).
- This meant that when the base `user` model was remapped away from `User` (to avoid colliding with an app's own domain `User` list — the ADR-0007 pattern), a plugin schema extension of `user` still resolved to `User` and got applied to the unrelated host list, injecting the plugin's field(s) and overwriting that list's access control with the plugin's permissive defaults.
- `convertBetterAuthSchema()` (`packages/auth/src/server/schema-converter.ts`) now accepts an optional `baseModelKeys` map (the same remapped list keys `deriveAuthLists` computes for `user`/`session`/`account`/`verification`) and prefers that remap over the plugin's own `modelName`/table name for those four base models. Other plugin-declared tables (e.g. `oauthApplication`) are unaffected.
- `authPlugin()` (`packages/auth/src/config/plugin.ts`) now builds this remap from `normalized.models` and passes it into `convertBetterAuthSchema`.

## Test plan

- [x] Added a regression test (`packages/auth/tests/plugin-derived-keys.test.ts`) covering the exact acceptance-criteria scenario: a remapped `user` model (`AuthUser`) + an unrelated host `User` list + a `betterAuthPlugins` entry with a `schema` extending `user` — asserts the extension's field lands on `AuthUser` and the host `User` keeps its own fields and access untouched.
- [x] Added unit tests in `packages/auth/tests/schema-converter.test.ts` for `convertBetterAuthSchema`'s new `baseModelKeys` parameter (remap wins over table name, remap wins over an explicit `tableSchema.modelName`, non-base tables unaffected).
- [x] `pnpm --filter @opensaas/stack-auth test` — 137 passed
- [x] `pnpm --filter @opensaas/stack-auth build`
- [x] `pnpm lint`

Closes #677

---
_Generated by [Claude Code](https://claude.ai/code/session_01WHRXKviN2SEXgbNUm1YyJ4)_

### DIFF
--- a/.changeset/gentle-otters-listen.md
+++ b/.changeset/gentle-otters-listen.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-auth': patch
+---
+
+Fix better-auth plugin schema extensions (e.g. `user`) resolving against a re-derived list key instead of the configured model-key remap, which could silently apply the extension's fields/access to an unrelated host list sharing the default key.
--- a/packages/auth/src/config/plugin.ts
+++ b/packages/auth/src/config/plugin.ts
@@ -46,12 +46,24 @@ export function authPlugin(config: AuthConfig): Plugin {
       // match the developer's live better-auth tables.
       const authLists = getAuthLists(normalized.extendUserList, normalized.models)
 
+      // The same base-model list keys the Auth lists above were derived under
+      // (e.g. `user.modelName: 'AuthUser'`). A provider plugin's schema
+      // extension of a base model (e.g. `user`) must resolve against this
+      // remap too, so it lands on the adopted Auth list rather than a
+      // re-derived key that can collide with an unrelated host list.
+      const baseModelKeys = {
+        user: normalized.models.user.modelName,
+        session: normalized.models.session.modelName,
+        account: normalized.models.account.modelName,
+        verification: normalized.models.verification.modelName,
+      }
+
       // Extract additional lists from Better Auth plugins
       for (const plugin of normalized.betterAuthPlugins) {
         if (plugin && typeof plugin === 'object' && 'schema' in plugin) {
           // Plugin has schema property - convert to OpenSaaS lists
           const pluginSchema = plugin.schema
-          const pluginLists = convertBetterAuthSchema(pluginSchema)
+          const pluginLists = convertBetterAuthSchema(pluginSchema, baseModelKeys)
 
           // Add or extend lists from plugin
           for (const [listName, listConfig] of Object.entries(pluginLists)) {
--- a/packages/auth/src/server/schema-converter.ts
+++ b/packages/auth/src/server/schema-converter.ts
@@ -282,18 +282,47 @@ export function convertTableToList(
 }
 
 /**
+ * Base better-auth model keys — the four tables the auth plugin's own model
+ * config (`user`/`session`/`account`/`verification`) can remap via `modelName`.
+ */
+type BaseAuthModelKey = 'user' | 'session' | 'account' | 'verification'
+
+/**
+ * Resolved list key for each base better-auth model, as derived by
+ * {@link deriveAuthLists} from the plugin's configured model-key remap (e.g.
+ * `user.modelName: 'AuthUser'`). Passed to {@link convertBetterAuthSchema} so a
+ * provider plugin's schema extension of a base model (e.g. `user`) lands on the
+ * same adopted Auth list rather than being re-derived from the plugin's own
+ * table name.
+ */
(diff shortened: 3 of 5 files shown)


borisno2:
## Self-review

**What it does:** `convertBetterAuthSchema()` derived a plugin schema extension's target list key from the plugin's own table name (`tableSchema.modelName || toPascalCase(tableName)`), ignoring `authPlugin`'s configured `modelName` remap for the base `user`/`session`/`account`/`verification` models. `authPlugin()` now passes the resolved base-model keys through, so an extension of e.g. `user` always lands on the adopted Auth list (`AuthUser`) instead of a re-derived `User` that could collide with an unrelated host list.

**Correctness:**
- ✅ Only the four base better-auth table names are affected by the remap; other plugin-declared tables (e.g. `oauthApplication`) fall through to the previous `modelName`/`toPascalCase` behavior unchanged.
- ✅ Add-vs-extend ordering still works: the plugin-schema loop runs before the Auth-lists loop in `plugin.ts`, so whichever runs first `addList`s and the second `extendList`s — merging plugin fields into the Auth list either way.
- ✅ Backward compatible: `convertBetterAuthSchema(tables)` (no second arg) behaves exactly as before since `baseModelKeys` defaults to `{}`.

**Fixed during self-review:** the initial implementation indexed `baseModelKeys[tableName.toLowerCase() as BaseAuthModelKey]`, a type assertion. Per this repo's CLAUDE.md ("do not use type casting"), replaced it with an explicit `switch` (`resolveBaseModelKey`) so no cast is needed — pushed in c6004c0.

**Test coverage:** regression test in `plugin-derived-keys.test.ts` covers the exact acceptance-criteria scenario (remapped `AuthUser` + unrelated host `User` + a `betterAuthPlugins` schema extending `user`), plus unit tests in `schema-converter.test.ts` for the new `baseModelKeys` parameter in isolation.

`pnpm --filter @opensaas/stack-auth test` (137 passed), `build`, `pnpm lint`, and `pnpm format` all pass clean.


---

PR #690, opened 2026-07-15
fix(core): relationship live-search 500 when target list's label field is virtual

### LINKED ISSUE #676: Relationship live-search 500s when the target list's label field is a virtual field
## What to build

The admin UI relationship-picker live-search endpoint orders results by the target list's **Label field** unconditionally. When that Label field is a `virtual()` field (computed at read time via `resolveOutput`, with no backing database column), the underlying Prisma query fails with a validation error and the request 500s. The picker swallows the error and degrades to "No results found", so a configuration that is valid everywhere else in the framework looks like an empty result set rather than a bug.

Make live-search work for **any** list regardless of whether its Label field is a stored column or a computed/virtual field. When the resolved Label field is not a real, orderable scalar column, fall back to a safe ordering (e.g. by `id`) instead of passing the virtual field name into `orderBy`. The search filter path already restricts itself to text label fields; ordering should follow the same "is this a real column?" guard so the two stay consistent.

**Where:** `getRelationshipOptions` in `packages/core/src/query/relationship-options.ts` builds `orderBy: { [labelField]: 'asc' }` without checking the field's type (the sibling `where` clause already guards on `type === 'text'`). Virtual fields are discriminable via `type: 'virtual'` / the `virtual: true` flag on the field config.

## Acceptance criteria

- [ ] Live-search against a list whose Label field is virtual returns matching rows (or all rows within the take limit) instead of 500ing.
- [ ] Ordering falls back to a stored scalar column (e.g. `id`) when the Label field is virtual/non-orderable; a plain `text()` label field still orders by that field.
- [ ] The relationship picker no longer surfaces a backend error as a misleading "No results found" for this case.
- [ ] Regression test covering a relationship whose target list uses a virtual Label field.

## Blocked by

borisno2:
## Summary

- `getRelationshipOptions` in `packages/core/src/query/relationship-options.ts` built `orderBy: { [labelField]: 'asc' }` without checking the label field's type. When the target list's Label field is a `virtual()` field (computed at read time via `resolveOutput`, with no backing database column), Prisma rejects the `orderBy` and the request 500s — which the relationship picker swallows and degrades into a misleading "No results found".
- Ordering now follows the same "is this a real column?" guard the sibling `where` clause already uses: it falls back to ordering by `id` (always a real, orderable column) whenever the Label field is virtual. A plain `text()` label field still orders by that field, and other stored scalars (e.g. `integer`) are unaffected.

## Acceptance criteria

- [x] Live-search against a list whose Label field is virtual returns matching rows instead of 500ing.
- [x] Ordering falls back to `id` when the Label field is virtual/non-orderable; a `text()` label field still orders by that field.
- [x] The relationship picker no longer surfaces a backend error as a misleading "No results found" for this case.
- [x] Regression test covering a relationship whose target list uses a virtual Label field.

## Test plan

- [x] Added regression test `falls back to ordering by id when the label field is virtual` in `relationship-options.test.ts`
- [x] `pnpm test` in `packages/core` — 770 passed
- [x] `pnpm lint` passes (no new errors)
- [x] Changeset added (`@opensaas/stack-core`: patch)

Closes #676

https://claude.ai/code/session_01KSUc7hGQE96ytnUxvQabxU

---
_Generated by [Claude Code](https://claude.ai/code/session_01KSUc7hGQE96ytnUxvQabxU)_

### DIFF
--- a/.changeset/quiet-otters-search.md
+++ b/.changeset/quiet-otters-search.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix relationship live-search 500 when the target list's label field is a virtual field by ordering by `id` instead of the non-orderable virtual column
--- a/packages/core/src/query/relationship-options.test.ts
+++ b/packages/core/src/query/relationship-options.test.ts
@@ -122,6 +122,20 @@ describe('getRelationshipOptions', () => {
     expect(result).toEqual([{ id: 'v1', label: 'Computed One' }])
   })
 
+  it('treats a field flagged virtual via type alone as non-orderable (orders by id)', async () => {
+    const config = makeConfig()
+    // Only `type: 'virtual'` is set here (no `virtual: true`) to lock in the
+    // discriminator against future refactors.
+    ;(config.lists.VirtualLabel.fields.displayName as { virtual?: boolean }).virtual = undefined
+    const delegate = makeDelegate([{ id: 'v1', displayName: 'Computed One' }])
+    const context = makeContext({ virtualLabel: delegate })
+
+    await getRelationshipOptions(context, config, 'VirtualLabel', { search: 'One' })
+
+    const call = delegate.findMany.mock.calls[0][0] as Record<string, unknown>
+    expect(call.orderBy).toEqual({ id: 'asc' })
+  })
+
   it('unions currently-selected ids even when beyond take / not matching search', async () => {
     // The bounded/search-scoped query only returns a1 (mimicking take:1 + search).
     const primaryDelegate = makeDelegate([authors[0]])
--- a/packages/core/src/query/relationship-options.test.ts
+++ b/packages/core/src/query/relationship-options.test.ts
@@ -39,6 +39,13 @@ function makeConfig(): OpenSaasConfig {
         ui: { labelField: 'rank' },
         access: { operation: { query: () => true } },
       },
+      VirtualLabel: {
+        fields: {
+          displayName: { type: 'virtual', virtual: true },
+        },
+        ui: { labelField: 'displayName' },
+        access: { operation: { query: () => true } },
+      },
     },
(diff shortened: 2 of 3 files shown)


borisno2:
## Code review

Ran a review of the diff via subagent. **Verdict: fix is correct and minimal — no blocking issues.**

The bug was that the resolved Label field name was passed straight into Prisma `orderBy`. A `virtual()` field has no backing column, so Prisma rejects the `orderBy` and 500s; the picker degrades that into a misleading "No results found". The `where`/`contains` filter was already virtual-safe (gated on `type === 'text'`), so `orderBy` was the only remaining unsafe sink. The fix branches to `orderBy: { id: 'asc' }` when the label field is virtual, which fully addresses the reported bug. The non-virtual path (including `text`, `integer`, and the `id`-fallback case) is unchanged.

### Follow-ups from review (all non-blocking, addressed or scoped out)

- **Build fix (addressed):** extracting `orderBy` into a variable dropped contextual literal typing, so `tsc` widened `'asc'` to `string` and failed — CI caught this. Fixed by annotating `orderBy: Record<string, 'asc'>` in `c1b847f`.
- **Discriminator robustness (addressed):** added a regression test asserting a field flagged virtual via `type: 'virtual'` alone (without the `virtual: true` flag) still falls back to `orderBy: { id: 'asc' }`.
- **Virtual label + `search` is a silent no-op (by design):** a computed column can't be SQL-filtered, so `where` stays `undefined` and the primary window returns the first N rows by `id`. Correct behavior; the current selection's label still renders via the `selectedIds` union.
- **Out of scope:** a `json`/`richText` field set as `ui.labelField` would also be non-orderable (Postgres has no ordering operator for `json`). Pathological and outside issue #676; left as-is. A future "field has no orderable column" guard would generalize this.

Tests: `packages/core` — 771 passed. `pnpm build` and `pnpm lint` green locally.

---

PR #691, opened 2026-07-15
docs: document ORM client access patterns, decline sync accessor (ADR-0014)

### LINKED ISSUE #682: No synchronous accessor for the configured ORM client, for module-init-time consumers that must build their own client
## What to build

The **Generated bundle** exposes its configured database client only indirectly, bundled inside an **async** context factory: `getContext` returns a `Promise` resolving to a full request-scoped context (ORM client + access-control + hooks + storage). The related seams are also async — `rawOpensaasContext` is a `Promise<Context>`, `config` is a `Promise`, and the private client accessor (`getPrisma()`) is async and unexported. There is no sanctioned way to obtain just the configured ORM client — connection/driver-adapter already resolved the same way the factory resolves it — **synchronously**.

That is a real integration problem for any consumer that must build its own client at **module load time** rather than per-request: most concretely, a third-party library (e.g. an authentication library's ORM adapter) whose own initialization contract constructs a client synchronously at import time (`const client = new SomeAdapter(new PrismaClient(...))` at a module top level). Top-level code cannot `await` the async factory, and the library's synchronous-construction requirement isn't something the host can change. Today the host's only options are to construct a **second, independent client by hand** against the same database — re-deriving the connection/driver-adapter selection the factory encapsulates — or to restructure the consumer to be async, which a third-party contract often forbids.

Filed **needs-triage** because this contends with a deliberate design choice and the direction is a maintainer call:

- Config resolution is **async by design** (the factory awaits `getConfig()` before constructing the client), so a synchronous client accessor is not a drop-in — it would require the configured client to be constructible without awaiting config resolution.
- The framework **already faced this exact need** for its own better-auth wiring and solved it *without* a synchronous seam: `createAuth` takes the async `rawOpensaasContext` promise, defers the entire `betterAuth()` construction behind a lazy Proxy, and reads `context.prisma` once the context resolves. So one legitimate outcome is to **document that async/lazy-proxy integration pattern as the sanctioned recipe** for module-init-time consumers, rather than adding a new synchronous API.

Decide between (a) a synchronous, client-only accessor (returns just the configured ORM client, no access-control/request-scoping, safe to call at module top level — which requires making the configured client synchronously constructible), and (b) a documented integration recipe for the "module-init consumer with its own client requirement" shape, covering how to reuse the factory's connection/driver-adapter selection as a single source of truth instead of a hand-rolled second client. Record the decision (likely an ADR, since it touches the bundle's public surface and the async-construction stance).

## Acceptance criteria

- [ ] A module-init-time consumer with a synchronous-construction requirement has a sanctioned, single-source-of-truth path to the configured ORM client — via either a synchronous client-only accessor or a documented recipe (mechanism TBD).
- [ ] The path reuses the same connection/driver-adapter selection the async context factory applies (no re-deriving `prismaClientConstructor` logic by hand).
- [ ] It carries no access-control or request-scoping — it is the bare client, distinct from the secured context.
- [ ] The decision (sync accessor vs documented async/proxy recipe) is recorded, noting the interaction with the async-config-resolution design.

## Blocked by

borisno2:
## Summary

Resolves the "no synchronous ORM client accessor" question raised in #682 as a documentation task, per the maintainer's grilling-session brief on the issue: **no new framework API**, just sanctioned recipes.

- Adds `docs/adr/0014-orm-client-access-stays-async-document-the-sync-escape-hatch.md`, recording the decision to keep client access async-only and documenting the synchronous escape hatch instead of adding a new accessor.
- Adds a new subsection to `CLAUDE.md`'s "Creating Context in Applications" covering three consumer shapes:
  1. `await`-capable code → `getContext()` / `context.prisma` (unchanged).
  2. Module-init-time consumers that only need to defer method calls → `rawOpensaasContext` + a lazy `Proxy`, the same pattern `@opensaas/stack-auth`'s `createAuth()` already uses for better-auth's Prisma adapter.
  3. Module-init-time consumers that need a resolved client value synchronously → import the generated `PrismaClient` and reuse the app's own `prismaClientConstructor` to build a second, independent client — explicitly called out as a second connection that bypasses access control.
- Fixes the generated `rawOpensaasContext` doc comment (`packages/cli/src/generator/context.ts`), which said "for synchronous initialization" — it's actually a `Promise<Context>` meant to be passed to a lazy-Proxy consumer, not awaited/used synchronously at module scope. Updated the corresponding CLI generator snapshots.
- Adds a changeset for `@opensaas/stack-cli` (patch — doc-comment fix only).

## Test plan

- [x] `pnpm build` (packages/core) so workspace deps resolve
- [x] `packages/cli`: `vitest run` — all 316 tests pass, including updated snapshots for the corrected doc comment
- [x] `pnpm lint` — 0 errors (3 pre-existing warnings unrelated to this change)
- [x] `pnpm format` and `pnpm manypkg fix` run clean
- [x] Changeset added for `@opensaas/stack-cli`

Closes #682

---
_Generated by [Claude Code](https://claude.ai/code/session_01X5C74fXfq3NmMfKSCBH6mv)_


borisno2:
## Self-review summary

Ran a two-angle review (correctness of doc code-examples; conventions/cross-references) before requesting human review. Findings and fixes:

**Confirmed and fixed:**
1. The synchronous-client example imported a nonexistent `prismaClientConstructor` named export from `opensaas.config.ts` — and worse, that file's default export can itself be a `Promise` when plugins are present, defeating the whole point of a "synchronous" recipe. Fixed by factoring the constructor into its own plain module (`lib/db-adapter.ts`) that both `opensaas.config.ts` and the synchronous consumer import.
2. The generated Prisma client import used a `.js` extension; the generator actually emits `.ts` (confirmed against `packages/cli/src/generator/context.ts:133` and its own generated-file convention). Fixed.
3. (Non-blocking, addressed anyway) `packages/auth/CLAUDE.md` already showed the `createAuth(config, rawOpensaasContext)` snippet but had no link back to the new ADR-0014/root-CLAUDE.md guidance, breaking the repo's existing "See ADR-000N and `packages/*/CLAUDE.md`" cross-reference convention (e.g. ADR-0012). Added the cross-link both directions.

**Checked and confirmed correct (no changes needed):**
- ADR-0014 structure matches ADR-0012/0013 exactly (H1 decision statement, `Status: accepted`, `## Context`/`## Considered options`/`## Consequences`).
- `createAuth`'s Proxy-based lazy-construction description matches `packages/auth/src/server/index.ts` exactly.
- The changeset (`@opensaas/stack-cli`, patch) matches the package name and is correctly scoped as a doc-comment fix.
- The changeset-bot comment listing 9 packages for release is expected — `.changeset/config.json` has `"fixed": [["@opensaas/stack-*"]]`, so all `@opensaas/stack-*` packages version in lockstep; not specific to this PR.

Test plan (all local, no CI-gated app to run since this is docs-only): `pnpm build` (core), `packages/cli` full `vitest run` (316 passed, including updated snapshots), `pnpm lint` (0 errors), `pnpm format` + `pnpm manypkg fix` clean.


---

PR #694, opened 2026-07-15
fix(cli): preserve host-added datasource keys in prisma.config.ts

### LINKED ISSUE #680: Generator drops host-added datasource keys when rewriting prisma.config.ts
## What to build

The generator rewrites `prisma.config.ts` on every run by regenerating the whole file from an internal template. That template's `datasource` block only ever emits the connection `url`, so any additional keys a host has added to the block (for example a shadow-database URL used for local migration diffing) are silently dropped the next time the generator runs — including when the generator is wired into a `postinstall`/build step.

Scope this issue to the clear defect: **preserve host-added `datasource` keys instead of overwriting the block wholesale.** The generator should update the keys it manages (the connection URL) while retaining keys it does not manage. This keeps `prisma.config.ts` generator-owned and root-located, consistent with the current documented stance, while no longer clobbering configuration the generator doesn't understand.

Filed **needs-triage** because the implementation approach is open — the current writer emits a template string with no read/parse/merge step, so a fix needs a decision on how to reconcile a generated block with an existing on-disk one. The acceptance criteria state the target behaviour, not the mechanism.

**Out of scope (deliberately):** making the `prisma.config.ts` location or full content broadly configurable, or letting a host point real migrations at a separate hand-authored schema. The file is currently treated as generator-owned and never relocated (see `output-paths.ts` and ADR-0003); reopening that stance is a larger decision and should be a separate issue if wanted.

**Where:** `writePrismaConfig` / `generatePrismaConfig` in `packages/cli/src/generator/prisma-config.ts` (wholesale `writeFileSync` of a freshly templated string); the `datasource` block is hardcoded to a single `url` key.

## Acceptance criteria

- [ ] Running generate preserves `datasource` keys the generator does not itself emit (e.g. `shadowDatabaseUrl`).
- [ ] The generator's own managed keys (the connection URL) are still written/updated correctly.
- [ ] No behaviour change for a project that has not added extra `datasource` keys.
- [ ] Test covering a pre-existing `prisma.config.ts` with an extra `datasource` key surviving a generate run.

## Blocked by

borisno2:
## Summary
- `writePrismaConfig` previously overwrote `prisma.config.ts` wholesale on every `generate` run, silently dropping any host-added `datasource` keys (e.g. a `shadowDatabaseUrl` used for local migration diffing).
- `writePrismaConfig` now reads the pre-existing file (if any) and extracts `datasource` entries it doesn't manage via a new `extractExtraDatasourceLines` helper, then re-emits them alongside the generator-managed `url` key.
- `generatePrismaConfig` gained an optional `extraDatasourceLines` parameter so it stays a pure generator function; `writePrismaConfig` is the only caller that reads from disk.
- No change in output for projects that haven't added extra `datasource` keys — byte-identical to before.

## Test plan
- [x] Added tests covering `extractExtraDatasourceLines` (extra key preserved / no extra key / no datasource block)
- [x] Added `writePrismaConfig` tests: a pre-existing file with an extra `shadowDatabaseUrl` key survives a re-generate, output is unchanged/byte-identical when there are no extra keys, and the file is created correctly when none exists yet
- [x] `pnpm --filter @opensaas/stack-cli` full test suite passes (323 tests)
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` run clean (no new warnings/errors)
- [x] `pnpm build` passes for `@opensaas/stack-cli`

Closes #680

---
_Generated by [Claude Code](https://claude.ai/code/session_01BL4EUk6wgRJpFPeQKtCDv2)_

### DIFF
--- a/.changeset/tame-eels-wonder.md
+++ b/.changeset/tame-eels-wonder.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+Preserve host-added `datasource` keys (e.g. `shadowDatabaseUrl`) in `prisma.config.ts` across `generate` runs instead of overwriting the block wholesale.
--- a/packages/cli/src/generator/prisma-config.test.ts
+++ b/packages/cli/src/generator/prisma-config.test.ts
@@ -1,5 +1,12 @@
-import { describe, it, expect } from 'vitest'
-import { generatePrismaConfig } from './prisma-config.js'
+import { describe, it, expect, beforeEach, afterEach } from 'vitest'
+import * as fs from 'fs'
+import * as path from 'path'
+import * as os from 'os'
+import {
+  generatePrismaConfig,
+  extractExtraDatasourceLines,
+  writePrismaConfig,
+} from './prisma-config.js'
 import type { OpenSaasConfig } from '@opensaas/stack-core'
 import { text } from '@opensaas/stack-core/fields'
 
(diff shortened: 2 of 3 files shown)


borisno2:
## Automated review

### Overview
Fixes #680: `writePrismaConfig` previously regenerated `prisma.config.ts` from a template on every `generate` run, silently dropping any host-added `datasource` keys (e.g. `shadowDatabaseUrl`). This PR reads the pre-existing file (if any), extracts non-`url` entries from its `datasource` block via a new `extractExtraDatasourceLines` helper, and re-emits them alongside the generator-managed `url` key.

### Code quality & style
- `generatePrismaConfig` stays a pure function — it takes `extraDatasourceLines` as a plain parameter rather than reading from disk itself; `writePrismaConfig` is the only I/O boundary. This matches the existing separation between the two functions in this file.
- JSDoc on both the new helper and the updated functions follows the file's existing documentation style (explains *why*, cross-references via `{@link}`).
- Naming (`extractExtraDatasourceLines`) is clear and consistent with the file's existing exports.

### Correctness
- The extraction is a documented "lightweight line-based" approach, not a real TS/AST parser — this is explicitly called out in the docstring, which is appropriate given the issue's stated scope (preserve flat `datasource` keys, not general config editing).
- Two edge cases worth being aware of (not blocking, both inherent to the documented approach):
  - A host `url:` line with non-standard formatting (e.g. a space before the colon) wouldn't be filtered by `!line.startsWith('url:')`, and would be re-emitted alongside the generator's own `url:` line. TypeScript object literals tolerate the duplicate key (last wins), so this is a formatting quirk rather than a functional bug.
  - A host value that spans multiple lines gets split per-line and re-indented uniformly, which is syntactically fine but may look different from the original formatting.
- The managed `url` key is always regenerated from the current env-var expression rather than preserved from disk — verified by the `preserves a host-added datasource key across a re-generate` test, which asserts the stale on-disk `url` value is replaced.

### Test coverage
Good coverage of the acceptance criteria:
- `extractExtraDatasourceLines`: extra key preserved, no extra key, no `datasource` block at all.
- `writePrismaConfig`: pre-existing file with extra key survives regenerate; output is byte-identical to before when there are no extra keys (no spurious churn for the common case); file is created correctly when none exists yet.

### Risk
Low — the change is additive and defaults to today's behavior (`extraDatasourceLines = []`) whenever there's nothing to preserve. CI (test, e2e, coverage) is green and the changed file shows 100% coverage.

No blocking issues found.


---

Issue #686, opened 2026-07-15
Plugin-injected lists should default closed, with an app-side seam to grant them access


borisno2:
> *This was generated by AI during triage.*

## Context

Per **ADR-0013** (access control belongs to the application, not plugins), operation-level access is the application's responsibility and plugin-injected lists should rely on the engine's deny-by-default (closed) behaviour rather than shipping permissive defaults. #678 enforces the narrow guardrail (an extension may not override a pre-existing list's access). This issue tracks the coupled hardening that ADR-0013 calls out as follow-up.

Two changes that must land together (closed-by-default is unusable without the seam):

1. **Strip plugins' permissive default access.** Plugin-created lists (e.g. the auth plugin's Session/Account/Verification/User) should not seed permissive operation access (`query: () => true`, self-only update/delete). With them removed, an injected list with no application-supplied access is closed. This is a security improvement: better-auth's own flows use the raw client and bypass access control, so login is unaffected — the only thing the permissive default did was expose every user/session to any `context.db` reader and the admin UI.

2. **Give the application a first-class seam to grant access to lists it did not declare.** Today the only way for an app to own a plugin list's access is to pre-declare that list itself and let the plugin merge into it — awkward and collision-prone, and unavailable for lists the app never declares (Session/Account/Verification). Design an app-side surface (e.g. an application-level access map keyed by list, applied after plugins run) so the application can open injected lists explicitly.

Filed **needs-triage**: the seam's shape is an open design question and warrants its own grilling pass before implementation.

## Acceptance criteria

- [ ] Plugin-created lists no longer ship permissive operation-level access; with no application-supplied access, an injected list is closed (deny-by-default).
- [ ] The application has a sanctioned, documented way to grant operation-level access to a plugin-injected list it did not declare, without pre-declaring the list to force a merge.
- [ ] better-auth login/session flows are unaffected (they use the raw client and bypass access control).
- [ ] The admin UI and `context.db` reads of auth lists reflect the application's configured access, defaulting to closed when none is set.
- [ ] Migration guidance for existing apps relying on the old permissive defaults.

## Blocked by

- None to start the design. Builds on ADR-0013 and is complementary to #678 (the guardrail) — land the seam before, or together with, stripping the defaults.

borisno2:
> *This was generated by AI during triage.*

[...]

PR #695, opened 2026-07-15
fix: enforce ADR-0013 — plugin extensions can't override existing list access

### LINKED ISSUE #678: Plugin list-extensions can silently overwrite an existing list's access configuration
## What to build

When a plugin extends an existing list, the operation-level access the extension carries is merged per-operation such that the extension's rules win for any operation it defines. As a result, a plugin extension can silently **replace** a list's existing access rules — for example turning an admin-only update into "anyone can update themselves" — with no warning. This is a general property of the extension/merge path, not specific to any single plugin (it surfaced via the auth provider-plugin path, but the mechanism is shared).

This issue is filed **direction-open**: the invariant we want is clear, but the right mechanism is a maintainer decision. Options raised:

- **Warn loudly** when an extension is about to replace access on a list that already has an access config it did not originate.
- **Merge additively** rather than letting the extension's rules win outright.
- **Require an explicit opt-in flag** before an extension may override an existing list's access.

Pick the semantics, record the decision (an ADR is likely warranted since this touches the access-control contract), then implement. The acceptance criteria below express the target invariant independent of mechanism.

## Acceptance criteria

- [ ] A plugin extension cannot silently drop or loosen a pre-existing list's operation-level access without either an explicit opt-in or a loud, visible warning (final mechanism TBD).
- [ ] Legitimate existing behaviour still works — plugins that add access for lists they create/own are unaffected.
- [ ] The chosen semantics are recorded (ADR or issue thread) before or alongside implementation.
- [ ] Tests cover the chosen guardrail, including the "extension would loosen host access" case.

## Blocked by

- None — can start immediately. Related to #677 (the auth model-key remap collision), which is the specific path this general guardrail would also have caught; the two fixes are independent and complementary.

borisno2:
## Summary

- `extendList()` in the plugin engine now throws a config-time error naming the plugin and the list when an extension carries `access.operation` for a list that **already exists** — per ADR-0013, operation-level access belongs to whoever owns the list (the application, or whichever party created it via `addList`), never a plugin extending it. Extending with only fields/hooks/mcp still works and leaves the target list's access untouched. `addList()` is unaffected — a plugin creating a new list may still set its access at creation.
- Removed the now-unused `mergeAccess()` per-key-spread helper that let an extension's access win per operation.
- The auth plugin (`packages/auth/src/config/plugin.ts`) no longer forwards `access` on either of its `extendList` paths (its own derived auth lists, and better-auth provider-plugin schema extensions) — it merges `fields`/`hooks`/`mcp` only. This was the reported footgun: an app's own `User` list (or any list colliding with a derived auth key) could have its access silently replaced by the auth plugin's permissive defaults (`query: () => true`, self-only update/delete).

## Test plan

- [x] `packages/core/tests/plugin-engine.test.ts` — replaced the two access-merging tests with: throws when an extension carries operation access for a pre-existing list; extension attempting to override existing access throws instead of silently winning; fields/hooks-only extension leaves access untouched; a plugin creating a new list may still set access via `addList`.
- [x] `packages/auth/tests/plugin-derived-keys.test.ts` — added a regression test: an app-declared `User` list with admin-only `query`/`update` access, extended by `authPlugin({})` at the default key, keeps its own access after the plugin runs (the acceptance-criteria scenario from #678).
- [x] `pnpm test` green in `packages/core` (773 tests), `packages/auth` (137 tests), `packages/rag` (178 tests)
- [x] Changeset added (`@opensaas/stack-core` + `@opensaas/stack-auth`, patch)

Closes #678

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xdv6YDJxoG3XWZzPHoc4Rq)_

### DIFF
--- a/.changeset/silent-plugins-guard-access.md
+++ b/.changeset/silent-plugins-guard-access.md
@@ -1,0 +1,6 @@
+---
+'@opensaas/stack-core': patch
+'@opensaas/stack-auth': patch
+---
+
+Fix a plugin's `extendList()` silently overwriting a pre-existing list's operation-level access. Per ADR-0013, an extension that carries `access.operation` for an existing list now throws a config-time error naming the plugin and the list; the auth plugin no longer forwards its own access when extending a list an app already declared.
--- a/packages/auth/src/config/plugin.ts
+++ b/packages/auth/src/config/plugin.ts
@@ -68,11 +68,13 @@ export function authPlugin(config: AuthConfig): Plugin {
           // Add or extend lists from plugin
           for (const [listName, listConfig] of Object.entries(pluginLists)) {
             if (context.config.lists[listName]) {
-              // List exists, extend it
+              // List already exists — merge fields/hooks/mcp in only. Access
+              // control belongs to whoever owns the list; per ADR-0013 an
+              // extension must never carry operation-level access for a
+              // pre-existing list (the plugin engine throws if it does).
               context.extendList(listName, {
                 fields: listConfig.fields,
                 hooks: listConfig.hooks,
-                access: listConfig.access,
                 mcp: listConfig.mcp,
               })
             } else {
@@ -94,11 +96,13 @@ export function authPlugin(config: AuthConfig): Plugin {
       // "merge auth fields into my User" behaviour.
       for (const [listName, listConfig] of Object.entries(authLists)) {
         if (context.config.lists[listName]) {
-          // A list already exists under this derived key — merge auth fields in.
+          // A list already exists under this derived key — merge auth fields
+          // in only. Access control belongs to whoever owns the list (the
+          // application declared it first), so the plugin never forwards its
+          // own access here — see ADR-0013.
           context.extendList(listName, {
             fields: listConfig.fields,
             hooks: listConfig.hooks,
-            access: listConfig.access,
             mcp: listConfig.mcp,
           })
         } else {
--- a/packages/auth/tests/plugin-derived-keys.test.ts
+++ b/packages/auth/tests/plugin-derived-keys.test.ts
@@ -72,6 +72,42 @@ describe('authPlugin - add-vs-extend with derived keys', () => {
     expect(user.fields).toHaveProperty('sessions')
   })
 
+  it("does not overwrite an app's admin-only access when the auth plugin extends the same default key (ADR-0013)", async () => {
+    // Regression test for #678: the auth plugin ships permissive defaults
+    // (query: () => true, self-only update/delete) for its own User list.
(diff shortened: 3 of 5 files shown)


borisno2:
## Self-review

**Overview:** Implements ADR-0013's chosen guardrail for #678. `extendList()` now throws a config-time error when a plugin extension carries `access.operation` for a list that already exists, instead of letting the extension's rules win per-operation. `addList()` is untouched. The auth plugin's two `extendList` call sites stop forwarding `access`, so extending an app-declared list (e.g. `User`) no longer replaces the app's access with the plugin's permissive defaults.

**Correctness**
- The guard checks `extension.access?.operation` before merging fields/hooks/mcp, and the merged list is built by spreading `...existing` first — so `access` is always inherited from the pre-existing list untouched, never explicitly reassigned. Confirmed via test that the throw happens before any mutation of the original config.
- `addList` path is unaffected — a plugin creating a new list still supplies its own access at creation, matching the ADR ("guardrail only fires on extending a pre-existing list").
- The auth plugin change is minimal and symmetric across both `extendList` call sites (its own derived auth lists, and better-auth provider-plugin schema extensions via `convertBetterAuthSchema`).

**Test coverage**
- `plugin-engine.test.ts`: covers the throw case, confirms the host's access is untouched pre-throw, confirms fields/hooks-only extensions still work and leave access alone, and confirms `addList` access still works.
- `plugin-derived-keys.test.ts`: new regression test matches the issue's exact acceptance-criteria scenario — an app `User` list with admin-only `query`/`update`, extended by `authPlugin({})` at the colliding default key, keeps its own access and the plugin's fields still merge in.

**One low-severity note (non-blocking):** `ListAccessControl` also supports a bare-function shorthand (`access: isAdminFn` applies to all operations), but `PluginContext.extendList`'s type only accepts the object form (`{ operation?: OperationAccess }`), so a well-typed plugin can't construct the shorthand here — the type system already closes that door. Only a plugin bypassing types (e.g. `as any`) could pass a shorthand function through `extendList`, and if it did, the current guard (`extension.access?.operation`) wouldn't detect it — the value would just be silently dropped (never applied, so no silent *overwrite*, but also no loud error). Given the extension type doesn't expose this shape, I don't think it's worth extra runtime code for now; flagging for awareness rather than requesting a change.

**Process:** `pnpm test` green in core/auth/rag, `pnpm lint`/`manypkg fix`/`format` clean, changeset added. No scope creep — field-level access on replaced field keys and the app-side access-passthrough seam are correctly left to their own tracked follow-ups (mentioned in the issue as out of scope / #686).

No changes requested.


---

PR #698, opened 2026-07-16
fix(auth): mirror better-auth FK shape on Session.user/Account.user

### LINKED ISSUE #679: Auth-derived user FK relationships must mirror better-auth's shape (no @@index, onDelete: Cascade)
## What to build

Manually-declared `relationship()` fields support suppressing the default foreign-key `@@index` via `isIndexed: false`. That is exactly what lets an adopting project match a live table whose FK column carries only an inline `@unique` constraint (or no index at all) without a destructive `DROP INDEX` / `CREATE INDEX` migration.

Relationships that the table-adoption helper derives automatically for adopted **Auth lists** do not expose that control — they always emit the default FK `@@index`. So an adopted schema can diverge from the live one on FK indexes, and the only current workaround is post-generate schema-text surgery (matching and stripping the generated `@@index` line), which defeats the point of adoption reaching **Schema parity** through config alone.

Extend the adoption helper's per-model/per-field configuration so adoption-derived relationships can set `isIndexed`, matching the manually-declared `relationship()` option. The generator already honours `isIndexed` for manual relationships (`field.isIndexed ?? true`); the gap is purely that the adoption/derivation path constructs its relationship fields with only `ref` + an optional FK column map and never threads an index option through.

Scope this issue to `isIndexed`. The same "adopt the exact existing shape" gap also covers FK nullability, referential action, and column defaults — note these as related follow-ups rather than pulling them in here unless trivially included.

**Where:** adoption config in `packages/auth/src/config/adopt-better-auth-tables.ts`; relationship construction in `packages/auth/src/config/derive-auth-lists.ts` (currently only sets `ref` and `db.foreignKey.map`).

## Acceptance criteria

- [ ] The adoption config surface accepts a per-relationship index control equivalent to `relationship({ isIndexed: false })`.
- [ ] Setting it to `false` suppresses the `@@index` on the derived foreign key; default behaviour (index emitted) is unchanged when the option is unset.
- [ ] An adopted schema with the option set diffs clean against a live table that has no separate FK index — no post-generate string surgery required.
- [ ] Regression test asserting the generated schema omits `@@index` for an adoption-derived relationship when the option is set, and still emits it by default.

## Blocked by

borisno2:
## Summary

- `Session.user` and `Account.user` (the many-to-one FKs on `userId`) are now derived with `isIndexed: false` and an `onDelete: Cascade` referential action (via the relationship's field-level `db.extendPrismaSchema`), matching a live better-auth database's shape.
- Previously the derived relationships only set the FK column map, so the generator emitted a default `@@index([userId])` and the default Prisma referential action (`Restrict`) — neither of which a standard better-auth schema has. This meant `migrate diff` always showed a spurious index drop and a referential-action change on every auth FK, in both greenfield and adopted installs.
- Both greenfield (`deriveAuthLists` with default models) and adoption (`adoptBetterAuthTables` + custom `modelName`/`fields`) go through the same `deriveAuthLists` derivation, so this is a single fix covering both paths — no new config surface.
- Existing `@@unique` mirroring on `User.email` / `Session.token` is unchanged.

## Test plan

- [x] Added regression tests in `packages/auth/tests/derive-auth-lists.test.ts` asserting:
  - `Session.user` / `Account.user` carry `isIndexed: false`
  - the `db.extendPrismaSchema` callback rewrites the relation line to include `onDelete: Cascade` (and leaves the FK line untouched)
  - the cascade `extendPrismaSchema` is retained alongside a `userId` column override (adoption case)
  - `@@unique` mirroring on `email`/`token` is unaffected
- [x] `pnpm test` in `packages/auth` (138 tests passing)
- [x] `pnpm test` in `packages/cli` (323 tests passing) — the underlying `isIndexed: false` + `extendPrismaSchema` generator mechanism is exercised there
- [x] Changeset added (`minor`, per the issue's own guidance — this changes generated schema output for existing greenfield stack-auth apps)

## Migration note

Existing greenfield stack-auth apps will see a schema diff on the next `pnpm generate`: a migration that drops the `userId` index on `Session`/`Account` and adds `onDelete: Cascade` to both foreign keys. Deleting a user will now cascade to their sessions/accounts instead of being blocked — call this out to app owners before they apply the generated migration.

Closes #679

---
_Generated by [Claude Code](https://claude.ai/code/session_01CX4WgRfZKRUf89Wh7921wC)_

### DIFF
--- a/.changeset/mirror-auth-fk-cascade.md
+++ b/.changeset/mirror-auth-fk-cascade.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-auth': minor
+---
+
+The derived `Session.user` and `Account.user` foreign keys now mirror a live better-auth database's shape: no `@@index([userId])` (better-auth carries no separate FK index) and `onDelete: Cascade` (Prisma's default was `Restrict`). This applies to both greenfield installs and adopted databases, so a generated Auth schema diffs clean against a standard better-auth Prisma schema.
+
+Migration note: existing greenfield stack-auth apps will see a schema diff on next `pnpm generate` — a migration that drops the `userId` index on `Session`/`Account` and adds `onDelete: Cascade` to both foreign keys. Review the generated migration before applying it; deleting a user now cascades to their sessions and accounts instead of being blocked.
--- a/packages/auth/src/config/derive-auth-lists.ts
+++ b/packages/auth/src/config/derive-auth-lists.ts
@@ -27,6 +27,7 @@
 import { list } from '@opensaas/stack-core'
 import { text, timestamp, checkbox, relationship } from '@opensaas/stack-core/fields'
 import type { ListConfig } from '@opensaas/stack-core'
+import type { RelationshipField } from '@opensaas/stack-core/fields'
 import type { ExtendUserListConfig } from '../lists/index.js'
 import type { AuthAccessConfig, NormalizedAuthModelConfig, NormalizedAuthModels } from './types.js'
 
@@ -106,14 +107,23 @@ function fieldDb(fieldName: string, fields: Record<string, string>): { map: stri
 }
 
 /**
- * Build the foreign-key `db` config for a `user` relationship, honouring a
- * `userId` column override from the better-auth `fields` map.
+ * Build the `db` config for a `user` relationship (`Session.user` /
+ * `Account.user`), honouring a `userId` column override from the better-auth
+ * `fields` map and mirroring better-auth's own FK shape: no separate FK index
+ * — the index is applied at the field level via `isIndexed: false` — and
+ * `onDelete: Cascade`, so a generated Auth schema diffs clean against a live
+ * better-auth database on both dimensions instead of showing a spurious index
+ * drop and a referential-action change (issue #679).
  */
-function userForeignKeyDb(
-  fields: Record<string, string>,
-): { foreignKey: { map: string } } | undefined {
+function userRelationshipDb(fields: Record<string, string>): NonNullable<RelationshipField['db']> {
   const column = fields.userId
-  return column ? { foreignKey: { map: column } } : undefined
+  return {
+    ...(column ? { foreignKey: { map: column } } : {}),
+    extendPrismaSchema: ({ fkLine, relationLine }) => ({
+      fkLine,
+      relationLine: relationLine.replace('@relation(', '@relation(onDelete: Cascade, '),
+    }),
+  }
 }
 
 /**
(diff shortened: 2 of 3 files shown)


borisno2:
## Self-review (via /review)

### Overview
Mirrors better-auth's own FK shape onto the derived `Session.user` / `Account.user` relationships: `isIndexed: false` (no `@@index([userId])`) plus `onDelete: Cascade` via `db.extendPrismaSchema`. Both greenfield and adoption configs flow through the same `deriveAuthLists` helper, so one change covers both. 3 files changed, +81/-9.

### Correctness
- Traced `getPrismaRelation` in `packages/core/src/fields/index.ts`: with `field.isIndexed === false`, `indexType !== false` is `false`, so `foreignKeyIndex` is `undefined` — no `@@index` emitted. Confirmed.
- `Session.user`/`Account.user` are the many-to-one (FK-owning) side of a bidirectional, non-one-to-one relationship (`User.sessions`/`User.accounts` are `many: true`), so `shouldHaveForeignKey` always returns `true` here — no ambiguity about which side owns the FK.
- The `extendPrismaSchema` rewrite (`'@relation(' → '@relation(onDelete: Cascade, '`) matches the exact pattern already documented in root `CLAUDE.md` and exercised by existing `packages/cli` generator tests (self-referential `onDelete: SetNull` case) — same mechanism, not a new code path in the generator.
- `userForeignKeyDb` → `userRelationshipDb` correctly merges the optional `foreignKey: { map }` override with the new `extendPrismaSchema`, verified by the new "keeps the cascade `extendPrismaSchema` alongside a `userId` column override" test (adoption case).

### Test coverage
- New tests assert `isIndexed: false` on both fields, exercise `extendPrismaSchema` with realistic `fkLine`/`relationLine` input and assert the exact output string, confirm `@@unique` mirroring (email/token) is untouched, and cover the adoption (`userId` column remap) case.
- Full generator-level assertion (actual `@@index` absence in generated schema text) isn't re-tested end-to-end inside `packages/auth` — `packages/auth` has no dependency on `packages/cli`'s generator, and the `isIndexed`/`extendPrismaSchema` → schema-text mechanism is already covered generically in `packages/cli`'s generator tests. Field-config-level assertions are the right scope given that package boundary; flagging only so it's a conscious call, not an oversight.
- `pnpm test` (auth: 138 passed, cli: 323 passed), `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all clean.

### Style / conventions
- Follows existing file conventions (JSDoc above each helper, `NonNullable<RelationshipField['db']>` return type, no switch statements). Minor nit: the updated JSDoc on `userRelationshipDb` ("no separate FK index — the index is applied at the field level via `isIndexed: false`") is a little dense on first read; not a blocker.

### Risks
- Behavior change is intentional and scoped to what the issue requested: this changes generated schema for existing greenfield stack-auth apps (index drop + cascade delete). Called out clearly in the PR description, changeset, and marked `minor` per the issue's explicit guidance despite being categorized as a bug fix.
- No security or access-control implications — schema/generator only.

**Verdict:** No blocking issues found. Change is minimal, correctly scoped, well-tested at the appropriate layer, and matches the issue's re-triaged acceptance criteria.


---

PR #713, opened 2026-07-18
Design system 1/7: token contract + theme compiler, proven through Button

borisno2:
Implements #705. Part of #704.

Foundation slice of the admin UI design system: the single theming contract and the config layer that compiles onto it, proven by one primitive (Button) consuming it end-to-end.

### Token contract (ui stylesheet — `packages/ui/src/styles/globals.css`)
- Full Theme token vocabulary: shadcn color set **plus `success`/`warning` (+foregrounds)** and the `gradientFrom`/`gradientTo` pair, `--font-sans`/`--font-mono`/`--font-heading` (heading defaults to sans), a single `--radius` knob with derived `sm`/`md`/`lg` sizes, and `--shadow-sm`/`--shadow-md`/`--shadow-lg`.
- Single-definition `light-dark()` structure: raw `--color-*-light` / `--color-*-dark` values live side by side in `:root`; the `@theme` block declares the contract tokens as `light-dark(var(--…-light), var(--…-dark))`. Utilities (`bg-primary`, `rounded-md`, `shadow-sm`, `font-sans`) resolve through these vars, so overriding either the raw `-light`/`-dark` vars or the contract token propagates.
- `data-theme` mechanism (light/dark/system via `color-scheme`) for later slices.

### Config layer (core — `packages/core/src/config/types.ts`)
- New `ThemeConfig` = `{ preset?, colors?, darkColors?, fonts?, radius?, shadows? }`; added `ThemeFonts`/`ThemeShadows`. Old bare-HSL-triplet `ThemeColors` format is gone (clean break, ADR-0015). Values are any valid CSS color string.

### Compiler (ui — `packages/ui/src/lib/theme.ts`)
- `compileTheme(theme: ThemeConfig): string` — pure `ThemeConfig → CSS` function. Preset merge (default `modern`) + light/dark override merge; emits color overrides onto the **same** `--color-*-light`/`-dark` vars the stylesheet declares, and fonts/radius/shadows onto their contract tokens (only when provided). Colors pass through **verbatim** — the compiler never parses color values.
- **Bare-triplet warning:** developer-supplied color values are tested against `/^[\d.]+\s+[\d.]+%\s+[\d.]+%$/`; a match fires a dev-mode `console.warn` suggesting an `hsl(...)` wrap (silent in production). Preset values are always valid, so only user input can trigger it.

### Button restyle (`packages/ui/src/primitives/button.tsx`)
[...]

Issue #706, opened 2026-07-18
Design system 2/7: dark mode end-to-end — data-theme mechanism + ThemeToggle


borisno2:
## Parent

#704 (spec: `specs/THEMING.md`, "Dark mode")

## What to build

User-controllable color scheme on top of the `light-dark()` token structure. A `data-theme="light" | "dark"` attribute on the root element sets `color-scheme` and overrides the system preference; attribute absent means system. Ship a small `ThemeToggle` client component with three states (light/dark/system) that sets the attribute, persists the choice to `localStorage`, and restores it on mount without a flash of wrong scheme; mount it in the default Admin chrome's user menu (removable via composition). Document the developer recipe for pinning the admin to a single scheme by setting the attribute statically.

## Acceptance criteria

- [ ] Setting `data-theme` on the root flips all tokens regardless of OS preference; removing it restores system behaviour
- [ ] `ThemeToggle` cycles light/dark/system, persists to `localStorage`, restores on mount without flashing the wrong scheme
- [ ] Toggle appears in the default user menu; composed custom chrome can omit it
- [ ] Browser-mode tests cover attribute setting, persistence, and restore-on-mount
- [ ] Demoable in the starter example
- [ ] Changeset for ui (`pr-changeset` skill)

## Blocked by

- #705

Issue #705, opened 2026-07-18
Design system 1/7: token contract + theme compiler, proven end-to-end through Button


borisno2:
## Parent

#704 (spec: `specs/THEMING.md`; ADR-0015, ADR-0016)

## What to build

The single theming contract and the config layer that compiles onto it, proven by one primitive consuming it end-to-end. Define the full Theme token vocabulary in the ui package stylesheet — the shadcn color set plus `success`/`warning` (+foregrounds) and the gradient pair, `--font-sans`/`--font-mono`/`--font-heading` (heading defaults to sans), `--radius` with derived sizes, `--shadow-sm/md/lg` — keeping the single-definition `light-dark()` structure. Replace the old `ThemeConfig` shape and CSS generation with the new config shape (`preset`/`colors`/`darkColors`/`fonts`/`radius`/`shadows`) and a pure compiler that emits token overrides verbatim (any valid CSS color string; never parses colors), with a dev-mode warning when a value matches the bare HSL-triplet pattern. Restyle Button (only) to the restrained direction consuming the new tokens, so the whole path is demoable: setting a brand color and a `next/font` variable in an example's config visibly changes the admin's buttons and fonts.

Clean break per ADR-0015 — no triplet auto-wrapping. Old gradient/preset key names survive where the spec says so.

## Acceptance criteria

- [ ] Stylesheet defines exactly the spec's token vocabulary (colors incl. success/warning, fonts, radius + derived, shadows) with light/dark values side by side via `light-dark()`
- [ ] New `ThemeConfig` type in core matches the spec shape; old HSL-triplet fields/format gone
- [ ] Compiler is a pure function `ThemeConfig → CSS string`: preset merge, verbatim value pass-through, fonts/radius/shadows emission
- [ ] Dev-mode warning fires on bare-triplet values with an `hsl()` wrapping suggestion
- [ ] Button consumes only new tokens (colors, radius, shadow, font) — no hardcoded values
- [ ] An example demonstrates config-set brand color + `next/font` font flowing through to rendered buttons
- [ ] Compiler unit-test suite covers presets, light/dark overrides, all token groups, and the warning
- [ ] Changesets for ui and core (`pr-changeset` skill)

## Blocked by

None - can start immediately

Issue #704, opened 2026-07-18
PRD: Admin UI design system — themeable tokens, restrained restyle, chrome polish


borisno2:
Spec: `specs/THEMING.md` on branch `claude/admin-ui-design-system-zhrdbw` (with ADR-0015, ADR-0016, and new `CONTEXT.md` glossary terms: Theme token, Theme preset, Slot, Admin chrome).

## Problem Statement

Developers building admin-heavy apps on the stack cannot make the admin UI look like *their* product. The current theming pipeline is broken — configured themes emit invalid CSS because the config layer and the stylesheet drifted onto incompatible color formats — fonts cannot be themed at all, and there is no supported way to change component design short of forking the package. The default look is also dated: a high-saturation neon palette that reads as a demo, not a tool teams stare at all day.

## Solution

A single, un-driftable theming contract: every visual decision in the admin UI flows through named Theme tokens (colours, fonts, radius, shadows) that a developer can override from CSS or from `opensaas.config.ts` — both layers writing to the same tokens so they can never diverge again. Component design is customizable through a clear ladder — tokens, `className`/`classNames` props, stable `data-slot` selectors, and composition — while the prebuilt admin gets a restrained, modern restyle with polished chrome (page headers, empty states, skeletons, dark-mode toggle). Presets keep their names (`modern`/`classic`/`neon`) with re-curated visuals, and the admin UI stays fully composable so developers and coding agents can assemble complete custom forms from the same components.

## User Stories

1. As a developer, I want to set my brand's primary colour once in `opensaas.config.ts`, so that buttons, focus rings, and active nav all follow my brand without touching CSS.
2. As a developer, I want to supply colours in any CSS format (oklch, hex, `hsl()`), so that I can paste values straight from my design tool.
3. As a developer, I want to override theme tokens from my own stylesheet, so that I can theme the admin without touching config or running a Tailwind build.
4. As a developer, I want to set `--font-sans`, `--font-mono`, and `--font-heading` from `next/font` variables, so that the admin uses my product's typography with proper font loading.
5. As a developer, I want to pick a preset (`modern`/`classic`/`neon`) as a starting point and override individual tokens, so that I get a coherent look with minimal effort.
6. As a developer, I want to adjust the global radius and shadow tokens, so that I can make the admin fully flat or soft without forking components.
7. As a developer, I want status colours (`success`, `warning`, `destructive`) to be tokens, so that badges and toasts match my brand's status palette.
[...]

PR #716, opened 2026-07-18
feat(ui): user-controllable dark mode — data-theme mechanism + ThemeToggle

borisno2:
Implements #706. Part of #704.

Completes the user-facing dark-mode story on top of the `light-dark()` token contract and `data-theme` / `color-scheme` mechanism already merged in #705 (#713 on `main`). No CSS was re-added — the token sheet already flips every token when `data-theme` is set and follows the system preference when it's absent.

## What's new

- **`ThemeToggle`** (`'use client'`) — a single-button control that cycles light → dark → system, writes `data-theme` on `<html>`, persists the choice to `localStorage` (`opensaas-theme`), and restores it on mount. Reads the persisted value through `React.useSyncExternalStore`, which is hydration-safe (stable `'system'` server snapshot) and avoids `setState`-in-effect; a single effect reconciles the DOM attribute.
- **`ThemeScript`** — a server-safe inline `<script>` for the document `<head>` that applies the saved scheme **before first paint**, preventing a flash of the wrong scheme before hydration. This is the flash-prevention mechanism; `suppressHydrationWarning` on `<html>` covers the intentional attribute diff.
- **Shared `lib/theme-mode.ts`** — single source for the storage key, `ThemeChoice` type, apply/read/persist helpers, a tiny same-tab subscribe store, and the inline-script source, so the script and component can never disagree.
- **Default chrome** — `ThemeToggle` appears in the user menu (`UserMenu`). It is **opt-out via composition**: custom chrome that builds its own `Navigation`/`UserMenu` simply omits it. When there is no session (e.g. the `starter`), the `Navigation` footer still renders the toggle so it stays reachable and demoable.
- **Docs** — `packages/ui/CLAUDE.md` documents the toggle, the `ThemeScript` setup, and the recipe for **pinning to a single scheme** statically (`<html data-theme="dark">`, no toggle/script). The `starter` layout wires `ThemeScript` in `<head>`.

## How the flash is prevented

`ThemeScript` runs a dependency-free inline script in `<head>` before any module loads: it reads `localStorage` and sets/removes `data-theme` synchronously, so the correct scheme is painted immediately. `ThemeToggle` then restores the same value on mount via `useSyncExternalStore` (no visible re-flash).

## Opt-out composition

The toggle is not forced by `AdminUI`. It lives inside the composable `UserMenu` (and the anonymous footer branch of `Navigation`). Any app assembling its own chrome from `/standalone` + `/primitives` gets it only if it renders `ThemeToggle`.

## Tests

- Browser-mode suite `tests/browser/components/ThemeToggle.browser.test.tsx` (7 tests): attribute setting across the cycle, `localStorage` persistence, restore-on-mount, stale-attribute reconciliation, and the pre-hydration script's effect. Asserts external behaviour only (root attribute, storage entry) — never internal state.
- `vitest.config.ts` gained an optional `PLAYWRIGHT_CHROMIUM_EXECUTABLE` env passthrough (unset in normal CI, so no behaviour change) to allow pointing at a preinstalled Chromium in sandboxed environments.

## Checks

- `pnpm lint`: pass (0 errors; 3 pre-existing warnings in blog/cli/core, none in changed files)
- Typecheck / `pnpm build` (ui + core): pass
- `pnpm test` (ui): 233 passed
- `pnpm test:browser` (ui): 61 passed (incl. the 7 new ThemeToggle tests)
- Changeset added (`@opensaas/stack-ui` minor)

---

### DIFF
--- a/packages/ui/src/components/ThemeScript.tsx
+++ b/packages/ui/src/components/ThemeScript.tsx
@@ -1,0 +1,31 @@
+import { themeInitScript } from '../lib/theme-mode.js'
+
+/**
+ * Pre-hydration inline script that restores the persisted color-scheme choice
+ * before first paint, preventing a flash of the wrong scheme.
+ *
+ * Render it as early as possible in the document — ideally the first child of
+ * `<head>` (or the top of `<body>`) in your root layout — so it runs before the
+ * admin chrome paints:
+ *
+ * ```tsx
+ * import { ThemeScript } from '@opensaas/stack-ui'
+ *
+ * export default function RootLayout({ children }) {
+ *   return (
+ *     <html lang="en" suppressHydrationWarning>
+ *       <head>
+ *         <ThemeScript />
+ *       </head>
+ *       <body>{children}</body>
+ *     </html>
+ *   )
+ * }
+ * ```
+ *
+ * It is a plain server-safe element (no client runtime); the `ThemeToggle`
+ * component reads and writes the same `localStorage` key at runtime.
+ */
+export function ThemeScript() {
+  return <script dangerouslySetInnerHTML={{ __html: themeInitScript }} />
+}
--- a/packages/ui/src/components/ThemeToggle.tsx
+++ b/packages/ui/src/components/ThemeToggle.tsx
@@ -1,0 +1,89 @@
+'use client'
+
+import * as React from 'react'
+import { Monitor, Moon, Sun } from 'lucide-react'
+import { Button } from '../primitives/button.js'
+import {
+  type ThemeChoice,
+  applyThemeChoice,
+  readStoredChoice,
+  storeChoice,
+  subscribeThemeChoice,
+} from '../lib/theme-mode.js'
+
+export interface ThemeToggleProps {
+  /** Extra classes merged onto the toggle button. */
+  className?: string
+}
+
+// Cycle order for the single-button control: light → dark → system → light.
+const NEXT_CHOICE: Record<ThemeChoice, ThemeChoice> = {
+  light: 'dark',
+  dark: 'system',
+  system: 'light',
+}
+
+const LABELS: Record<ThemeChoice, string> = {
+  light: 'Light',
+  dark: 'Dark',
+  system: 'System',
+}
+
+const ICONS: Record<ThemeChoice, typeof Sun> = {
+  light: Sun,
+  dark: Moon,
+  system: Monitor,
+}
+
+// Stable server snapshot: before hydration we can't read localStorage, so the
+// server (and first client paint) assume 'system'. `useSyncExternalStore` then
+// reconciles to the persisted value on the client without a hydration warning.
+const getServerChoice = (): ThemeChoice => 'system'
+
+/**
+ * User-facing color-scheme control for the admin chrome. A single button that
+ * cycles light → dark → system, writing the choice to the document root's
+ * `data-theme` attribute (see `styles/globals.css`) and persisting it to
+ * `localStorage`. On mount it restores the persisted choice; a first-paint
+ * flash is prevented separately by the `ThemeScript` inline script in `<head>`.
+ *
+ * This is a Client Component. It is mounted in the default chrome's user menu
+ * but is fully composable — custom chrome simply omits it, or places it
+ * anywhere via the exported component.
+ */
+export function ThemeToggle({ className }: ThemeToggleProps) {
+  // Read the persisted choice through an external store so the rendered icon
+  // stays in sync with `localStorage` (this tab and others) without a
+  // setState-in-effect or a hydration mismatch.
+  const choice = React.useSyncExternalStore(subscribeThemeChoice, readStoredChoice, getServerChoice)
+
+  // Reconcile the DOM attribute with the current choice. On mount this restores
+  // the persisted scheme (the inline ThemeScript may already have pinned it
+  // before paint); on change it applies the new choice.
+  React.useEffect(() => {
+    applyThemeChoice(choice)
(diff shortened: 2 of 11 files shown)


packages/ui/src/components/ThemeScript.tsx
```diff
@@ -0,0 +1,31 @@
...
+ *     </html>
+ *   )
+ * }
+ * ```
+ *
+ * It is a plain server-safe element (no client runtime); the `ThemeToggle`
+ * component reads and writes the same `localStorage` key at runtime.
+ */
+export function ThemeScript() {
+  return <script dangerouslySetInnerHTML={{ __html: themeInitScript }} />
```
borisno2, line 30: Minor (non-blocking): this inline `dangerouslySetInnerHTML` script will be blocked by a strict `script-src` CSP that requires a nonce (common in Next.js apps that set a CSP). Consider accepting an optional `nonce?: string` prop and forwarding it to the `<script nonce={nonce}>` so CSP-hardened apps can still use the flash-prevention script. The content itself is safe (no interpolated user input — only `JSON.stringify(THEME_STORAGE_KEY)`), so this is purely about CSP compatibility.

---

packages/ui/src/components/ThemeToggle.tsx
```diff
@@ -0,0 +1,89 @@
...
+  // Read the persisted choice through an external store so the rendered icon
+  // stays in sync with `localStorage` (this tab and others) without a
+  // setState-in-effect or a hydration mismatch.
+  const choice = React.useSyncExternalStore(subscribeThemeChoice, readStoredChoice, getServerChoice)
+
+  // Reconcile the DOM attribute with the current choice. On mount this restores
+  // the persisted scheme (the inline ThemeScript may already have pinned it
+  // before paint); on change it applies the new choice.
+  React.useEffect(() => {
+    applyThemeChoice(choice)
```
borisno2, line 64: Nit (non-blocking): the toggle's icon/label render from the `'system'` server snapshot on first paint, then reconcile to the persisted choice after hydration via `useSyncExternalStore`. That means a user who has pinned e.g. Dark will briefly see the "System" label flicker to "Dark" on load. The page *scheme* itself does not flash (that's handled pre-paint by `ThemeScript`), so this is cosmetic-only and the correct trade-off for hydration safety — just calling it out. No change required.

---

borisno2 (commented):
## Review: feat(ui) user-controllable dark mode — `data-theme` + ThemeToggle

**Verdict: APPROVE** (submitting as a Comment because GitHub won't let the authoring identity file a formal Approve). Clean, well-tested, and the risk areas hold up under scrutiny. Two minor, non-blocking notes inline.

### What it does
Adds a user-facing color-scheme control on top of the already-merged `light-dark()` token contract: a `ThemeToggle` client component (cycles light → dark → system, writes `data-theme` on `<html>`, persists to `localStorage['opensaas-theme']`), a pre-hydration `ThemeScript` for `<head>` to prevent a flash of the wrong scheme, and shared helpers in `lib/theme-mode.ts`. Wires the toggle into the default `Navigation`/`UserMenu` chrome and the starter example layout.

### Risk areas — verified
- **Flash-of-wrong-scheme / hydration:** Correct. `ThemeScript` runs the inline `themeInitScript` in `<head>` before paint, pinning `data-theme` from `localStorage`, so the page paints the right scheme immediately. `ThemeToggle` reads via `useSyncExternalStore` with a **stable `'system'` server snapshot** (`getServerChoice`), so React's initial client render matches SSR and reconciles post-hydration without a warning. The only server/client DOM divergence is the `data-theme` attribute mutated by the inline script, and `suppressHydrationWarning` is correctly scoped to the single `<html>` element (it does not recurse into children, so it isn't masking real mismatches). `getSnapshot` returns a primitive string, so no `useSyncExternalStore` tearing/loop.
- **Prop serializability:** `ThemeToggle` exposes only `className?: string`. Rendered as `<ThemeToggle />` / `<ThemeToggle className="…" />` — no non-serializable props cross the server/client boundary; props are minimal.
- **vitest.config.ts passthrough:** True no-op when `PLAYWRIGHT_CHROMIUM_EXECUTABLE` is unset — `playwright({})` is equivalent to the prior `playwright()`. `process.env.X` is `string | undefined` and handled via ternary; no `any`/casts, normal CI behavior unchanged.
- **Opt-out composition:** The toggle is injected only by the *default* `Navigation`/`UserMenu`. Apps composing custom chrome build their own `Navigation` and simply don't render it — AdminUI does not force-inject it in a way that can't be removed. Pinning to a single scheme (omit both, set `data-theme` statically) is documented.
- **localStorage / conventions:** Single source of truth for the key (`THEME_STORAGE_KEY`, reused in `themeInitScript` via `JSON.stringify`). `read`/`store`/`subscribe` all wrap access in `try/catch` for private-mode/quota failures. Same-tab writes notify in-page listeners (the `storage` event only fires cross-tab) — correct. ESM `.js` import extensions used throughout. No `any` or type casts introduced. A `@opensaas/stack-ui` **minor** changeset is present.

### Test coverage
Good — browser tests assert external behavior only (root `data-theme`, `localStorage`, restore-on-mount, cycle order, and both flash-prevention branches of the inline script). Appropriately avoids asserting internal state.

### Minor notes (non-blocking, see inline)
1. `ThemeScript` inline script will be blocked by a strict nonce-based CSP — consider an optional `nonce` prop for CSP-hardened apps.
2. Toggle label/icon flickers from "System" to the persisted choice on hydration (cosmetic only; page scheme does not flash). Acceptable trade-off for hydration safety.

Nothing here blocks merge.

---

Issue #707, opened 2026-07-18
Design system 3/7: re-curated presets (modern / classic / neon)


borisno2:
## Parent

#704 (spec: `specs/THEMING.md`, "Visual direction")

## What to build

The preset catalog re-expressed in the new token vocabulary, keeping the existing names so preset-only configs upgrade unchanged. `modern` (default): the restrained direction — low-chroma neutral surfaces, one saturated brand color used sparingly, gradient pair as garnish, quieter muted text, softer radius. `classic`: flat, no gradients, blue primary, enterprise-safe. `neon`: the current high-chroma cyan/purple/pink personality preserved. Each preset supplies light and dark values for every token including the new `success`/`warning` and shadow tokens.

**HITL — design sign-off required:** the PR must include light and dark screenshots of representative admin screens under all three presets for maintainer review before merge.

## Acceptance criteria

- [ ] All three presets define every token (light + dark) in the new vocabulary; no missing tokens fall through to another preset
- [ ] `modern` matches the restrained direction; `classic` is flat/gradient-free; `neon` preserves the current high-chroma identity
- [ ] Switching `preset` in an example's config visibly reskins the admin with no other changes
- [ ] PR includes light/dark screenshots of all three presets; maintainer sign-off recorded
- [ ] Changeset for ui (`pr-changeset` skill)

## Blocked by

- #705

PR #718, opened 2026-07-18
feat(ui): re-curate theme presets (modern / classic / neon)

borisno2:
Implements #707. Part of #704.

Re-expresses the preset catalog in the new token vocabulary (built on the #705 token contract already on `main`), keeping the names `modern` / `classic` / `neon` so preset-only configs upgrade unchanged.

## What changed

- **`packages/ui/src/lib/theme.ts`** — re-curated the three presets and extended each preset with optional `radius` + `shadows` so a preset can express its own shape/elevation. `compileTheme` now merges the selected preset's radius/shadows under any config overrides (colors/fonts behaviour unchanged). Every `#705` compiler-contract test stays green — `modern` deliberately omits radius/shadows so `compileTheme({})` still emits none and defers to the stylesheet.
- **`packages/ui/tests/lib/theme.test.ts`** — added suites for (1) preset completeness (every preset defines every color token in light + dark, no unknown tokens), (2) `modern` ↔ `globals.css` sync (fails if the default palette drifts from the stylesheet — the "un-driftable" guarantee), and (3) per-preset shape/elevation (classic flat, neon cyan/purple/pink + rounder radius).
- **Changeset** (`ui`, minor) via the `pr-changeset` skill.

## Preset direction

| Preset | Direction |
| --- | --- |
| **modern** (default) | Restrained, Linear-class: low-chroma neutral surfaces, one saturated indigo brand color used sparingly, gradient pair as garnish, quiet muted text, soft radius. Inherits the stylesheet radius/shadows (kept byte-for-byte in sync). |
| **classic** | Flat, enterprise-safe: blue primary, **no gradient** (both stops collapse to the primary), squared-off radius (`0.375rem`), elevation removed (`--shadow-*: none`) so hierarchy comes from crisp borders. |
| **neon** | The high-chroma **cyan / purple / pink** personality preserved: pink primary, purple accent, a cyan→pink signature gradient, rounder radius (`0.75rem`). |

Each preset supplies light **and** dark values for every token, including the new `success` / `warning` (+foregrounds) and the gradient pair.

## HITL — design sign-off screenshots

Captured from the `examples/starter` admin (`/admin/post`) with **only** `ui.theme.preset` changed between runs (brand overrides removed), driven with Playwright, light + dark for all three presets. Files also live in `docs/theming/preset-review/` (review-only; can be dropped before merge).

### modern
| Light | Dark |
| --- | --- |
| modern light | modern dark |

### classic
| Light | Dark |
| --- | --- |
| classic light | classic dark |

### neon
| Light | Dark |
| --- | --- |
| neon light | neon dark |

> Screenshots are committed to the branch (embedded above via `?raw=true`) because binary attachment upload isn't available in the build environment. **Maintainer sign-off on the visual direction is still required before merge.**

## Acceptance criteria

- [x] All three presets define every token (light + dark); no token falls through to another preset (enforced by a new completeness test)
- [x] `modern` restrained; `classic` flat/gradient-free; `neon` high-chroma cyan/purple/pink identity preserved
- [x] Switching `preset` in the starter config visibly reskins the admin with no other changes (see screenshots)
- [x] Light/dark screenshots of all three presets attached; **maintainer sign-off pending**
- [x] Changeset for `ui` (minor)

## Verification

- `packages/ui` vitest: **254 passed** (theme suite: 39, incl. the new completeness/sync/shape tests)
- `tsc --noEmit` on `packages/ui`: **clean**
- `pnpm lint`: **0 errors** (pre-existing warnings only), `pnpm manypkg fix`, `pnpm format` run

---


borisno2 (commented):
## Code review — PR #718 (theme preset re-curation)

**Verdict: REQUEST CHANGES** (posted as a Comment because the review identity is the PR author, so GitHub blocks a formal Request-Changes). One blocking item; everything else passes.

_Scope: code correctness, the #705 compiler contract, and test enforcement. Visual/design direction is out of scope (deferred to the committed screenshots for HITL sign-off)._

### Overview
Re-curates the `modern` / `classic` / `neon` presets in the token vocabulary, extends `PresetDefinition` with optional `radius`/`shadows`, and teaches `compileTheme` to merge preset shape/elevation under config overrides. Adds completeness, sync, and shape/elevation test suites plus a `ui` minor changeset.

### 🔴 Blocking: the `modern ↔ globals.css` sync test does not check values — only token names
`packages/ui/tests/lib/theme.test.ts` (`modern preset ↔ stylesheet sync`). Both helpers capture only the **variable-name** group and discard the value:

```js
const matches = globalsCss.matchAll(new RegExp(`--color-([a-z-]+)-${suffix}\\s*:`, 'g'))
return [...matches].map((m) => `--color-${m[1]}-${suffix}`).sort()   // names only
```

So the assertion `declaredRawVars(suffix)).toEqual(compiledVars(suffix))` proves the **same set of tokens** is declared in both places — it does **not** prove the values agree. The PR description / docstring claim the palette is "kept byte-for-byte in sync" and "This test fails if they drift", but a value change on either side (e.g. re-tuning `--color-primary-light` in `globals.css` without updating `presetThemes.modern`) would leave this test green. That value drift is exactly the failure mode the "un-driftable" guarantee promises to prevent, so the guard doesn't deliver its headline job.

The values happen to match today (globals.css L25–76 == the `modern` preset), so this is a test-strength gap, not a live bug. Fix: assert the value map, not just the key set — parse `--color-<name>-<suffix>: <value>;` from `globals.css` into `{ name: value }` and compare against the same map extracted from `compileTheme({ preset: 'modern' })`. That makes the guard mean what the docstring says. (The structural drift it currently catches — added/removed/renamed tokens — is worth keeping; just extend it to values.)

### ✅ Verified / passing
- **#705 compiler contract preserved.** `modern` omits `radius`/`shadows`, so `radius = theme.radius ?? base.radius` stays `undefined` and `shadows = { ...base.shadows, ...theme.shadows }` collapses to `{}` → no `--radius`/`--shadow-*` emitted for `compileTheme({})`. The existing radius/shadow/font "only when provided" tests remain valid, and the new `modern inherits stylesheet shape and elevation (emits neither)` test locks it in. Merge is per-key, so a partial `theme.shadows` override on `classic` still keeps the preset's other stops. Emission stays verbatim/pure.
- **No `any`/casts introduced.** New `PresetDefinition` type and `const shadows: ThemeShadows = …`; spreading a possibly-`undefined` `base.shadows` is legal TS and needs no cast.
- **Preset token-set completeness is enforced.** `COLOR_TOKENS` (25 tokens) matches `ThemeColors` and the `globals.css` vocabulary. The completeness suite checks both directions (every token present via `toBeTypeOf('string')` **and** no unknown tokens via sorted-key equality) for all three presets in light + dark, including `success`/`warning` (+foregrounds) and the gradient pair. A new token added to `globals.css` would fail the sync (name) test, forcing `COLOR_TOKENS` and all presets to update — so the token set is pinned to the stylesheet transitively.
- **Preset values are valid CSS.** All `modern` values are `oklch(...)`, all `classic`/`neon` values are wrapped `hsl(...)`; no bare HSL triplets, so nothing trips the dev-mode warning (which only inspects developer overrides anyway).
- **`classic` flat / `neon` identity** are asserted (shadows `none`, radius `0.375rem`, collapsed gradient; neon pink primary / purple accent / cyan→pink gradient / `0.75rem`).
- **ESM `.js` import extensions** correct (`../../src/lib/theme.js`); type-only import from `@opensaas/stack-core/internal`.
- **Changeset** present and correct (`@opensaas/stack-ui`, minor).

### Minor (non-blocking)
- The sync test relies on `import.meta.dirname` (Node ≥ 20.11). Fine given the toolchain, just noting the floor.

Net: the implementation and compiler contract are correct and the completeness enforcement is solid; the one change I'd ask for before merge is tightening the modern↔stylesheet guard to compare **values**, so it actually backs the byte-for-byte claim.

---

PR #726, opened 2026-07-18
Make the modern preset baseline un-driftable (single-source stylesheet defaults)

### LINKED ISSUE #714: Design system follow-up: make the default `modern` baseline un-driftable (single source for stylesheet defaults + preset)
## Context

Follow-up from review of #713 (implements #705, part of PRD #704).

The config→stylesheet drift the token contract set out to eliminate is genuinely closed: `ui.theme` overrides write onto the same `--color-*`/`--font-*`/`--radius`/`--shadow-*` tokens the stylesheet declares. This issue is about a **separate, secondary** drift surface the reviewer flagged.

## Problem

The default `modern` preset values are currently hand-maintained in **two** places, reconciled only by "keep in sync" comments:

- `presetThemes.modern` in `packages/ui/src/lib/theme.ts`
- the raw `--color-*` values in `packages/ui/src/styles/globals.css` `:root`

Runtime consequence: when `ui.theme` is set at all, `AdminUI` injects `compileTheme` output that overrides every `--color-*-light/-dark` with `theme.ts`'s numbers; when no theme is set, `globals.css`'s numbers win. If the two copies ever drift, **identical-looking configs render differently depending on whether `ui.theme` is present**.

## Suggested direction

Generate the stylesheet's default token values from `presetThemes.modern` (or a shared single source) so the two can never diverge. Keep it a pure build/codegen step consistent with ADR-0015 (CSS variables remain the contract).

## Also fold in (cosmetic, from the same review)

- `specs/THEMING.md:116` still references `generateThemeCSS` as an "old internal detail that may change" — now fully removed; tighten the wording.

## Out of scope

No change to the public `ThemeConfig` API or token vocabulary — this is an internal de-duplication of the default baseline.

borisno2:
Closes #714. Follow-up to #704.

## Problem

The default `modern` palette lived in **two** hand-maintained copies, reconciled only by a "keep in sync" test that *caught* drift after the fact rather than *preventing* it:

- `presetThemes.modern` in `packages/ui/src/lib/theme.ts`
- the raw `--color-*-light`/`-dark` values in `packages/ui/src/styles/globals.css` `:root`

Runtime consequence flagged in the issue: when `ui.theme` is set, `AdminUI` injects `compileTheme` output (theme.ts's numbers); when it isn't, `globals.css`'s numbers win. If the two ever diverged, identical-looking configs would render differently depending on whether `ui.theme` was present.

## Approach — single source via codegen

`presetThemes.modern` is now the **single source of truth**; the stylesheet's baked-in color defaults are **derived** from it.

- New `packages/ui/scripts/generate-globals-css.mts` emits the `--color-*-light`/`-dark` `:root` block from `presetThemes.modern`, splicing it between stable `stack:generated-start/-end modern-preset-colors` marker comments. Everything outside the markers (`@theme` contract tokens, the radius knob, dark-mode mechanism, body styles) stays hand-written.
- Wired as the first step of the ui `build` (`npm run generate:css && tsc && npm run build:css`). Also exposed as `generate:css` / `generate:css:check` scripts.
- **No new dependencies**: runs on Node's native TypeScript type-stripping. `presetThemes` uses only `import type`, so importing `theme.ts` from the script pulls in no runtime deps.
- `--check` mode fails if the checked-in `globals.css` isn't byte-identical to a fresh regeneration.

## Tests

The `modern preset ↔ stylesheet sync` suite is strengthened: alongside the existing value-map comparison, a new test runs the real generator in `--check` mode and asserts it exits 0 — so drift is now impossible to commit without a failing test, not merely detectable. All theme tests pass (40), full ui suite passes (367 node + 69 browser).

## Proof of no visual change

The `--color-*` values in `globals.css` are **unchanged** — the diff adds only marker comments and trims the two "kept in sync" comments. `git diff` of the color-value lines is empty, and the generator's `--check` confirms the committed CSS is byte-identical to what the single source produces. No theme (`modern`/`classic`/`neon`) changes appearance.

## Also folded in (from the same review)

- Tightened the stale `generateThemeCSS` wording in `specs/THEMING.md:116` (that internal helper is fully removed) and documented that the `--color-*` defaults are now generated from `presetThemes.modern`.

## Out of scope

No change to the public `ThemeConfig` API or token vocabulary — this is purely an internal de-duplication of the default baseline.

---

### DIFF
--- a/.changeset/tidy-preset-baseline.md
+++ b/.changeset/tidy-preset-baseline.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+Derive the stylesheet's `modern` color defaults from `presetThemes.modern` via a `generate:css` codegen step (wired into build), so the preset and `globals.css` can no longer drift. No visual or behavioral change — the emitted CSS is byte-identical to the previous values.
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -33,7 +33,9 @@
     "./styles": "./dist/styles/globals.css"
   },
   "scripts": {
-    "build": "tsc && npm run build:css",
+    "build": "npm run generate:css && tsc && npm run build:css",
+    "generate:css": "node --experimental-strip-types scripts/generate-globals-css.mts",
+    "generate:css:check": "node --experimental-strip-types scripts/generate-globals-css.mts --check",
     "build:css": "mkdir -p dist/styles && postcss ./src/styles/globals.css -o ./dist/styles/globals.css",
     "dev": "tsc --watch",
     "test": "vitest",
--- a/packages/ui/scripts/generate-globals-css.mts
+++ b/packages/ui/scripts/generate-globals-css.mts
@@ -1,0 +1,100 @@
+/*
+ * Codegen: emit the baked-in `modern`-preset color defaults in
+ * `src/styles/globals.css` from the single source of truth — `presetThemes.modern`
+ * in `src/lib/theme.ts`.
+ *
+ * Why: the default `modern` baseline used to live in two hand-maintained copies
+ * (the preset catalog AND the raw `--color-*-light`/`-dark` `:root` vars),
+ * reconciled only by a test that CAUGHT drift after the fact. This script makes
+ * the stylesheet's copy DERIVED, so the two can never diverge (issue #714).
+ *
+ * The generated region is delimited by the `stack:generated-*` marker comments
+ * in globals.css; everything outside them (the `@theme` contract tokens, the
+ * radius knob, the dark-mode mechanism, body styles) stays hand-written.
+ *
+ * Run directly (`pnpm --filter @opensaas/stack-ui generate:css`) or as the first
+ * step of `build`. `--check` mode verifies the checked-in file is already in sync
+ * (used by the test suite) instead of writing.
+ *
+ * Runs on Node's native TypeScript type-stripping — `presetThemes` uses only
+ * `import type`, so importing `theme.ts` here pulls in no runtime dependencies.
+ */
+import { readFileSync, writeFileSync } from 'node:fs'
+import { fileURLToPath } from 'node:url'
+import { dirname, join } from 'node:path'
+import { presetThemes } from '../src/lib/theme.ts'
+import type { ThemeColors } from '@opensaas/stack-core/internal'
+
+const START_MARKER = '  /* stack:generated-start modern-preset-colors */'
+const END_MARKER = '  /* stack:generated-end modern-preset-colors */'
+
+/** camelCase token key → kebab-case CSS variable segment (matches lib/theme.ts). */
+function toKebabCase(key: string): string {
+  return key.replace(/([A-Z])/g, '-$1').toLowerCase()
+}
+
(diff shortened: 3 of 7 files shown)


borisno2 (commented):
## Review — Verdict: **APPROVE** ✅

(Submitted as a Comment review because GitHub blocks a formal Approve when the integration identity authored the PR. The verdict is APPROVE — no changes required.)

Clean, well-tested build-tooling refactor. I reconstructed the PR's file layout and ran the generator locally to verify the claims empirically, plus checked CI wiring, prettier consistency, and the published-package path. Everything holds up.

### What it does
Makes `presetThemes.modern` the single source of truth for the baked-in `--color-*-light/-dark` `:root` defaults in `globals.css`. New `scripts/generate-globals-css.mts` emits that block between `stack:generated-start/-end` markers, wired as the first ui `build` step, with a `--check` mode the test suite asserts exits 0.

### Verification performed (all passed)
- **Byte-identical / no visual change** — ran the generator in `--check` against the committed `globals.css`: **exit 0**. Every `--color-*` value is unchanged; the diff is only comments + markers.
- **Idempotency** — write mode then re-run produces **no diff**.
- **Marker safety** — with a marker removed, `render()` throws a clear error and does **not** corrupt the file.
- **Drift guard genuinely fails** — tampered one preset value (`background`) and `--check` correctly **exited 1**. The guard does what it claims.
- **CI Node / type-stripping** — CI uses **Node 24** (`.github/actions/setup`, `node-version: '24'`), where TS type-stripping is on by default and importing a non-entry `.ts` file works. I reproduced the run on Node 22 in a dir with **no `node_modules`** and it succeeded — proving `theme.ts`'s `import type { ThemeColors } from '@opensaas/stack-core/internal'` is fully erased at runtime (no runtime dep / import cycle). No uninstalled deps.
- **Generator runs in CI even for tests** — turbo `test` `dependsOn: ["build"]`, and `build` = `generate:css && tsc && build:css`, so the generator executes in both the `test` and `e2e` jobs.
- **Published package never stale** — `build:css` compiles `src → dist/styles/globals.css` **after** `generate:css`, and `src` is committed already in-sync, so a consumer can't receive an ungenerated stylesheet.
- **Prettier consistency** — `globals.css` is in prettier's scope; running `prettier --write` over the generated block yields **no diff**. Generator formatting matches prettier, so `format:check` and `--check` won't oscillate.
- **Conventions** — no `any`/casts (fully typed helpers), ESM `.mts`, `@opensaas/stack-ui` **patch** changeset present. Patch is appropriate for a no-visual-change refactor.

### Minor observations (non-blocking)
- `build` mutates a **tracked source file** (`src/styles/globals.css`), not just `dist/`. Acceptable here since output is deterministic/idempotent and committed in-sync — the tree only goes "dirty" precisely when `theme.ts` changed without regenerating, which is the intended signal. A one-line note in the script header would help contributors who are surprised by it.
- The generator's output file isn't a declared turbo `build` output, so on a build cache hit `generate:css` is skipped — harmless, because a `theme.ts` edit busts the build input hash and the always-run (`cache: false`) `--check` test is the real enforcement layer.

Nice work making the baseline un-driftable.

---

Issue #729, opened 2026-07-19
Admin UI: Cell registry with default cells for core field types


borisno2:
## Parent

Spec: #728

## What to build

List tables render every value through a **Cell** resolved by a cell-component registry that mirrors the form-field registry's priority chain: per-field override → custom type registry → field-type registry → plain-text fallback. Core field types ship default cells: text (plain), integer (right-aligned, tabular figures), select (Badge, colored by new option-level UI metadata `{ value, label, ui: { variant } }`, neutral when unmapped), timestamp (formatted date), checkbox (mark), to-one relationship (Item label). A developer can override any field's cell with their own component, and a third-party field package can register a cell exactly like it registers a form component. An admin viewing a list with a select status field sees colored status badges instead of raw values.

## Acceptance criteria

- [ ] Cells resolve via the documented priority chain; a per-field override component wins over registries
- [ ] Option-level badge-variant metadata (additive to existing option shape) renders in select cells; unmapped options get the neutral badge
- [ ] Integer/timestamp/checkbox/to-one cells render per spec; unknown/third-party types without a registered cell fall back to plain text
- [ ] Existing list pages render identically-or-better with zero config changes
- [ ] Component tests cover the resolution priority; e2e asserts badge rendering on a select column
- [ ] New components consume only Theme tokens and ship named Slots; changeset included

## Blocked by

None — can start immediately.

Issue #728, opened 2026-07-19
Spec: Generic reusable admin UI — filter builder, bulk actions, rich cells, relationship tables


borisno2:
# Generic reusable admin UI

Source design: Claude Design project "Generic reusable admin UI" (https://claude.ai/design/p/a295857b-9a5e-4569-810c-a830036f5d84?file=Admin+UI.dc.html). Decision record: grilling session on branch `claude/admin-ui-design-review-rdkzar` — glossary terms in `CONTEXT.md` (Filter builder, Filter token, Filter spec, Bulk action, Cell, Relationship table, Inline cell edit) and ADR-0017 / ADR-0018.

## Problem Statement

Admins running stack-built applications can only browse their data through a bare table: plain text cells, a single contains-search over text fields, column sort, and pagination. Narrowing a list precisely ("active editors with more than 5 orders"), acting on many records at once, or working with a record's related rows (a user's orders) forces them out of the admin UI into ad-hoc pages or the database. The approved design solves this — but it was drawn against a specific Users/Orders example, and the stack's admin UI must deliver it for **any** list a project configures, with zero required config, without ever bypassing the access-control engine.

## Solution

Evolve `@opensaas/stack-ui` in place (no v2 fork). The list view gains a Filter builder (tokenized, URL-backed, server-executed queries with structural autocomplete), row selection with Bulk actions, and rich field-type-aware Cells. The Admin chrome gains opt-in per-list nav counts. The item view becomes a shape-derived layout: scalar/to-one fields in a details card with whole-form save, and each to-many relationship as a Relationship table with per-cell inline editing, a pre-linked create drawer, and disconnect-by-default row removal. Every capability is driven by the field/list config and extensible through the same registry pattern the form components already use, so third-party fields participate without core changes.

## User Stories

1. As an admin user, I want to type field-scoped tokens (e.g. `role:Editor status:Active`) into a single filter input, so that I can narrow a list precisely without a multi-widget filter UI.
2. As an admin user, I want autocomplete suggestions for filterable fields and their allowed values while I type, so that I can discover what's filterable without documentation.
3. As an admin user, I want comparison operators on numeric and date fields (`orders:>5`, `joined:>2024-01-01`), so that I can express ranges.
4. As an admin user, I want quoted values (`name:"Ada Lovelace"`), so that multi-word values filter correctly.
5. As an admin user, I want bare words to search as free text across text fields, so that quick lookups stay one keystroke away.
[...]

PR #745, opened 2026-07-19
Admin UI: Cell registry with default cells for core field types

borisno2:
Implements #729
Part of #728

## What this does

List tables now render every value through a **Cell** resolved by a cell-component registry that **exactly mirrors the form-field registry's** resolution chain and registration mechanism:

1. **per-field override** — `field.ui.cell` (mirrors `ui.component` for forms)
2. **custom type registry** — `field.ui.fieldType` (shared with forms)
3. **field-type registry** — `field.type`
4. **plain-text fallback** — `TextCell` for unknown/third-party types

`registerCellComponent(type, Cell)` is the peer of `registerFieldComponent`, and `CellRenderer` is the peer of `FieldRenderer` (same inner-component pattern so the resolution stays out of the render scope). There is **no switch on field type** anywhere — each field type provides its own default Cell and the renderer delegates, per the field-builder self-containment rule.

### Default cells (core field types)

- **text** — plain text (also the fallback for unregistered types)
- **integer** — tabular figures (right alignment stays on the numeric column)
- **select** — `Badge`, coloured by the new option-level UI metadata; neutral badge when unmapped
- **timestamp** — formatted local date/time
- **checkbox** — accessible mark (`Yes`/`No` preserved as the accessible name)
- **relationship (to-one)** — Item label link; arrays keep the existing comma-separated links

All new components consume **only Theme tokens** and ship named **Slots** (`data-slot="cell-*"`).

### Additive option-shape change (core)

`SelectOption` gains an optional, additive `ui.variant`:

```typescript
status: select({
  options: [
    { label: 'Draft', value: 'draft', ui: { variant: 'secondary' } },
    { label: 'Published', value: 'published', ui: { variant: 'success' } },
  ],
})
```

This is purely additive to the shared `{ value, label }` shape — existing consumers are untouched, options without `ui.variant` render the neutral badge, and unmapped values still render. `SelectOptionVariant` mirrors the UI `Badge` primitive's variants and is exported from `@opensaas/stack-core/fields`.

## How existing pages stay working

`ListView` (server) now passes serialized field configs to `ListViewClient`, which routes each cell through `CellRenderer`. `ListViewClient` synthesizes a minimal field config from `fieldTypes` when no serialized `fields` prop is supplied, so all existing call sites and tests render unchanged. `ListTable` routes scalar cells through the registry while keeping its own relationship rendering (`stopPropagation` for row clicks). Zero config changes required.

## Tests

- **Component** — `CellRenderer.test.tsx` locks down the four-step resolution priority (override wins over both registries; custom-type registry; field-type registry; plain-text fallback). `Cells.test.tsx` covers select badge variants (mapped `success`, neutral fallback for both unmapped-option and unknown-value), integer tabular figures, timestamp formatting, checkbox marks, and relationship link resolution.
- **e2e** — `03-admin-ui.spec.ts` asserts the select column renders a `data-slot="cell-select"` Badge with the `text-success` token class for a published post.
- Existing UI (387) and core (774) suites pass unchanged (two checkbox assertions updated from visible `Yes`/`No` text to the accessible mark's name).

## Verification

- `pnpm --filter @opensaas/stack-core build` + `@opensaas/stack-ui build` — pass (tsc clean)
- core vitest 774 passed; ui vitest 387 passed (+ new cell tests)
- `pnpm lint` — 0 errors (3 pre-existing warnings in untouched core files); `pnpm manypkg fix`; `pnpm format`
- Changeset included (`minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`)

---

### DIFF
--- a/.changeset/rich-cells-registry.md
+++ b/.changeset/rich-cells-registry.md
@@ -1,0 +1,40 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Add a Cell registry with default cells for core field types
+
+List tables now render every value through a **Cell** resolved by a
+cell-component registry that mirrors the form-field registry's priority chain:
+per-field override → custom type registry → field-type registry → plain-text
+fallback. Each core field type ships a default Cell — text (plain), integer
+(tabular figures), select (coloured Badge), timestamp (formatted date), checkbox
+(mark), and to-one relationship (Item label link). Unknown/third-party types
+without a registered Cell fall back to plain text.
+
+Select options gain optional, additive per-option UI metadata mapping a value to
+a badge variant. Existing options keep working unchanged; unmapped options render
+the neutral badge.
+
+```typescript
+// opensaas.config.ts — colour a status value in list-table cells
+status: select({
+  options: [
+    { label: 'Draft', value: 'draft', ui: { variant: 'secondary' } },
+    { label: 'Published', value: 'published', ui: { variant: 'success' } },
+  ],
+})
+```
+
+Register a Cell for a custom/third-party field exactly as you register its form
+component, or override a single field's Cell:
+
+```typescript
+'use client'
+import { registerCellComponent } from '@opensaas/stack-ui'
+registerCellComponent('myField', MyCell)
+
+// or per-field override (highest priority)
+price: integer({ ui: { cell: CurrencyCell } })
+```
--- a/e2e/starter-auth/03-admin-ui.spec.ts
+++ b/e2e/starter-auth/03-admin-ui.spec.ts
@@ -110,6 +110,32 @@ test.describe('Admin UI', () => {
         page.getByRole('row', { name: /Full Post/ }).getByRole('cell', { name: 'published' }),
       ).toBeVisible()
     })
+
+    test('should render the select column value as a coloured status badge', async ({ page }) => {
+      // Create a published post so the status column has a value to badge.
+      await page.goto('/admin/post')
+      await page.waitForLoadState('networkidle')
+
+      await page.click('text=/create|new/i')
+      await page.waitForLoadState('networkidle')
+      await page.fill('input[name="title"]', 'Badged Post')
+      await page.fill('input[name="slug"]', 'badged-post')
+      await page.fill('textarea[name="content"]', 'Content here')
+      await page.getByLabel('Status').click()
+      await page.getByRole('option', { name: 'published' }).click()
+      await page.click('button[type="submit"]')
+      await page.waitForURL(/admin\/post/, { timeout: 10000 })
+
+      // The status cell renders through the select Cell → a Badge (Slot
+      // `cell-select`) coloured by the option's `ui.variant` (success → the
+      // success token class), not raw text.
+      const badge = page
+        .getByRole('row', { name: /Badged Post/ })
+        .locator('[data-slot="cell-select"]')
+      await expect(badge).toBeVisible()
+      await expect(badge).toHaveText('Published')
+      await expect(badge).toHaveClass(/text-success/)
+    })
   })
 
   test.describe('Create Form', () => {
--- a/examples/starter-auth/opensaas.config.ts
+++ b/examples/starter-auth/opensaas.config.ts
@@ -127,8 +127,8 @@ export default config({
         }),
         status: select({
           options: [
-            { label: 'Draft', value: 'draft' },
-            { label: 'Published', value: 'published' },
+            { label: 'Draft', value: 'draft', ui: { variant: 'secondary' } },
+            { label: 'Published', value: 'published', ui: { variant: 'success' } },
           ],
           defaultValue: 'draft',
           ui: { displayMode: 'segmented-control' },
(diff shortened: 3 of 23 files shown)


borisno2 (commented):
## Review: Cell registry with default cells for core field types

**VERDICT: APPROVE** — minor, non-blocking suggestions only. (Submitted as a Comment because GitHub blocks a formal Approve when the same identity authored the PR.) Reviewed strictly from the PR diff plus the files it depends on but does not change (`Badge` primitive, `FieldRenderer`, `getFieldDisplayValue`), verified against head sha `40e5fce`.

### Overview
Introduces a Cell-component registry (`registerCellComponent` / `getCellComponent` / `CellRenderer`) mirroring the form-field registry and its resolution chain, plus default Cells for text/integer/select/timestamp/checkbox/relationship. Adds an additive, optional `ui.variant` to `SelectOption` in core. `ListView` now serialises field configs to `ListViewClient`, which routes every cell through `CellRenderer`. Well-scoped, well-tested, consistent with the field-builder self-containment rule.

### Scrutinised risk areas — all clear
- **Additive option-shape change** — `SelectOption` is a strict superset (`{label,value}` + optional `ui`); existing consumers (`buildFallbackUIProps` → `SelectField`) still compile and behave identically. `SelectCell` renders the neutral (`secondary`) badge for both unmapped options and values absent from the option list, and a dash for empty — no crash, no missing render. Covered by `Cells.test.tsx`.
- **No `switch (field.type)`** — `CellRenderer` resolves via `field.ui.cell || (ui.fieldType ? get(fieldType) : get(type)) || TextCell`, a line-for-line mirror of `FieldRenderer` (including "fieldType set but unregistered → fallback, not field.type"). No type switch in core or the registry.
- **Prop serialisability** — `CellComponentProps` carries only `value` (JSON), serialised `field`, `fieldName`, `basePath` — consistent with the existing serialised-config path used by the forms.
- **Type safety** — no `any` in the new files; `CellComponent = ComponentType<CellComponentProps>` is fully typed; `SelectOptionVariant` is a proper named union exported from `@opensaas/stack-core/fields` and exactly matches the `Badge` primitive's variants (`default|secondary|success|warning|destructive|outline`). Internal casts only, same style as surrounding code.
- **Zero-config regression** — `ListViewClient` synthesises a minimal `{type, ref?}` config when `fields` is absent, so legacy call sites render via the field-type registry. Existing pages render identically-or-better (select→badge, integer→tabular, checkbox→accessible mark).
- **Changeset** — present and correctly scoped: `minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`.

### Findings (by severity)

**MEDIUM — `ListTable` (standalone) diverges from the new path** (`packages/ui/src/components/standalone/ListTable.tsx`)
It retains `fieldTypes[column] === 'relationship' ? renderRelationshipCell(...) : <CellRenderer field={{ type: fieldTypes[column] }} .../>`. Two consequences:
1. It keeps the old duplicate relationship renderer even though `RelationshipCell` already performs the `stopPropagation` that was the stated reason to keep it — a residual field-type conditional + duplicated logic remain in `ListTable` while `ListView` is fully registry-driven.
2. It passes `field={{ type }}` with **no `options`**, so a `select` column in a standalone `ListTable` renders the raw value in a neutral badge with no option-label mapping and no `ui.variant` colour. Parity-or-better vs the prior raw-text output (not a regression), but inconsistent with `ListView`. Consider routing `ListTable` through `serializeFieldConfig` too, or documenting the gap.

**LOW — "Strips out functions" comment now inaccurate** (`packages/ui/src/lib/serializeFieldConfig.ts`)
`config.ui` is a spread of the field's `ui` (minus `valueForClientSerialization`), so `ui.component` and the new `ui.cell` functions are retained. Fine for client-component references (RSC allows those across the boundary) and mirrors the existing `component` handling — but a plain server-defined `cell`/`component` would throw at the `ListView` server→client boundary. Worth a doc note that `ui.cell` must be a client-component reference, and softening the "Strips out functions" comment.

**LOW — `CheckboxCell` renders null/undefined as "No"** (`packages/ui/src/components/cells/CheckboxCell.tsx`)
`if (value)` treats `null`/`undefined` like `false` and shows the "No" mark. The prior `getFieldDisplayValue` returned a neutral `-`. Minor semantic shift (null ≠ false); consider a dash for null/undefined and reserving the mark for actual booleans.

**LOW — `CellComponentProps.value: unknown` is an exported cross-package type** (`packages/ui/src/components/cells/registry.ts`)
CLAUDE.md discourages exposing `unknown` as an external type. `unknown` is the type-safe choice here (forces narrowing per cell) and mirrors the already-public `FieldRendererProps.value: unknown`, so a borderline nit rather than a defect — flagging for consistency with the stated convention.

**NIT** — `getFieldDisplayValue` is no longer used by `ListViewClient`/`ListTable` (verify no dead export). The type-only `registry.ts ↔ serializeFieldConfig.ts` import cycle is harmless (erased at compile).

### Tests
Good coverage: `CellRenderer.test.tsx` locks all four resolution steps; `Cells.test.tsx` covers mapped/neutral/unknown select variants, tabular integer, timestamp formatting, checkbox accessible marks, and relationship link + raw-record fallback; the e2e asserts the `cell-select` badge with `text-success`. Two checkbox assertions correctly updated to the accessible-name query.

Nothing here blocks merge. Faithful mirroring of the form-field registry.

_Reviewed by Claude Code._

---

Issue #730, opened 2026-07-19
Admin UI: Filter engine — Filter spec contract and URL-driven server filtering


borisno2:
## Parent

Spec: #728

## What to build

An admin can put a filter query in the list URL (e.g. `role:Editor status:Active orders:>5 name:"Ada Lovelace" beta`) and the list shows exactly the matching, access-scoped rows — filtering runs server-side through the secured context. Fields participate by declaring a **Filter spec** via a new optional method on the field-builder contract (peer of the existing generation methods): supported operators, a server-side token→condition mapper, and serializable suggestion metadata. Core field types implement it (text contains, select/checkbox equality against enumerated values, integer/timestamp comparisons, relationship by label lookup); a field without a Filter spec is not filterable. Bare words search as free text across text-searchable fields, replacing the current hard-coded text-type search. Grammar per ADR-0017: implicit AND, quoted values, comparisons; unknown syntax degrades to free text, never errors.

This ticket creates the one new test seam: pure `parse(queryString) → Filter tokens` and `tokens + Filter specs → query conditions`.

## Acceptance criteria

- [ ] Filter spec is an optional, additive method on the field contract; all core field types implement it
- [ ] URL query param filters the list server-side, merged with access control via the secured context — never client-side
- [ ] ADR-0017 grammar: implicit AND, quoted multi-word values, numeric/date comparison operators, bare-word free text; malformed tokens degrade to free text
- [ ] Hard-coded `type === 'text'` search removed; free-text behavior driven by Filter specs
- [ ] Unit tests lock down the parse and mapping seam, including graceful degradation
- [ ] E2e: a URL-typed query returns the filtered, access-scoped table; changeset included (core + ui)

## Blocked by

None — can start immediately.

PR #746, opened 2026-07-19
Filter engine: Filter spec contract and URL-driven server filtering

borisno2:
Implements #730. Part of #728.

## What this adds

The admin UI's **filter engine** (ADR-0017). An admin can put a filter query in the list URL (e.g. `role:Editor status:Active orders:>5 name:"Ada Lovelace" beta`) and the list shows exactly the matching, access-scoped rows — filtering runs server-side through the secured context.

### Filter spec — an additive field-builder method
Fields declare their filtering capability through a new **optional** `getFilterSpec` method on the field-builder contract, a peer of `getPrismaType`/`getTypeScriptType`. A spec reports:
- the **operators** it supports,
- a pure **token → condition** mapper, and
- serializable **suggestion** metadata (enumerated values for select/checkbox, relationship label search, structure-only for unbounded fields).

All core field types implement it: text (`contains` + free text), integer/decimal/timestamp/calendarDay (comparisons), select/checkbox (equality against enumerated values), relationship (nested label lookup). A field **without** a spec — `password`, `json`, `virtual`, or any third-party field that hasn't adopted one — is simply not filterable, so the addition degrades gracefully everywhere. `FilterSpec` is exported from `@opensaas/stack-core/extend` for third-party field authors.

### The pure seam (unit-tested)
Kept free of any DB/Prisma imports:
- `parseFilterQuery(query) → FilterToken[]`
- `buildFilterWhere(tokens, specs) → where`

Grammar per ADR-0017: implicit AND, quoted multi-word values, `>`/`>=`/`<`/`<=` comparisons, bare-word free text. **Unknown syntax degrades to free text, never errors** (unknown field, unsupported operator, or unmappable value all fall back to a free-text search across text-searchable fields).

### Merged through the secured context
`ListView` composes the two steps via `buildListFilterWhere` and hands the `where` to `context.db.<list>.findMany`/`count`. The access engine ANDs it with the operation-level access filter, so the filter UI can only ever **narrow**, never widen, visibility. The hard-coded `type === 'text'` search is removed; free-text behavior is now driven by each text field's Filter spec (preserving prior behavior).

## Acceptance criteria
- [x] Filter spec is an optional, additive method on the field contract; all core field types implement it
- [x] URL query param filters the list server-side, merged with access control via the secured context — never client-side
- [x] ADR-0017 grammar: implicit AND, quoted values, numeric/date comparisons, bare-word free text; malformed tokens degrade to free text
- [x] Hard-coded `type === 'text'` search removed; free-text driven by Filter specs
- [x] Unit tests lock down the parse and mapping seam, including graceful degradation
- [x] E2e: a URL-typed query returns the filtered, access-scoped table; changeset included (core + ui)

## Verification
- `pnpm build` (full workspace): green
- core vitest: **805 passed** (incl. new `src/filter/filter.test.ts` — parser, mapper, per-field specs, collect helpers, degradation)
- ui vitest: **374 passed** (incl. new `ListView` tests asserting the filter `where` reaches the secured `context.db`)
- e2e `05-filter.spec.ts`: **3 passed** against the real starter-auth app (free text, select token, unknown-syntax degradation)
- `pnpm lint` (0 errors), `pnpm manypkg fix`, `pnpm format`
- changeset: `minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`

---

### DIFF
--- a/.changeset/filter-engine-filter-spec.md
+++ b/.changeset/filter-engine-filter-spec.md
@@ -1,0 +1,30 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Add the admin UI filter engine: a Filter spec field-builder contract and URL-driven server-side list filtering (ADR-0017).
+
+Fields now declare their filtering capability through a new optional `getFilterSpec` method — a peer of `getPrismaType`/`getTypeScriptType` on the field-builder contract. It reports the operators a field supports, a pure token→condition mapper, and serializable suggestion metadata. Core field types implement it (text contains + free text, integer/decimal/timestamp/calendarDay comparisons, select/checkbox equality against enumerated values, relationship by label lookup). A field without a spec — `password`, `json`, `virtual`, or any third-party field that hasn't adopted one — is simply not filterable, so the addition degrades gracefully everywhere.
+
+The admin list view now parses the URL filter query (the list's `search` param) through the engine and merges the result into the access-controlled query via the secured context, so filtering runs server-side and can only ever narrow — never widen — what a session may see. This replaces the previous hard-coded `type === 'text'` search; free-text behavior is now driven by each text field's Filter spec.
+
+Grammar (ADR-0017): implicit-AND tokens, quoted multi-word values, `>`/`>=`/`<`/`<=` comparisons on numeric/date fields, and bare words as free text. Unknown syntax degrades to free text, never errors.
+
+New exports from `@opensaas/stack-core`:
+
+```typescript
+import {
+  parseFilterQuery, // (query) => FilterToken[]  — pure
+  buildFilterWhere, // (tokens, specs) => where   — pure
+  collectFilterSpecs, // (listConfig, listKey, config) => specs
+  buildListFilterWhere, // (query, listConfig, listKey, config) => where
+  collectFilterSuggestions, // serializable autocomplete metadata
+} from '@opensaas/stack-core'
+
+// e.g. "status:Published views:>10 author:\"Ada Lovelace\" beta"
+const where = buildListFilterWhere(query, listConfig, listKey, config)
+const rows = await context.db.post.findMany({ where }) // ANDed with the access filter
+```
+
+Third-party field authors can implement `FilterSpec` (exported from `@opensaas/stack-core/extend`) to make their field filterable.
--- a/.changeset/filter-engine-filter-spec.md
+++ b/.changeset/filter-engine-filter-spec.md
@@ -11,6 +11,8 @@ The admin list view now parses the URL filter query (the list's `search` param) 
 
 Grammar (ADR-0017): implicit-AND tokens, quoted multi-word values, `>`/`>=`/`<`/`<=` comparisons on numeric/date fields, and bare words as free text. Unknown syntax degrades to free text, never errors.
 
+Multi-word free-text UX shift (intentional, per ADR-0017): bare words now combine with AND, so `hello world` requires each word to match separately (not the literal substring `hello world`). To match a contiguous phrase, quote it: `"hello world"`. A pasted URL such as `http://x` is treated as a single free-text token and searched verbatim — the `http:` prefix is not parsed as a field.
+
 New exports from `@opensaas/stack-core`:
 
 ```typescript
--- a/e2e/starter-auth/05-filter.spec.ts
+++ b/e2e/starter-auth/05-filter.spec.ts
@@ -1,0 +1,81 @@
+import { test, expect, type Page } from '@playwright/test'
+import { signUp, generateTestUser } from '../utils/auth.js'
+
+/**
+ * Filter engine (issue #730 / ADR-0017).
+ *
+ * A URL-typed filter query narrows the list server-side through the secured
+ * context: field-scoped tokens, bare-word free text, and graceful degradation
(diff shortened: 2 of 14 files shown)


borisno2 (commented):
## Review — Filter engine (ADR-0017)

**Verdict: APPROVE** (posted as a Comment review; a formal Approve is blocked because the review identity is the PR author). Solid, well-factored, well-tested. The security-critical seams hold up. A few non-blocking findings below, one of which contradicts the code's own docstring.

### Overview
Adds a URL-driven, server-side list filter engine: an optional additive `getFilterSpec` field-builder method, a pure `parse → tokens → where` seam, config-aware collect helpers, and `ListView` wiring that hands the resulting `where` to the secured context. Also removes the hard-coded `type === 'text'` list search.

### Risk-area verification (all pass)
- **Merges through the secured context ✅** — `buildListFilterWhere` returns a plain `where` fragment; `ListView` passes it to `dbContext[key].findMany/count` where `dbContext = context.db`. There is no path to raw Prisma; the access engine ANDs it, so the filter can only narrow. Field names in conditions come from the field's *own* closed-over config key, and user-supplied `token.field` is only ever used as a lookup key into `specs` (unknown → degrade), so no raw `where`/operator injection (`AND`/`OR`/arbitrary field) is possible.
- **Pure seam isolation ✅** — `parse.ts` and `map.ts` import types only, no DB/Prisma. Grammar (implicit AND, quoted multi-word, `>`/`>=`/`<`/`<=`, bare-word free text) present; parser never throws; unknown field / unsupported operator / unmappable value all degrade to free text. Good coverage in `filter.test.ts`.
- **Additive/optional spec ✅** — `getFilterSpec?` is optional; `collectFilterSpecs` skips non-functions; `password`/`json`/`virtual` and un-adopted third-party fields simply aren't filterable. Relationship spec also returns `undefined` when the target label field is virtual — nice.
- **Hard-coded text search removed ✅** — the `type === 'text'` block is gone, replaced by text fields' `freeText` specs.
- **Changeset ✅** — `minor` for both `@opensaas/stack-core` and `@opensaas/stack-ui`, well described.

### Findings (non-blocking, by severity)

**1. (Low–Med, correctness) `field:` prefix regex swallows URL schemes — contradicts the docstring.** `FIELD_PREFIX_RE = /^([A-Za-z_][A-Za-z0-9_]*):/` matches `http:` in `http://x`, so `parseFilterQuery('http://x')` yields `{ field: 'http', operator: 'eq', value: '//x' }`, which then degrades to a free-text search for `//x` — **not** the whole `http://x`. The parser docstring explicitly claims `http://x` → "a bare free-text token (field: null)", which is inaccurate, and there is no test for it. Degrades gracefully (no crash), but a pasted URL is mangled. Consider only treating a prefix as a field when the field name is a known filterable key (or the value isn't a `//`-scheme), and fix/remove the docstring example.

**2. (Low–Med, behavior change) Multi-word free text is now word-AND, not substring.** Old search: one `OR` of `contains(<whole string>)`. New: each whitespace-separated word becomes its own OR-across-free-text-fields clause, AND-ed together — so `hello world` now requires both words separately rather than the substring "hello world" (which now needs quoting: `"hello world"`). This is a reasonable ADR-0017 design and arguably an improvement, but the PR/changeset claim of "preserving prior behavior" is overstated for multi-word queries. Worth a note in the changeset so the UX shift is intentional and documented.

**3. (Low, security consideration) Nested relationship label filter is not access-scoped on the related list.** `{ author: { is: { name: { contains } } } }` runs as part of the Post query. Top-level Post access still applies, but the nested condition on `User` isn't filtered by the session's access to `User`. This is inherent to Prisma nested relation filtering (and matches Keystone), and only narrows Posts the session can already see, so the exposure is marginal — but it's an asymmetry worth acknowledging given the suggestions metadata is careful to route label *lookups* through the access-controlled path.

**4. (Low, type surface) `unknown`/`any` in exported types.** `FilterCondition = Record<string, unknown>` and `collectFilterSpecs(listConfig: ListConfig<any>, …)` are part of the public API. Both are pragmatic (opaque Prisma `where`; TypeInfo-agnostic list) and consistent with existing repo conventions (the old `where` was `Record<string, unknown>` too), so acceptable — flagging only against the repo's "no `unknown`/`any` as external types" guidance.

**5. (Nit) `collectFilterSuggestions` is exported and tested but not yet consumed** — expected, since this is "Part of #728" and the Filter builder UI lands later. No action needed.

Nice work: pure seam is genuinely pure, degradation is thorough, and the `ListView` tests assert the `where` actually reaches `context.db`. Recommend addressing #1 (or at least correcting the docstring) before merge; the rest are optional.

---

borisno2:
Pushed a scoped fix for the filter parser's `field:` prefix bug (`1a72f4c`): the parser no longer treats `word:` as a field prefix when the colon is immediately followed by `//`, so a pasted URL like `http://x` / `https://…` stays a single free-text token and is searched verbatim instead of being mangled into `{ field: 'http', value: '//x' }`. Kept the parser pure and field-agnostic. Also corrected the inaccurate `parse.ts` docstring example, added unit tests (parse + where-clause + never-throws for the URL cases), and noted the intentional multi-word free-text word-AND UX shift (ADR-0017) in the changeset. Core suite green (807 passed), core build clean, lint 0 errors.

---

Issue #736, opened 2026-07-19
Admin UI: custom bulk actions from list config


borisno2:
## Parent

Spec: #728

## What to build

A developer declares list-specific **Bulk actions** in the list's config — key, label, button variant, and a server-side handler receiving the selected ids and the secured context — and an admin sees those buttons in the selection bar next to the built-in Delete, gated on what the session may do. The handler runs via a server action that rebuilds the session context; the list refreshes and the outcome is reported when the action completes. CSV export ships as a documented recipe using this surface, not as a built-in.

## Acceptance criteria

- [ ] Config surface accepts multiple actions with label/variant; buttons render in declaration order in the selection bar
- [ ] Handlers execute server-side with the secured context and the explicit id set — never raw db access
- [ ] Outcome feedback on completion; selection clears; table refreshes
- [ ] An export recipe is documented as a custom action example
- [ ] E2e exercises a custom action end-to-end on an example list; changeset included

## Blocked by

- #733

Issue #733, opened 2026-07-19
Admin UI: row selection and built-in bulk delete


borisno2:
## Parent

Spec: #728

## What to build

An admin selects rows with per-row checkboxes; the header checkbox toggles the visible page; the selection (an explicit id set) persists while paging and clears when the filter changes. A selection bar appears showing the count, a Clear action, and — only when the list's delete access allows — a built-in **Bulk action** Delete that confirms first, deletes each selected row through the secured context honouring Silent failure, and reports "N of M deleted". Never "select all matching the filter".

## Acceptance criteria

- [ ] Selection semantics: page-toggle header checkbox, cross-page accumulation, cleared on filter change
- [ ] Delete hidden when delete access is statically false; partial denials reported as "N of M deleted" without revealing which/why
- [ ] Deletion goes row-by-row through the secured context; ConfirmDialog before executing
- [ ] Component tests for selection state; e2e for select-across-pages → delete → reported count and refreshed table
- [ ] Theme tokens only, named Slots shipped; changeset included

## Blocked by

- #729

PR #750, opened 2026-07-19
Admin UI: row selection and built-in bulk delete

borisno2:
Implements #733. Part of #728.

## What this adds

Row selection and a built-in **Bulk action** Delete on the admin list view, layered onto the existing `ListView`/`ListViewClient` table (and the #729 Cell registry / #730 filter engine already on `main`).

- **Selection semantics** — a checkbox column: the header checkbox toggles the **visible page**, per-row checkboxes accumulate an explicit **id set** that persists while paging and **clears when the filter changes**. Selection is persisted per-list, tagged with the active filter query, via `useSyncExternalStore` over `sessionStorage` — so it survives the soft navigation a page/sort change triggers (remount or not) without a setState-in-effect or hydration mismatch, and reads back empty under a different filter. There is no "select all matching the filter" (deliberately out of scope).
- **Selection bar** — shows the count, a **Clear** action, a named `data-slot="selection-actions"` **Slot** seam where #736 will register custom per-list bulk actions, and — only when delete access allows — the built-in **Delete**.
- **Bulk delete** — confirms first (`ConfirmDialog`), then performs a single `serverAction({ action: 'bulkDelete', ids })` round-trip. The `bulkDelete` action deletes **row-by-row through the secured context** and reports **"N of M deleted"**.

## The tricky bits

- **Silent failure → "N of M"**: `bulkDelete` (in `packages/core/src/context/index.ts`) loops the secured, access-controlled, hook-firing delegate (`db[dbKey].delete`), not raw `prisma.deleteMany`. A denied/missing row comes back as `null` (Silent failure) and is simply **not counted**; a per-row throw is caught and skipped so one row never aborts the rest. It returns `{ deleted, total: ids.length }` → `N` = successes, `M` = ids attempted; the report never reveals which rows were denied or why. The client uses a single `bulkDelete` round-trip (not a per-row `{action:'delete'}` loop, which the item-form `serverAction` wrapper would `redirect()` on and abort); the distinct `{deleted,total}` result shape is what stops that wrapper hijacking the bulk op. The status line lives in `ListViewClient` (always mounted) so it survives the selection clearing after the delete.
- **Post-delete refresh**: after a bulk delete the client calls `router.refresh()` (invalidate the Router Cache) then `router.push` to page 1 of the current filter — so an out-of-range page left after deleting a cross-page selection doesn't show stale rows. The active filter/sort/pageSize are preserved.
- **Static delete-access gating**: `ListView` evaluates the list's `access.operation.delete` up front. Delete (and the whole selection affordance) is hidden when it is **statically false** — absent (deny-by-default) or the literal `false`. A function or `true` can't be proven denied up front, so Delete is offered and per-row Silent failure handles real denials.
- **Filter-engine integration (#746)**: selection is keyed on the active filter query (`?search=`, which #746 repurposed as the URL filter query). Changing any filter token clears the accumulated id set (reads back empty), while page/sort keep it. The `?pageSize=` param coexists with the filter params without clobbering.
- **Seam for #736**: the `RowSelectionBar` ships an `actions` prop rendered in the `data-slot="selection-actions"` region; the selection-enabled gate is a single `selectionEnabled = canDelete` that #736 widens to `canDelete || hasCustomBulkActions`.

Also adds an optional `?pageSize=` URL param (preserved across sort/search/paging), which makes cross-page selection exercisable in e2e without seeding 50+ rows.

## Theming & API

- Bar and checkboxes use **Theme tokens only** (no raw colours) and ship named **Slots**: `selection-bar`, `selection-count`, `selection-actions`, `selection-delete`, `selection-clear`, `selection-status`, `selection-header`, `selection-cell`.
- New exports: `RowSelectionBar` (`RowSelectionBarProps` / `RowSelectionBarClassNames`), `useRowSelection`, and the pure `isPageFullySelected` / `getPageCheckboxState` helpers.
- Core: adds the `bulkDelete` action to the context/server-action surface (additive widening of the `serverAction` return union with `{ deleted, total }`).

## Tests

- **Component tests** (`RowSelection.test.tsx`): page-toggle header, per-row accumulation, Clear, cross-page accumulation on remount, clear-on-filter-change, "N of M deleted" absorbing a per-row denial via a single secured round-trip, Delete hidden without `serverAction`, `canDelete=false` hides everything, plus the pure selection helpers. Full UI suite green.
- **e2e** (`05-bulk-delete.spec.ts`): sign up → create 3 posts → `?search=…&pageSize=2` → select page 1, page to page 2 (selection persists) → select the rest → Delete → assert "3 of 3 deleted" and the rows gone from the refreshed table. Passing in CI.

Changesets: `minor` for `@opensaas/stack-core` (`bulkDelete` action) and `minor` for `@opensaas/stack-ui` (row selection + bulk delete UI).

---

### DIFF
--- a/.changeset/bulk-delete-server-action.md
+++ b/.changeset/bulk-delete-server-action.md
@@ -1,0 +1,23 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Add a `bulkDelete` server action for list-level bulk deletion
+
+`context.serverAction` now accepts `{ listKey, action: 'bulkDelete', ids }`. It
+deletes each id row-by-row through the secured context, honouring Silent failure
+(a denied or missing row returns `null` and is not counted; one row's error does
+not abort the rest), and returns `{ deleted, total }`.
+
+The result is deliberately a count shape rather than the single-op `{ success }`
+shape, so a UI `serverAction` wrapper that redirects on a single-item success
+(the item-form pattern) does not hijack a list-level bulk operation.
+
+```ts
+const result = await context.serverAction({
+  listKey: 'Post',
+  action: 'bulkDelete',
+  ids: ['a', 'b', 'c'],
+})
+// result: { deleted: 2, total: 3 }  // one row was denied/missing
+```
--- a/.changeset/row-selection-bulk-delete.md
+++ b/.changeset/row-selection-bulk-delete.md
@@ -1,0 +1,35 @@
+---
+'@opensaas/stack-ui': minor
+---
+
+Add row selection and a built-in Bulk action Delete to the admin list view
+
+The list table now renders a selection checkbox column when the list's delete
+access is not statically false. The header checkbox toggles the visible page,
+per-row checkboxes accumulate an explicit id set across pages, and the selection
+clears when the filter changes. A selection bar shows the count, a Clear action,
+a named `data-slot="selection-actions"` seam for future custom bulk actions, and
+— only when delete access allows — a Delete that confirms first, deletes each
+selected row through the secured context honouring Silent failure, and reports
+"N of M deleted" (partial access denials are visible without revealing which or
+why).
+
+The admin list view also honours an optional `?pageSize=` URL param, preserved
+across sorting, searching and paging.
+
+New exports: `RowSelectionBar` (with `RowSelectionBarProps` /
+`RowSelectionBarClassNames`) and the `useRowSelection` hook plus the pure
+`isPageFullySelected` / `getPageCheckboxState` helpers.
+
+```tsx
+import { RowSelectionBar, useRowSelection } from '@opensaas/stack-ui'
+
+const selection = useRowSelection('Post', filterKey)
+<RowSelectionBar
+  count={selection.selectedCount}
+  onClear={selection.clear}
+  onDelete={async () => {
+    /* delete the selected ids through the secured context */
+  }}
+/>
+```
--- a/e2e/starter-auth/05-bulk-delete.spec.ts
+++ b/e2e/starter-auth/05-bulk-delete.spec.ts
@@ -42,8 +42,9 @@ test.describe('Bulk delete', () => {
     // Narrow to these three via search, two per page → page 1 has two, page 2
     // has one. The search (the filter) is kept across page nav, so the selection
     // persists.
-    await page.goto(`/admin/post?search=${stamp}&pageSize=2`)
-    await page.waitForLoadState('networkidle')
+    // Sort by title so pagination is deterministic: page 1 = [A, B], page 2 = [C].
+    await page.goto(`/admin/post?search=${stamp}&pageSize=2&sort=title:asc`)
+    await expect(page.getByText(titles[0])).toBeVisible()
 
     const toolbar = page.getByRole('toolbar', { name: /selection/i })
 
@@ -51,9 +52,15 @@ test.describe('Bulk delete', () => {
     await page.getByRole('checkbox', { name: /select all rows on this page/i }).click()
     await expect(toolbar).toContainText('2 selected')
 
-    // Page 2: the page-1 selection persists across the navigation…
+    // Go to page 2 and wait for its (distinct) row to load before interacting —
+    // the selection persists on both pages, so a content wait is what proves the
+    // navigation actually landed on page 2.
     await page.getByRole('button', { name: /next/i }).click()
-    await page.waitForLoadState('networkidle')
+    await page.waitForURL(/[?&]page=2\b/)
+    await expect(page.getByText(titles[2])).toBeVisible()
+    await expect(page.getByText(titles[0])).toHaveCount(0)
+
+    // The page-1 selection persisted across the navigation…
     await expect(toolbar).toContainText('2 selected')
 
     // …and adding the remaining row accumulates to three.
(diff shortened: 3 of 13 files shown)


borisno2 (commented):
## Review — PR #750: Admin UI row selection + built-in bulk delete

**Verdict: APPROVE** (posted as a Comment review — GitHub blocks a formal Approve because the authenticated identity authored the PR). No blocking issues found; the items below are minor.

Reviewed strictly from the PR diff and PR-head file contents. Focus was the risk areas called out for #733 on top of the rebased #746 filter engine.

### Risk areas — all pass

**Bulk delete security & the "N of M" contract — sound.**
`bulkDelete` in `packages/core/src/context/index.ts` loops `model.delete({ where: { id } })` where `model = db[dbKey]`, and `db` is populated by `populateDbDelegate(...)` — i.e. the **secured, access-controlled, hook-firing delegate**, not raw `prisma.deleteMany`. A denied/missing row returns `null` and is not counted; a per-row throw is caught and skipped so one row never aborts the rest. Returns `{ deleted, total: props.ids.length }` → N = successes, M = ids attempted. The report never reveals which rows were denied or why. The client (`ListViewClient.handleBulkDelete`) correctly makes a single `serverAction({ action: 'bulkDelete', ids })` round-trip instead of the per-row `{action:'delete'}` loop (which the item-form `serverAction` wrapper would `redirect()` on and abort) — and the distinct `{deleted,total}` result shape is what stops that wrapper hijacking the bulk op. ✔

**Filter-engine integration — correct.**
Selection is keyed on `filterKey = initialSearch ?? ''` (the `?search=` filter query). `handleSort` and `buildPaginationUrl` both re-emit `search`, so page/sort preserve the set; `handleSearch`/`handleClearSearch` change `search` → new `filterKey` → `useRowSelection` reads back **empty** (clear-on-filter-change falls out of the read, no effect needed). `withPageSize()` appends `?pageSize=` onto params that already carry `search`/`sort`, so they coexist without clobbering. ✔

**Post-delete refresh — correct.** `router.refresh()` (invalidate Router Cache) then `router.push(buildPaginationUrl(1))`, and `buildPaginationUrl` preserves `search` + `sort` + `pageSize` — the active filter is not dropped. ✔

**Static delete-access gating — sound & fails safe.** `canDeleteList()` in `ListView.tsx`: `undefined → false` (deny-by-default hidden), literal `false → false` (hidden), boolean `true → true`, function → `true` (offered, per-row Silent failure absorbs real denials). ✔

**Theme tokens / Slots / typing.** `RowSelectionBar` uses tokens only (`bg-muted/50`, `text-foreground`, `border`, `variant="destructive|outline"`, `border-primary-foreground`) — no raw colours. All named slots present (`selection-bar/-count/-actions/-delete/-clear/-status/-header/-cell`) and documented as API surface. No `any`; the two `as` casts (`parseBulkDeleteResult`, `parseStored`) are internal narrowing at the serialization/storage boundary over `unknown`, not exposed externally. `serverAction?: (input) => Promise<unknown>` and `canDelete: boolean` props are serialisable (server-action reference + boolean). ✔

**Test coverage — adequate and not weakened.** Component tests cover page-toggle, per-row accumulation, cross-page accumulation on remount, clear-on-filter-change, Delete-hidden-without-serverAction, `canDelete=false` hides everything, and the "1 of 2 deleted" denial-absorbing path with a single secured round-trip. The e2e asserts the real 3-of-3-across-pages criterion (select page 1 → navigate → assert persisted 2 selected → select rest → `3 of 3 deleted` + rows gone). ✔

### Changeset scoping — description is stale; changesets are correct
The **PR description** says "minor for `@opensaas/stack-ui` (no `@opensaas/stack-core` changes were needed)", but the diff **does** modify `packages/core` (`context/index.ts` adds the `bulkDelete` action + widens the `serverAction` return type). The actual changeset files are correct — **both** `@opensaas/stack-core: minor` (`bulk-delete-server-action.md`) and `@opensaas/stack-ui: minor` (`row-selection-bulk-delete.md`). A core changeset **is** warranted. Action: update the PR description body to match (the changesets themselves need no change).

### Minor observations (non-blocking)
1. **`serverAction` return union widened** with `| { deleted; total }` in core. Additive and CI-green; in-repo consumers forward it or narrow via `in`/casts (`useItemForm`, `useRelationshipSearch` treat it as `unknown`), so nothing breaks. External consumers doing an exhaustive `.success` check on the typed method would need to handle the new member — worth a one-liner in the changeset if that matters.
2. **`DEFAULT_PAGE_SIZE = 50` is duplicated** in `ListViewClient.tsx` and `ListView.tsx` (verified they match today). A shared constant would remove the fragile coupling.
3. **Error feedback on failure.** The per-row `catch {}` in `bulkDelete` swallows infra errors (they just don't count) and `handleBulkDelete`/`RowSelectionBar.handleConfirm` have no `catch` — a thrown server action gives the user no feedback and leaves the selection intact. Normal denials return counts so this only bites on infra failure. Also, the persisted "N of M" banner (sessionStorage, keyed by listKey only) survives navigation and can reappear stale until a new selection starts.
4. **Nits:** row checkbox `aria-label={`Select row ${rowId}`}` uses the raw cuid (item label would read better); the `window.addEventListener('storage')` in both stores is effectively dead for `sessionStorage` (harmless).
5. **Dependency assumption:** clear-on-filter-change relies on the *entire* filter living in `?search=` (per #746). Verified today; if #746 ever adds filter params outside `?search=`, `filterKey` must be extended to include them.

Nice work — the security model (secured per-row delete, silent-failure absorbed into an opaque N-of-M count) and the persistence design (`useSyncExternalStore` over sessionStorage surviving the refresh remount) are both well-reasoned and well-tested.

---

Issue #739, opened 2026-07-19
Admin UI: Relationship-table row removal — disconnect by default, delete opt-in


borisno2:
## Parent

Spec: #728

## What to build

The ✕ control on a Relationship table row, per ADR-0018: by default it disconnects the related row from the current record (non-destructive, gated on update access of the FK-owning side). A per-relationship config opts into `delete` (ConfirmDialog, gated on the related list's delete access) or `none`. Where the schema makes disconnect statically impossible (required FK on the related side), the control is hidden unless delete was opted into. Failures follow Silent-failure reporting — the row stays with a visible message.

## Acceptance criteria

- [ ] Default ✕ disconnects; the related row still exists and appears on its own list page
- [ ] `delete` opt-in confirms then deletes via the secured context; `none` hides the control
- [ ] Required-FK relationships hide the control unless delete is opted in — statically, no runtime error path
- [ ] Access denials leave the row in place with a visible reason
- [ ] E2e covers disconnect default and delete opt-in on distinct relationships; changeset included

## Blocked by

- #734

Issue #738, opened 2026-07-19
Admin UI: pre-linked create drawer for Relationship tables


borisno2:
## Parent

Spec: #728

## What to build

"+ Add" on a Relationship table opens a drawer hosting the existing standalone create form for the related list, with the back-reference to the current record preset and hidden. The form enforces the related list's full validation and required fields (which is why this is a drawer, not an inline draft row — a deliberate deviation from the design mock recorded in the spec); on success the drawer closes and the table refreshes with the new row. The control is hidden when the session cannot create on the related list.

## Acceptance criteria

- [ ] Drawer reuses the standalone create form unmodified in behavior; back-reference preset and hidden
- [ ] Related-list validation and required fields fully enforced; errors display in the drawer
- [ ] Success closes the drawer and refreshes the table; the add control is hidden without create access
- [ ] Works for a related list whose required fields are not table columns
- [ ] E2e covers add-with-required-fields end-to-end; changeset included

## Blocked by

- #734

Issue #737, opened 2026-07-19
Admin UI: Inline cell editing in Relationship tables


borisno2:
## Parent

Spec: #728

## What to build

**Inline cell edit** in Relationship tables: an admin clicks a cell, edits in place (text/number input or inline select for options), commits on Enter or blur, cancels on Escape. Each commit is a single-field update on the **related list** through the secured context, applied optimistically: a Silent failure (access denied / row gone) or validation error reverts the cell and shows the reason. Cells whose field the session cannot write — or where the related list's update access is statically false — render read-only with no edit affordance; row-level (filter-scoped) denials surface through the commit-and-revert path. Main list tables keep click-to-navigate; this behavior is Relationship-table-only.

## Acceptance criteria

- [ ] Enter/blur commits, Escape cancels, one field per commit through the secured context
- [ ] Optimistic update reverts with a visible reason on Silent failure or validation error
- [ ] No edit affordance where update access is statically false or the field is not writable
- [ ] Select-type cells edit via inline dropdown; committed values re-render through the cell registry (badges stay badges)
- [ ] Component tests for commit/revert against a stubbed action; e2e for a real denied edit reverting
- [ ] Theme tokens only, named Slots shipped; changeset included

## Blocked by

- #734

Issue #734, opened 2026-07-19
Admin UI: item view layout with read-only Relationship tables and totals footer


borisno2:
## Parent

Spec: #728

## What to build

A record's edit page derives its layout from the list's shape: scalar and to-one fields in a details card (whole-form save via header Save/Cancel, unchanged semantics), and each to-many relationship rendered as a read-only **Relationship table** — one relationship gives the two-column split, none gives a single centered card, several stack. Table columns default to the related list's own column curation minus the back-reference to the parent, overridable on the relationship field; cells come from the cell registry so badges/formatting match the related list's page. A footer always shows the row count and additionally sums explicitly configured numeric columns, formatted by each column's Cell. Item-view config can reorder placement or demote a relationship to the existing compact picker. Editing rows arrives in a follow-up ticket — here the tables are read-only with row click navigating to the related record.

## Acceptance criteria

- [ ] Layout derives correctly for zero / one / several to-many relationships with zero config
- [ ] Columns = related list's curation minus back-reference; per-relationship override works
- [ ] Footer: count always; sums only for configured columns, rendered via that column's Cell
- [ ] Config can demote a relationship back to the compact picker; details-card save behavior unchanged
- [ ] Rows show only access-visible data; e2e covers the derived layout and footer on an example list
- [ ] Theme tokens only, named Slots shipped; changeset included

## Blocked by

- #729

PR #751, opened 2026-07-19
feat(ui,core): derive item view layout with read-only Relationship tables and totals footer

borisno2:
Implements #734. Part of #728.

## What this does

The record edit page now **derives its layout from the list's shape**:

- Scalar and to-one fields (plus any relationship demoted to the picker) stay in a **details card** — whole-form Save/Cancel with one update / one validation+hook pass, unchanged from before.
- Each to-many relationship renders as a read-only **Relationship table** section. The arrangement follows purely from the count:
  - **0** to-many relationships → single centered details card
  - **1** → two-column split (details card beside the table)
  - **several** → details card on top, tables stacked
- Relationship-table **columns** default to the related list's own column curation (`ui.listView.initialColumns`, else all non-system fields) **minus the back-reference** to the parent, overridable per relationship.
- **Cells come from the cell registry** (`CellRenderer`, #729) — badges/formatting match the related list's own page. No new renderer.
- A **totals footer** always shows the row count and additionally sums only explicitly-configured numeric columns, each rendered through that column's Cell.
- Rows and counts are fetched through the **secured context** (`context.db` include), so only access-visible data shows. A row click navigates to the related record.

## How the layout is derived

`deriveItemViewLayout(config, listKey)` (pure, in `packages/ui/src/lib/deriveItemView.ts`, unit-tested for zero/one/several + overrides/demote/order) walks the list's fields: every to-many relationship not demoted to the picker becomes a `RelationshipTableSection`, everything else is a details-card field, and `arrangement` is `single | split | stacked` from `sections.length`. `ItemForm` (edit mode) uses it to do **one** secured `findUnique` — details relationships one level deep, each Relationship table nested-included so its cells resolve — then splits the field set so the details card's whole-form save is byte-identical to the pre-#734 form. Create mode and zero-table edit fall through to the existing single-card path unchanged.

## Config (additive, `@opensaas/stack-core` minor)

```typescript
posts: relationship({
  ref: 'Post.author',
  many: true,
  ui: {
    itemView: {
      columns: ['title', 'status', 'viewCount'], // override columns
      sum: ['viewCount'],                          // sum in the footer
      // displayMode: 'picker',                    // demote to the details-card picker
    },
  },
})
// list level: ui: { itemView: { order: ['posts'] } }  // reorder sections
```

## Cells reuse & Slots for follow-ups

Relationship-table cells go through the existing **cell registry** (`CellRenderer`) — per-field override → custom-type → field-type → text fallback. The client component ships named Slots as the reviewed extension seams, no edit/add/remove affordances rendered:

- `relationship-table` — section container
- `relationship-table-toolbar` — where the pre-linked create drawer's "+ Add" (**#738**) mounts
- `relationship-table-row` — where row removal (**#739**) adds its affordance
- `relationship-table-cell` — where inline cell edit (**#737**) wraps the Cell
- `relationship-table-footer` — totals footer

## Tests

- **Unit** (`packages/ui`): `deriveItemViewLayout` — zero/one/several arrangements, columns-minus-back-reference, per-relationship column override, summed columns, demote-to-picker, list-level reorder, list-only ref.
- **E2E** (`e2e/starter-auth/05-item-view.spec.ts`): the User edit page renders the `posts` Relationship table with curated columns and a footer (`2 rows`, summed `viewCount` = 15); closed `Session` list renders an empty table (access-visible only); row click navigates to the post; a Post (no to-many) renders a single details card.

Ran green: core vitest (774), ui vitest (395 + 8 new), the new item-view e2e (3), and the existing admin-ui e2e (15, no regression). `pnpm lint` clean, `pnpm manypkg fix`, `pnpm format`. Changeset included (`@opensaas/stack-ui` + `@opensaas/stack-core` minor).

---

### DIFF
--- a/.changeset/sharp-otters-drift.md
+++ b/.changeset/sharp-otters-drift.md
@@ -1,0 +1,52 @@
+---
+'@opensaas/stack-ui': minor
+'@opensaas/stack-core': minor
+---
+
+Derive the admin item view from the list shape, with read-only Relationship tables and a totals footer (#734)
+
+A record's edit page now derives its layout from the list's shape. Scalar and
+to-one fields stay in a details card (whole-form Save/Cancel, unchanged), and
+each to-many relationship renders as a read-only **Relationship table**: one
+to-many relationship gives a two-column split, none gives a single centered
+card, several stack. Table columns default to the related list's own column
+curation minus the back-reference to the parent, cells come from the cell
+registry, and a totals footer always shows the row count plus sums for any
+explicitly-configured numeric columns (each formatted by that column's Cell).
+Rows are fetched through the secured context, so only access-visible data shows.
+Rows are read-only here — a row click navigates to the related record.
+
+`@opensaas/stack-core` gains additive item-view config (no breaking changes):
+
+```typescript
+lists: {
+  User: list({
+    fields: {
+      posts: relationship({
+        ref: 'Post.author',
+        many: true,
+        ui: {
+          itemView: {
+            // Override the Relationship table's columns…
+            columns: ['title', 'status', 'viewCount'],
+            // …and sum numeric columns in the totals footer.
+            sum: ['viewCount'],
+            // Or demote it back to the compact picker in the details card:
+            // displayMode: 'picker',
+          },
+        },
+      }),
+    },
+    // Reorder the Relationship-table sections:
+    ui: { itemView: { order: ['posts'] } },
+  }),
+}
+```
+
+New `@opensaas/stack-ui` exports: `RelationshipTable`, `RelationshipTableClient`,
+and the pure `deriveItemViewLayout` helper (with `ItemViewLayout`,
+`ItemViewArrangement`, `RelationshipTableSection`). The Relationship table ships
+named Slots (`relationship-table`, `relationship-table-toolbar`,
+`relationship-table-row`, `relationship-table-cell`, `relationship-table-footer`)
+as extension seams for the follow-up inline-edit, create-drawer, and row-removal
+work.
--- a/e2e/starter-auth/05-item-view.spec.ts
+++ b/e2e/starter-auth/05-item-view.spec.ts
@@ -1,0 +1,119 @@
+import { test, expect, type Page } from '@playwright/test'
+import { signUp, generateTestUser } from '../utils/auth.js'
+
+/**
+ * Item view layout (issue #734).
+ *
+ * The record edit page derives its layout from the list's shape: a User has one
+ * to-many relationship (`posts`) so its edit page is a two-column split — a
+ * details card beside a read-only Relationship table of the user's posts, with
+ * a totals footer (row count + summed `viewCount`). A Post has no to-many
+ * relationship, so its edit page is a single details card (no Relationship
+ * table). Rows shown come through the secured context, so only access-visible
+ * data appears.
+ */
+
+async function createPost(
+  page: Page,
+  { title, slug, viewCount }: { title: string; slug: string; viewCount: number },
+) {
+  await page.goto('/admin/post')
+  await page.waitForLoadState('networkidle')
+  await page.click('text=/create|new/i')
+  await page.waitForLoadState('networkidle')
+  await page.fill('input[name="title"]', title)
+  await page.fill('input[name="slug"]', slug)
+  await page.fill('textarea[name="content"]', `${title} content`)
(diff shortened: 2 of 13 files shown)


borisno2 (commented):
## Review — #751 derive item-view layout with read-only Relationship tables

**Verdict: APPROVE** (posted as a Comment because the reviewing identity is the PR author, so a formal Approve is blocked). Two items below are worth addressing or explicitly confirming before/after merge, but none is a correctness blocker.

This is a clean, well-scoped implementation with strong test coverage. I verified each flagged risk area against the diff:

### Risk areas — all confirmed good
- **Shape-derived layout** — `deriveItemViewLayout` is pure and config-only: 0 sections → `single`, 1 → `split`, ≥2 → `stacked`. Overrides work: `displayMode: 'picker'` demotes to `detailsFields`, `columns`/`sum` and list-level `order` are all honoured. Unit tests cover every branch (single/split/stacked, column override, sum, demote-to-picker, reorder, list-only ref).
- **Cell registry reuse (#729)** — tables render through the existing `CellRenderer` (`./cells/CellRenderer.js`), no bespoke renderer. Columns default to the related list's curation (`ui.listView.initialColumns` else non-system fields) minus the back-reference, overridable per relationship. ✔
- **Read-only scope** — client renders no edit/add/remove affordances; row click navigates via `router.push`. `#737`/`#738`/`#739` are left as empty named `data-slot` seams only. ✔
- **Access-scoped data** — one secured `context.db[getDbKey(listKey)].findUnique({ where, include })`; rows/counts come from the access-filtered include. E2E confirms the closed `Session` list renders an empty table rather than leaking. ✔
- **Totals footer** — row count always shown; sums only for explicitly-configured `sumColumns`, each rendered through the column's Cell. ✔
- **Whole-form save preservation** — verified `prepareItemForm` iterates the *passed* `listConfig.fields` / `serializeFieldConfigs(listConfig.fields)`, so restricting to `detailsListConfig` genuinely drops the to-many fields from the form payload. Save remains one update / one validation+hook pass. ✔
- **Changeset** — present, `minor` for both `@opensaas/stack-ui` and `@opensaas/stack-core` (config types added). ✔

### Findings (by severity)

**Medium**
1. **Unbounded related-row fetch — no pagination/limit.** `buildItemViewInclude` includes the to-many rows with no `take`, and the footer `count` is `rows.length` (all included rows). A record with many related rows (e.g. a long-lived User's sessions/posts) loads and renders every row on each edit-page open — memory, serialized payload, and DOM cost. List views paginate; these tables don't. Fine for the read-only ticket's happy path, but flag it as a scalability risk — consider a `take` + "showing N of M" (or an explicit follow-up ticket). `packages/ui/src/components/ItemForm.tsx` (`buildItemViewInclude`) / `RelationshipTable.tsx`.

2. **Unrelated generated-schema drift.** `examples/starter-auth/prisma/schema.prisma` drops `@@index([userId])` on `Session` and `Account` and adds `onDelete: Cascade`, with no generator/auth change in this diff to explain it. If it's just regenerating a stale committed file to current generator output, fine — but removing foreign-key indexes is a silent perf regression in the example and the cascade is a behaviour change. Please confirm it's intended current output, not accidental drift.

**Low / nits**
3. **Footer count can vanish when a section has zero columns.** In `RelationshipTableClient`, the `count` renders inside `columns.map` guarded by `index === 0`; if `columns` is empty (related list curates to nothing but the back-reference) the footer emits no cells and the "always shown" row count disappears. Edge case, but the empty-body path already special-cases `colSpan={Math.max(columns.length, 1)}` — the footer could do likewise.
4. **Default columns can surface sensitive fields.** The Account relationship table on the User edit page defaults to Account's curation, which includes `accessToken`/`refreshToken`/`idToken` (`password` is excluded, tokens aren't). Field access control still governs actual values, but this is a conscious default worth noting for the example.
5. **Type-safety convention.** Several guarded `as` casts (`readStringArray`, `sectionRows`, the `JSON.parse(...) as ...` round-trip) and `ListConfig<any>` with eslint-disables. All internal/guarded and lint-clean, consistent with the repo's "internal only where necessary" rule — acceptable, though the cast-after-guard helpers could be user-defined type predicates to fully honour the no-casting rule.
6. Trivial: unit test `'falls back to id when a related list has no curated columns'` actually asserts `columns` = `['name']`; the title is misleading.

Nice work — the derivation is well-factored (pure + unit-tested), the secured single-read is the right call, and the named Slots keep #737/#738/#739 cleanly out of scope.

---

Issue #731, opened 2026-07-19
Admin UI: Filter builder input — structural suggestions and interpreted-as chips


borisno2:
## Parent

Spec: #728

## What to build

The list view's **Filter builder** input: an admin types and gets structural autocomplete — filterable field prefixes, operators for numeric/date fields, config-enumerated values for select/checkbox, access-controlled label search for relationship tokens, and a "search anywhere" free-text fallback. The parsed query renders as removable "interpreted as" **Filter token** chips, with a live match count and a Clear action (also offered on the filtered empty state). Committing updates the URL param (from #730), so the filtered view is shareable. Suggestions are structural only — never data-derived values for unbounded fields; the client receives only serialized suggestion metadata.

## Acceptance criteria

- [ ] Suggestions cover: field prefixes, operators, enumerated values, relationship label search, free-text fallback — driven by serialized Filter spec metadata, no field-type switches
- [ ] No distinct-values queries for unbounded fields
- [ ] Chips reflect the parsed query; removing a chip updates query, URL, and results
- [ ] Match count and Clear behave per spec, including the empty-state reset
- [ ] Component tests for suggestion behavior; e2e for the type→suggest→pick→filtered-table round trip
- [ ] Theme tokens only, named Slots shipped; changeset included

## Blocked by

- #730

Issue #709, opened 2026-07-18
Design system 5/7: fields & standalone pass — classNames slots, status tokens, form rhythm


borisno2:
## Parent

#704 (spec: `specs/THEMING.md`)

## What to build

Extend the design system through the field components and standalone composites. Field components (text, integer, checkbox, select, timestamp, password, relationship, json, image/file, combobox) restyle onto the new tokens with consistent label/help/error rhythm. Composites (ListTable, ItemCreateForm, ItemEditForm, SearchBar, DeleteButton, RelationshipManager) gain structured `classNames` slot props (e.g. row/header/cell on tables) merged per part, plus `data-slot` attributes on their parts. All status rendering (badges, toasts, destructive confirmations, validation states) moves onto the `success`/`warning`/`destructive` tokens — no hardcoded status colors. Forms get the spec's section rhythm. Demoable by composing a fully custom create/edit form and list page in the composable-dashboard example that looks native to the design system.

## Acceptance criteria

- [ ] All field components consume only tokens and share consistent label/help/error styling
- [ ] Composites accept structured `classNames` props; each part carries a `data-slot` attribute; contract tests cover both
- [ ] No hardcoded status colors remain — success/warning/destructive rendering uses the tokens
- [ ] Composable-dashboard example includes a custom-composed form + list page demonstrating parity with the prebuilt admin
- [ ] Changeset for ui (`pr-changeset` skill)

## Blocked by

- #708

PR #754, opened 2026-07-20
Filter builder input UI for the admin list view (#731)

borisno2:
Implements #731. Part of #728.

## What

Adds the user-facing **Filter builder** input UI to the admin list view, layered on the already-merged filter engine (#730 / ADR-0017). The builder constructs the same `?search=` query grammar the engine already consumes — it does not invent a parallel filter path.

The builder offers:
- a **free-text search box** (bare-word search), and
- **structured condition rows** — a field picker, an operator picker (only for fields that support comparisons), and a value control that adapts per field (enum dropdown for select/checkbox, text input for text/number/date/relationship).

Applied filters are written into the `?search=` URL param and flow through the secured `context.db`, so filtering can only ever narrow what a session may see.

## How

- **`@opensaas/stack-core`** gains `serializeFilterQuery(tokens)` — the exact inverse of `parseFilterQuery`, co-located with the parser so the quoting/operator-prefix rules can never drift. Exported from the root and `./filter`.
- **`@opensaas/stack-ui`**:
  - `FilterBuilder` (`src/components/FilterBuilder.tsx`) — a `'use client'` component. Available fields/operators/value suggestions are derived entirely from each field's self-contained `getFilterSpec` via the serializable `collectFilterSuggestions` metadata — **no `switch` on field type**, and **no functions cross the server/client boundary**. Theme-token styling only; ships `data-slot` parts (`filter-builder`, `filter-builder-condition`, `filter-builder-field/operator/value`, `filter-builder-add/apply/clear`, …) and per-part `classNames` slots, consistent with #709/#729/#733.
  - Pure state helpers (`src/lib/filterBuilderState.ts`): `queryToState` / `stateToQuery` round-trip UI state ↔ query through the engine's `parseFilterQuery` + `serializeFilterQuery`.
  - `ListView` collects suggestions server-side and passes them to `ListViewClient`, which renders `FilterBuilder` in place of the old bespoke search box (keyed on the active query so it re-hydrates from the URL after each apply).
  - Exported from the root and `./standalone` subpaths.

## Tests

- **core** (`packages/core`): `serialize.test.ts` — operator prefixes, quoting rules, empty-value omission, and `parse`/`serialize` round-trip + idempotence.
- **ui** (`packages/ui`): `filterBuilderState.test.ts` (state↔query round-trips, verified against the real `buildFilterWhere`) and `FilterBuilder.test.tsx` (field/operator/value pickers per spec, enum dropdown, hydration, clear/remove, the serializable-props boundary, and the standalone navigation fallback). Existing `ListViewClient`/`RowSelection`/`AdminUISingleton` tests updated for the new search UI.
- **e2e** (`e2e/starter-auth/06-filter-builder.spec.ts`): drives the real builder — free-text search, a structured status filter, and Clear — asserting the `?search=` query and the access-scoped table.

Local results: core `832 passed`; ui `424 passed`; e2e `22 passed` (03-admin-ui, 05-bulk-delete, 05-filter, 06-filter-builder — no regressions); `pnpm lint` 0 errors; `tsc` clean.

## Changeset

`minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`.

---

### DIFF
--- a/.changeset/filter-builder-input-ui.md
+++ b/.changeset/filter-builder-input-ui.md
@@ -1,0 +1,41 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Add the Filter builder input UI for the admin list view (#731)
+
+The admin list view now ships a `FilterBuilder` that constructs the `?search=`
+filter query the filter engine already consumes (ADR-0017) — a free-text search
+box plus structured field / operator / value rows. Available fields, operators,
+and value suggestions are derived entirely from each field's self-contained
+`getFilterSpec` (via the serializable `collectFilterSuggestions` metadata), so
+there is no field-type `switch` and no functions cross the server/client
+boundary. Applied filters flow through the same secured `context.db`, so
+filtering can only ever narrow what a session may see.
+
+`@opensaas/stack-core` gains `serializeFilterQuery(tokens)` — the exact inverse
+of `parseFilterQuery` — so the builder produces the grammar the engine parses
+with the quoting and operator-prefix rules kept next to the parser.
+
+The `FilterBuilder` is composable (exported from `@opensaas/stack-ui` and
+`@opensaas/stack-ui/standalone`) with theme-token styling and `data-slot` parts
+for extension:
+
+```tsx
+import { FilterBuilder } from '@opensaas/stack-ui/standalone'
+import { collectFilterSuggestions } from '@opensaas/stack-core'
+
+// Server component: collect serializable suggestion metadata for the list.
+const suggestions = collectFilterSuggestions(listConfig, 'Post', config)
+
+// Client: build and apply a `?search=` query.
+<FilterBuilder
+  suggestions={suggestions}
+  defaultValue={search}
+  onApply={(query) => router.push(`/admin/post?search=${encodeURIComponent(query)}`)}
+/>
+```
+
+The list view wires this in automatically; existing `?search=` URLs keep
+working unchanged.
--- a/e2e/starter-auth/06-filter-builder.spec.ts
+++ b/e2e/starter-auth/06-filter-builder.spec.ts
@@ -1,0 +1,103 @@
+import { test, expect, type Page } from '@playwright/test'
+import { signUp, generateTestUser } from '../utils/auth.js'
+
+/**
+ * Filter builder input UI (issue #731, part of #728; layered on the #730 filter
+ * engine / ADR-0017).
+ *
+ * These specs drive the REAL builder in the admin list view — the free-text box
+ * and the structured field/operator/value rows — and assert it writes the
+ * `?search=` filter query the secured server-side engine consumes, narrowing the
+ * table to the matching, access-scoped rows. (05-filter.spec.ts types the query
+ * straight into the URL; this exercises the UI that produces it.)
+ */
+test.describe('Filter builder input UI', () => {
+  let testUser: ReturnType<typeof generateTestUser>
+
+  test.beforeEach(async ({ page }) => {
+    testUser = generateTestUser()
+    await signUp(page, testUser)
+  })
+
+  async function createPost(
+    page: Page,
+    { title, slug, status }: { title: string; slug: string; status: 'draft' | 'published' },
+  ) {
+    await page.goto('/admin/post')
+    await page.waitForLoadState('networkidle')
+    await page.click('text=/create|new/i')
+    await page.waitForLoadState('networkidle')
+    await page.fill('input[name="title"]', title)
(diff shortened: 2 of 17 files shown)


borisno2 (commented):
## Review — Filter builder input UI (#731)

**Verdict: APPROVE** (submitted as a Comment review — the integration identity authored the PR, so a formal Approve is not available). No blocking issues found. CI is green and every high-risk area holds up.

### Scope reviewed
17 files. Core change is limited to `serializeFilterQuery` (+ its export + test); the parser/mapper/specs are the already-merged engine (#746/ADR-0017). UI adds `FilterBuilder`, `filterBuilderState`, and rewires `ListView`/`ListViewClient`. Reviewed strictly from the PR diff and the files it touches.

### What passes (the risk areas called out)

- **Serializable client boundary — PASS.** `FilterBuilder` receives `suggestions: FilterFieldSuggestion[]` (plain data — `field`/`operators`/`freeText`/`valueSource` only). No `getFilterSpec`/`toCondition` closures cross the boundary; those stay server-side in `collectFilterSpecs`/`buildListFilterWhere`. `onApply` is client→client (`ListViewClient` → `FilterBuilder`). `FilterBuilder.test.tsx` even JSON-round-trips the props to prove it.
- **No field-type switch — PASS.** The value control is driven by the `valueSource.kind` discriminated union (`enum` → `<select>`, else `<Input>`), and operators come from `suggestion.operators`. Registry-style resolution via each field's `getFilterSpec`, no `switch (field.type)`.
- **Access scoping — PASS.** The client only writes the `?search=` string. `ListView` (server) re-parses it via `buildListFilterWhere(...)` and hands the fragment to `context.db[key].findMany/count`, where the secured context ANDs it with the access filter. No client-trusted `where` path; filters can only narrow.
- **Author-flagged behaviours — confirmed acceptable.** (a) `hello world` → two implicit-AND bare tokens (tested). (b) relationship value is a plain label-substring `<Input>` — fine for this ticket. (c) `applyQuery` sets `page=1` on every change incl. Clear, and **preserves** `sort`/`pageSize` (`withPageSize` + the `sortBy` re-set) rather than stripping them; row-selection `filterKey` keys on `search`, so #733 selection still clears on filter change. Tests assert `?sort=title%3Adesc&page=1` and Clear→`?page=1`.
- **Regression surface — PASS.** Updated tests still assert real behaviour, not loosened: `ListViewClient.test` checks the new nav URLs (`?search=test&page=1`, Clear→`?page=1`, sort preserved); `RowSelection`/`AdminUISingleton` only add the now-required `usePathname` mock. #733 selection + bulk delete path is untouched.
- **Type safety / theming — PASS.** No `any` in the new UI code; theme tokens only (`border-input`, `bg-background`, `text-foreground`, `ring-ring`, `muted`…); comprehensive `data-slot` parts + per-part `classNames` slots.
- **Changeset — PASS.** `@opensaas/stack-core: minor` (new public `serializeFilterQuery`) + `@opensaas/stack-ui: minor` (new `FilterBuilder`), matching exactly the packages changed.

### Non-blocking findings (low severity)

1. **Round-trip gap: embedded double-quote** — `packages/core/src/filter/serialize.ts` `quoteIfNeeded`. A value containing a `"` is wrapped in quotes, but the grammar has no escape mechanism, so it does not round-trip: `serializeFilterQuery(parseFilterQuery('a"b'))` → `"a"b"`, which `parseFilterQuery` then re-tokenises as two tokens (`a`, `b"`). The doc comment calls this the "exact inverse"; that claim is slightly overstated for this case. In practice benign (rare input; `buildFilterWhere` degrades gracefully), but worth either a doc caveat or quote-doubling. Not covered by `serialize.test.ts`.

2. **Round-trip gap: bare free-text token that looks like `ident:value`** — same file, the `field === null` branch emits the value unquoted. A free-text token like `foo:bar` re-parses as a *field* token rather than free text. Cushioned by the engine degrading unknown fields to free text, but if `foo` is a real filterable field the token silently promotes to a structured filter. Could be closed by quoting bare words matching the parser's `FIELD_PREFIX_RE`.

3. **Standalone fallback drops sibling params** — `FilterBuilder.applyQuery` default path does `router.push(\`${pathname}?search=...\`)`, discarding any existing `sort`/`page`/`pageSize`. Fine in the admin list view (which supplies `onApply` that preserves them), but a composable/standalone consumer may be surprised. Consider merging existing params in the fallback, or documenting it.

4. **Nit:** `e.target.value as FilterOperator` in the operator `<select>` handler is a cast; the option set is constrained to the field's operators so it's safe, but a lookup/guard would avoid the cast.

None of these block merge. Nice work — the server/client seam and access-narrowing invariant are handled correctly, and the tests are substantive.

---

Issue #735, opened 2026-07-19
Admin UI: chrome polish — opt-in nav counts and avatar label cells


borisno2:
## Parent

Spec: #728

## What to build

Two per-list opt-ins. (1) Nav counts: a list config flag adds the access-scoped row count next to that list's Admin chrome nav item; default off, no count queries for lists that don't opt in. (2) Avatar label cells: a list config flag renders the label column with an initials bubble (initials and color derived deterministically from the row) ahead of the emphasized Item label; default remains text-only, and a per-field cell override still wins.

## Acceptance criteria

- [ ] Nav counts appear only for opted-in lists, computed through the secured context, reflecting what the session may see
- [ ] No count queries execute for lists that haven't opted in
- [ ] Avatar cells are deterministic per row, opt-in per list, and overridable via the field's cell override
- [ ] Avatar palette is Theme-token-derived — no raw hex
- [ ] E2e covers an opted-in count and an avatar list vs a default list; changeset included

## Blocked by

- #729

PR #755, opened 2026-07-20
feat(core,ui): admin chrome polish — opt-in nav counts + avatar label cells (#735)

borisno2:
Implements #735. Part of #728.

## What

Two per-list admin-chrome opt-ins, both **off by default** so upgrading changes nothing until a list opts in.

### 1. Nav counts (`ui.navCount: true`)
An access-scoped record count next to the list's chrome nav item.
- Fetched through the **secured `context.db`** — never a raw prisma total — so it only ever reflects what the session may see.
- **No count query runs** for lists that don't opt in.
- A list whose **query access is statically denied** (absent / literal `false`) renders **no count**, not a misleading `0`. An access-scoped `0` (session may query but sees none) still renders as `0`.
- Singletons are skipped (no meaningful count).

### 2. Avatar label cells (`ui.avatar: true`)
The list's label column renders an initials bubble ahead of the emphasized Item label.
- Initials **and** colour derive **deterministically from the row** (stable across pages/reloads — no `Math.random`).
- Palette is **Theme-token-derived** (`bg-primary/15 text-primary`, …) — **no raw hex**.
- Degrades gracefully: empty label → neutral `?` bubble with `aria-label="Unknown"`.
- Accessible: `role="img"` + `aria-label` (full name announced, initials hidden from a11y tree).
- A **per-field cell override (`ui.cell`) still wins** — the client routes to the override first.

## How

- **core**: `navCount` + `avatar` added to `ListUIConfig`; new `resolveNavCounts(context, config)` and `isListQueryStaticallyDenied(listConfig)` helpers (exported from the root). `resolveNavCounts` only touches opted-in, non-singleton, not-statically-denied lists and reads each count concurrently via `context.db[key].count()`.
- **ui**: `Avatar` primitive + `getInitials` / `getAvatarTone` / `AVATAR_TONES` helpers; `AvatarLabelCell`; `AdminUI` (now async) resolves counts and passes `navCounts` to a still-synchronous, presentational `Navigation`, which renders the count via the existing `Badge`. `ListView` resolves the avatar column through the shared label seam (`getLabelFieldName`) and `ListViewClient` routes the label column to `AvatarLabelCell` unless a per-field cell override wins.
- New **Slots** (part of the reviewed API surface): `avatar`, `cell-avatar-label`, `nav-count`.
- **example**: starter-auth opts `Post` into both features; `User` left default so the e2e contrasts opted-in chrome against the default.

## Tests (all green locally)

- **core** (`nav-count.test.ts`): static-deny truth table; `resolveNavCounts` counts only opted-in lists, omits statically-denied ones (no query, no misleading zero), passes function-gated counts through the secured context, reports access-scoped `0`, skips singletons.
- **ui**: avatar initials/tone determinism + empty-label fallback + token-only palette; `Avatar` primitive a11y/slot/degradation; `AvatarLabelCell` render + empty; `ListViewClient` avatar-column routing + **per-field override precedence** + default text-only; `Navigation` nav-count badge presence/absence/`0`.
- **e2e** (`06-chrome-polish.spec.ts`): nav count on the opted-in list only, cross-checked equal to the access-scoped list-page total; avatar cell on the opted-in list vs text-only on the default list. Ran green against real Chromium.

Full suites: core 816 passed, ui 437 passed, both e2e specs passed; `pnpm lint` 0 errors, `pnpm format` clean, `tsc` clean (via `pnpm build`).

## Changeset
`minor` for `@opensaas/stack-core` and `@opensaas/stack-ui` (additive config + exports).

## Reviewer notes
- `AdminUI` became `async` to resolve counts before rendering the shell; existing `AdminUI` tests already `await AdminUI(...)` and only inspect `<main>`, so they're unaffected. Nav counts add a small serial query per opted-in list before first paint (zero queries when nothing opts in).
- Coordinated with the concurrent #731 filter-UI lane: changes to shared files (`AdminUI`, `ListView`/`ListViewClient`, `index.ts` exports, `Navigation`) are additive and should rebase trivially.

---

### DIFF
--- a/packages/core/src/config/nav-count.ts
+++ b/packages/core/src/config/nav-count.ts
@@ -1,0 +1,85 @@
+import type { AccessContext } from '../access/types.js'
+import { getDbKey } from '../lib/case-utils.js'
+import type { ListConfig, OpenSaasConfig, TypeInfo } from './types.js'
+
+/**
+ * The minimal shape a `context.db[list]` delegate needs for a nav count. The
+ * access-controlled DB carries an `any` index signature, so this narrows the
+ * lookup to just the access-scoped `count` we call.
+ */
+type CountDelegate = {
+  count?: (args?: { where?: unknown }) => Promise<number>
+}
+
+/**
+ * Whether a list's query access is **statically denied** — provably no session
+ * can read any row without evaluating a session-dependent function.
+ *
+ * Mirrors the list view's `canDeleteList` static check (issue #733): query
+ * access is statically denied when it is absent (deny-by-default, matching the
+ * access engine) or the literal boolean `false`. A function — or `true` —
+ * cannot be decided up front, so the count is fetched through the secured
+ * context, which re-applies the real rule per row.
+ *
+ * A statically-denied list must render no nav count at all: a `0` there would
+ * imply "this list is empty" when the truth is "you may see none of it".
+ */
+export function isListQueryStaticallyDenied(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
+  listConfig: ListConfig<any>,
+): boolean {
+  const queryAccess = listConfig.access?.operation?.query
+  if (queryAccess === undefined) return true
+  if (typeof queryAccess === 'boolean') return !queryAccess
+  return false
+}
+
+/**
+ * Resolve the access-scoped nav counts for the lists that opt in via
+ * `ui.navCount` (issue #735).
+ *
+ * The returned map is keyed by list key and contains an entry **only** for a
+ * list that:
+ *
+ * 1. opts in (`ui.navCount === true`) — no count query runs for any other list,
+ * 2. is not a singleton (a single-record list has no meaningful count), and
+ * 3. does not have statically-denied query access (see
+ *    {@link isListQueryStaticallyDenied}).
+ *
+ * Each count is read through the secured `context.db`, whose `count` applies the
+ * access filter, so the number reflects exactly what the current session may
+ * see (a denied row is not counted). Counts are fetched concurrently.
+ */
+export async function resolveNavCounts(
+  context: AccessContext,
+  config: OpenSaasConfig,
+): Promise<Record<string, number>> {
+  const counts: Record<string, number> = {}
+  const lists = config.lists ?? {}
+
+  const optedIn = Object.keys(lists).filter((listKey) => {
+    const listConfig = lists[listKey] as ListConfig<TypeInfo> | undefined
+    if (!listConfig) return false
+    if (listConfig.ui?.navCount !== true) return false
+    if (listConfig.isSingleton) return false
+    return !isListQueryStaticallyDenied(listConfig)
+  })
+
+  await Promise.all(
+    optedIn.map(async (listKey) => {
+      const delegate: CountDelegate | undefined = context.db?.[getDbKey(listKey)]
+      if (!delegate?.count) return
+      // A single list's count failing (a DB hiccup, a throwing access filter or
+      // hook) must not blank the whole admin chrome — `AdminUI` awaits this
+      // before rendering the shell. Degrade like the sibling `ListView`: log and
+      // omit just that badge. Other lists' counts are unaffected.
+      try {
+        counts[listKey] = await delegate.count()
+      } catch (error) {
+        console.error(`Failed to resolve nav count for ${listKey}:`, error)
+      }
+    }),
+  )
+
+  return counts
+}
--- a/packages/ui/src/components/ListViewClient.tsx
+++ b/packages/ui/src/components/ListViewClient.tsx
@@ -18,6 +18,7 @@ import { Checkbox } from '../primitives/checkbox.js'
 import { EmptyState } from './EmptyState.js'
 import { CellRenderer } from './cells/CellRenderer.js'
 import { FilterBuilder } from './FilterBuilder.js'
+import { AvatarLabelCell } from './cells/AvatarLabelCell.js'
 import { RowSelectionBar } from './RowSelectionBar.js'
 import { useRowSelection, getPageCheckboxState } from '../lib/useRowSelection.js'
 import { useBulkStatus } from '../lib/useBulkStatus.js'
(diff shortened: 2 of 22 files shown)


packages/core/src/config/nav-count.ts
borisno2, line 1: Minor (non-blocking) — resilience: `resolveNavCounts` runs each `count()` inside `Promise.all` with no error handling, and `AdminUI` `await`s it directly before rendering the shell. If any opted-in list's `count()` rejects (DB hiccup, an access filter/hook that throws), `Promise.all` rejects and the **entire admin chrome fails to render** — for every page, not just the affected list.

Note that the sibling `ListView` deliberately wraps its own `findMany`/`count` in try/catch and logs, precisely so a data-fetch failure degrades gracefully rather than taking down the view. A decorative nav count should be at least as forgiving. Consider catching per-list so a failing count simply omits that badge:

```ts
optedIn.map(async (listKey) => {
  const delegate = context.db?.[getDbKey(listKey)]
  if (!delegate?.count) return
  try {
    counts[listKey] = await delegate.count()
  } catch (error) {
    console.error(`Failed to resolve nav count for ${listKey}:`, error)
  }
})
```

Not a blocker — the access-scoping itself is correct — but worth hardening since this now sits on the render path of every admin page.

---

packages/ui/src/components/ListViewClient.tsx
borisno2, line 1: Nit — `renderCell` gates the avatar treatment on `!field.ui?.cell`, but `CellRenderer`'s own precedence chain is `ui.cell` → `ui.fieldType` → `ui.type`. So if the label field carries a `ui.fieldType` (a registered custom cell) but no `ui.cell`, the avatar wins over that custom cell — an inconsistency with the documented "per-field override still wins" seam. Probably fine (a list-level `ui.avatar` opt-in reasonably outranks a type-registry hint), and the PR only ever claims `ui.cell` precedence, but worth a one-line note or deciding intentionally. No change required.

---

borisno2 (commented):
## Review — PR #755: admin chrome polish (nav counts + avatar label cells)

**Verdict: APPROVE** ✅ _(posted as a Comment review — GitHub blocks a formal Approve because the PR is attributed to the integration identity. This is an explicit approval.)_

Reviewed strictly from the PR diff plus the surrounding code it touches (`CellRenderer`, `ListView`/`ListViewClient`, `label.ts`, `canDeleteList`, the example admin pages). CI is fully green (test, e2e, coverage, security, Vercel). Two non-blocking notes inline; nothing here should hold the merge.

### Access-scoped nav counts (top concern) — correct
- Counts are fetched through the **secured** `context.db[getDbKey(key)].count()`, never raw prisma — the number can only reflect what the session may see. ✔
- `isListQueryStaticallyDenied` is a faithful inverse of the existing `canDeleteList` static check: `undefined` → denied, literal `false` → denied, `true`/function → decided per request. A statically-denied list is filtered out of `optedIn` entirely, so **no query runs and no misleading `0` renders**. ✔
- An access-scoped `0` (query allowed, session sees none) is preserved as `0` and rendered as a distinct badge. ✔
- Opted-out lists run **zero** count queries; singletons are skipped. ✔
- Truth table well covered in `nav-count.test.ts` (opt-in filter, static-deny omission with no query, function-gated pass-through, access-scoped `0`, singleton skip); `NavigationNavCounts.test.tsx` covers badge presence/absence/`0`.

### Avatar accessible-name regression & fix — verified
- `Avatar` with `decorative` renders `aria-hidden="true"` and **no** `role="img"`/`aria-label`; the inner initials span is `aria-hidden` too — the bubble contributes nothing to the cell's accessible name. ✔
- `AvatarLabelCell` uses `<Avatar name={label} decorative />`, so the label cell's accessible name stays exactly the label. `06-chrome-polish.spec.ts` re-asserts `getByRole('cell', { name: 'Avatar Post', exact: true })` — the exact lookup `05-filter` regressed on — and the full e2e suite passes, so the regression is gone. ✔
- Standalone `Avatar` (default `decorative=false`) stays a labelled image (`role="img"` + `aria-label`, `Unknown` fallback). ✔
- Determinism: djb2-style hash in `getAvatarTone`, no `Math.random`; initials derived purely from the label. Palette is theme-token class pairs only (`bg-primary/15 text-primary`, …) — a test asserts no raw hex. ✔

### Per-field cell override precedence — correct
`renderCell` routes to `AvatarLabelCell` only when `column === avatarColumn && !field.ui?.cell`; otherwise it falls through to `CellRenderer` (`ui.cell` → `ui.fieldType` → `type`). A `ui.cell` override wins, confirmed by `ListViewClientAvatar.test.tsx`. (Minor `ui.fieldType`-vs-avatar edge noted inline.)

### AdminUI async change — safe
`AdminUI` became `async` to resolve counts. All example consumers render `<AdminUI/>` inside async Server Component pages (`app/admin/[[...admin]]/page.tsx`), which supports async components natively; existing AdminUI tests already `await AdminUI(...)`. When nothing opts in, `resolveNavCounts` issues zero queries.

### Boundary / conventions
- `avatarColumn` (string) and `navCounts` (`Record<string, number>`) are both serialisable across the server/client boundary. ✔
- No exposed `any` / external casts; the `ListConfig<any>` disables match the existing `label.ts` convention, and `unknown` stays internal to `CountDelegate`. ✔
- Named data-slots present: `avatar`, `cell-avatar-label`, `nav-count`. ✔
- Changeset scoped `@opensaas/stack-core: minor` + `@opensaas/stack-ui: minor`, matching the additive surface. ✔

### Non-blocking
1. **Resilience (inline on `nav-count.ts`)** — `resolveNavCounts` has no error handling; a single rejecting `count()` would crash the whole admin shell via `AdminUI`'s `await`. `ListView` guards its own count for exactly this reason; a per-list try/catch would let a failed count degrade to "no badge". Worth hardening since it now sits on every admin page's render path.
2. **Nit (inline on `ListViewClient.tsx`)** — avatar wins over a label field's `ui.fieldType` custom cell (only `ui.cell` is gated). Likely intentional; worth a note.

Clean, well-tested, well-documented work — the security-sensitive path (access-scoped counts) is exactly right.

---

Issue #753, opened 2026-07-20
Investigate: generator output drops @@index([userId]) and adds onDelete: Cascade on auth Session/Account


borisno2:
## Context

Surfaced during review of #751 (item view, #734) and verified by the implementer — filed as non-blocking so #751 could merge.

While regenerating `examples/starter-auth/prisma/schema.prisma`, the current generator output **differs from the committed schema on `main`** (the committed file was stale). The drift was confirmed **reproducible from `main`'s own unmodified `opensaas.config.ts`** (i.e. independent of #734's changes):

```diff
model Session {                                     model Account {
-  user  User? @relation(fields: [userId], references: [id])
+  user  User? @relation(onDelete: Cascade, fields: [userId], references: [id])
-  @@index([userId])
```

So current `pnpm generate` output for the auth lists (a) **adds `onDelete: Cascade`** on the `user` relations of `Session`/`Account`, and (b) **drops `@@index([userId])`** on those models. #751 merged the fresh regeneration, so this output is now on `main`.

## Why it's worth a look

- **`onDelete: Cascade`** is plausibly intended (deleting a user should clean up their sessions/accounts) — but it should be a deliberate, documented choice in the auth plugin's relationship definitions, not an unexplained generator behaviour.
- **Dropping `@@index([userId])`** is a potential **silent performance regression**: `Session`/`Account` are looked up by `userId` on every auth check, and without the index those become table scans as the tables grow. If the index is wanted, the generator (or the auth relationship definitions) should emit it.

## What to determine

- [ ] Is the `onDelete: Cascade` on auth relations intended and documented? If yes, note it in the auth plugin / ADR; if no, fix the relationship definition.
- [ ] Should `@@index([userId])` (and any other FK indexes) be emitted for these relations? If yes, why did the generator stop emitting them — is this a regression in relationship-field index generation, or a change in how the auth plugin declares the relations?
- [ ] If FK indexes are desired generally, ensure the generator emits `@@index` for relationship foreign keys (or document why not) and regenerate the affected examples.
- [ ] Add a test locking down the expected auth-schema generator output so future drift is caught.

## Notes

Not a blocker for #728 wave work. The schema on `main` currently reflects the generator's actual output; this issue is about whether that output is *correct*, not about #751.

PR #756, opened 2026-07-20
feat(core,ui): relationship-table row removal — disconnect by default, delete opt-in (#739)

borisno2:
Implements #739. Part of #728.

## What

Adds a `✕` row-removal control to the read-only Relationship tables delivered by #734 on the admin item view, per **ADR-0018**. It mounts on the `relationship-table-row` seam #734 left for exactly this ticket.

Removal semantics (config-driven via `ui.itemView.removeAction`):

- **`'disconnect'` (default)** — non-destructively unlinks the related row from the current record. The related row still exists and appears on its own list page. Gated on the **related list's update access**.
- **`'delete'` (opt-in)** — truly deletes the related row, behind a `ConfirmDialog`, gated on the **related list's delete access**.
- **`'none'`** — hides the control.
- Where the schema makes disconnect statically impossible (a required FK on the related side — the back-reference declares `db.isNullable: false`), the control is hidden unless `delete` is opted in. No runtime error path.

## How

- **`@opensaas/stack-core`**
  - New `removeRelated` server action (`mode: 'disconnect' | 'delete'`). It operates on the **related row** through the secured context, so the related list's own operation/field access and hooks apply — never the parent's. Disconnect is an `update` nulling the back-reference (`{ disconnect: true }` for a to-one back-ref; `{ disconnect: { id: parentId } }` for a to-many/M2M back-ref); delete is a secured `delete`.
  - It returns a **distinct `{ removed }` shape** (mirroring `bulkDelete`'s `{ deleted, total }`), so a `serverAction` wrapper that `redirect()`s on `{ success: true }` — the item-form pattern used in the example — does **not** hijack an in-place removal. This was the concrete bug caught in testing: without it, removing a row redirected to the related list.
  - New `RelationshipItemViewConfig.removeAction` config option.
- **`@opensaas/stack-ui`**
  - `deriveItemViewLayout` now resolves `removeAction` and a static `disconnectable` flag per section.
  - `RelationshipTable` (server) evaluates the related list's operation access via the existing `isOperationPotentiallyAllowed` helper and passes a resolved `removeMode` (hard-deny hides the control; a filter result shows it and surfaces row-level denials at commit time).
  - `RelationshipTableClient` renders the accessible `✕` (aria-label, keyboard, `data-slot="relationship-table-remove"`), confirms before delete, optimistically hides + `router.refresh()` on success, and reverts with a visible `role="alert"` reason on a Silent failure.
  - Read-only behaviour for everything else is unchanged — no inline cell edit (#737) or create drawer (#738).
- **`examples/starter-auth`** — adds a `Note` owned-child list; the User item view now shows `posts` (disconnect default) and `notes` (delete opt-in) as two distinct relationship tables.

## Removal semantics rationale

Per ADR-0018 and `CONTEXT.md` ("every edit in a Relationship table is an operation on the related list — subject to that list's access control, never the parent's"), both disconnect and delete are operations on the **related** list, gated on the **FK-owning (related) side**. Disconnect is the default because a generic admin UI can't know whether a related row is an owned child (delete is right) or a shared reference like a tag (delete would destroy data other records use).

## Tests (all green locally)

- **core** (`context.test.ts`): `removeRelated` disconnect (to-one), delete, to-many/M2M disconnect payload, and access-denied Silent failure. 836 core tests pass.
- **ui unit**: `deriveItemView` (default/delete/none removeAction, required-FK + list-only ⇒ not disconnectable) and `RelationshipTableClient` (disconnect happy path + refresh, delete confirm, denied revert-with-reason, no control when `removeMode` null). 446 unit + 69 browser tests pass.
- **e2e** (`07-relationship-row-removal.spec.ts`): disconnect default (post disconnected, still on its list) and delete opt-in (note confirmed + deleted, gone from its list). Existing `05-item-view` / `03-admin-ui` specs re-run green against the modified example.
- `pnpm lint` (0 errors), `pnpm manypkg check`, `prettier --check`, and `tsc` (via package builds + `next build`) all clean.

## Changeset

`minor` for `@opensaas/stack-core` (new `removeRelated` action + `removeAction` config — additive) and `@opensaas/stack-ui` (removal affordance).

## Reviewer focus

- **Access-control path**: disconnect/delete are scoped through the related list's access; a hard deny hides the affordance, a filter-scoped denial surfaces as commit-and-revert with a generic reason (no leak of denied-vs-absent).
- The `{ removed }` distinct-shape decision (vs. reusing `delete`/`success`) — it's what keeps redirect-on-success wrappers from hijacking the removal.
- `disconnectable` derivation uses `db.isNullable === false` as the required-FK signal (relationship FKs are generated nullable by default, so this reflects declared intent); list-only refs are treated as not disconnectable.

---

### DIFF
--- a/.changeset/relationship-row-removal.md
+++ b/.changeset/relationship-row-removal.md
@@ -1,0 +1,38 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Add relationship-table row removal to the admin item view (ADR-0018)
+
+Each read-only Relationship table row now has a ✕ removal control. By default it
+**disconnects** the related row from the current record (non-destructive — the
+row survives and still appears on its own list), gated on the related list's
+update access. A per-relationship opt-in truly deletes the related row (behind a
+confirmation, gated on the related list's delete access), or hides the control
+entirely. Where the schema makes disconnect impossible (a required foreign key on
+the related side) the control is hidden unless delete is opted in. Removals run
+through the secured context, so an access-denied removal is a Silent failure: the
+row stays with a visible reason.
+
+Configure per relationship via `ui.itemView.removeAction`:
+
+```typescript
+User: list({
+  fields: {
+    // Default: ✕ disconnects the post (it still exists).
+    posts: relationship({ ref: 'Post.author', many: true }),
+    // Opt in to destructive delete (confirmed).
+    notes: relationship({
+      ref: 'Note.owner',
+      many: true,
+      ui: { itemView: { removeAction: 'delete' } }, // 'disconnect' (default) | 'delete' | 'none'
+    }),
+  },
+})
+```
+
+`@opensaas/stack-core` adds a `removeRelated` server action (distinct
+`{ removed }` result shape, like `bulkDelete`, so a redirect-on-success wrapper
+never hijacks an in-place removal) and the `RelationshipItemViewConfig.removeAction`
+option.
--- a/e2e/starter-auth/07-relationship-row-removal.spec.ts
+++ b/e2e/starter-auth/07-relationship-row-removal.spec.ts
@@ -1,0 +1,121 @@
+import { test, expect, type Page } from '@playwright/test'
+import { signUp, generateTestUser } from '../utils/auth.js'
+
+/**
+ * Relationship-table row removal (issue #739, ADR-0018).
+ *
+ * The User item view renders two to-many Relationship tables with different
+ * removal semantics:
+ * - `posts` (default `disconnect`): the ✕ unlinks the post from the user. The
+ *   post itself survives and still appears on the Post list.
+ * - `notes` (opt-in `delete`): the ✕ confirms, then truly deletes the note.
+ *
+ * Both removals run through the secured context, so they respect the related
+ * list's own access control.
+ */
+
+async function createPost(page: Page, { title, slug }: { title: string; slug: string }) {
+  await page.goto('/admin/post')
+  await page.waitForLoadState('networkidle')
+  await page.click('text=/create|new/i')
+  await page.waitForLoadState('networkidle')
+  await page.fill('input[name="title"]', title)
+  await page.fill('input[name="slug"]', slug)
+  await page.fill('textarea[name="content"]', `${title} content`)
+  await page.click('button[type="submit"]')
+  await page.waitForURL(/admin\/post$/, { timeout: 10000 })
+}
+
+async function createNote(page: Page, body: string) {
+  await page.goto('/admin/note')
+  await page.waitForLoadState('networkidle')
+  await page.click('text=/create|new/i')
+  await page.waitForLoadState('networkidle')
+  await page.fill('textarea[name="body"]', body)
+  await page.click('button[type="submit"]')
+  await page.waitForURL(/admin\/note$/, { timeout: 10000 })
+}
+
+async function openCurrentUserEditPage(page: Page, email: string) {
+  await page.goto('/admin/user')
+  await page.waitForLoadState('networkidle')
+  await page.locator('tr', { hasText: email }).locator('a:has-text("Edit")').click()
+  await page.waitForURL(/\/admin\/user\/[^/]+$/, { timeout: 10000 })
+}
+
+function tableByHeading(page: Page, name: string) {
+  return page.locator('[data-slot="relationship-table"]', {
+    has: page.getByRole('heading', { name, exact: true }),
+  })
+}
+
+test.describe('Relationship-table row removal', () => {
+  test('default ✕ disconnects a post (the post still exists)', async ({ page }) => {
+    const user = generateTestUser()
+    await signUp(page, user)
+
+    const stamp = `${Date.now()}`
+    await createPost(page, { title: `Keep ${stamp}`, slug: `keep-${stamp}` })
+    await createPost(page, { title: `Drop ${stamp}`, slug: `drop-${stamp}` })
+
+    await openCurrentUserEditPage(page, user.email)
+
+    const posts = tableByHeading(page, 'Posts')
+    await expect(posts).toBeVisible()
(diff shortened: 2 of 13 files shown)


borisno2 (commented):
## Review — PR #756 (relationship-table row removal, #739)

**Verdict: APPROVE** (posted as a Comment because the integration identity authored the PR, so a formal GitHub Approve is blocked). No blocking issues. The access-control path — the top risk — is correct and leak-free.

### What it does
Adds a `✕` row-removal control to the read-only Relationship tables from #734, config-driven via `ui.itemView.removeAction` (`'disconnect'` default | `'delete'` | `'none'`), implemented against the existing ADR-0018. Core gains a secured `removeRelated` server action; UI gains gating + the accessible control.

### Access-control correctness & no-leak (top concern) — PASS
- **Runs through the secured context, no raw prisma.** `removeRelated` operates on `model = db[dbKey]` where `db` is the `populateDbDelegate` wrapper (`packages/core/src/context/index.ts:453`), so the **related** list's own access + hooks apply, never the parent's — matching CONTEXT.md.
- **Correct gating.** disconnect → `model.update` (related list **update** access); delete → `model.delete` (related list **delete** access). `resolveRemoveMode` (`RelationshipTable.tsx`) evaluates the *related* list's `access.operation` via `isOperationPotentiallyAllowed`.
- **Hard-deny hides, filter-scoped shows-then-reverts.** `isOperationPotentiallyAllowed` returns `false` only for a static `false`/missing rule (control hidden); a filter object is "potentially allowed" (control shown), and a row that fails the filter at commit returns `null` → `{ removed: false, error: 'Access denied or operation failed' }` → row reverts with a `role="alert"` reason. Exactly the (a)/(b) behaviour required.
- **No denied-vs-absent leak.** Silent-failure `null` maps to a single generic message, identical whether the row was denied or absent (`context/index.ts:506`). Post-check DB/validation errors surface their own message, but only after access already passed — no leak beyond what the caller is authorized for.

### Removal semantics per ADR-0018 — PASS
- disconnect is a non-destructive `update` nulling the back-ref: `{ disconnect: true }` for a to-one back-ref, `{ disconnect: { id: parentId } }` for a to-many/M2M back-ref (`backRefIsMany` from the related field's `many === true`). `parentId` is a required client prop, always supplied — no `{ id: undefined }` path.
- `disconnectable` uses `db.isNullable === false` as the required-FK signal (`deriveItemView.ts` `isDisconnectable`), and treats list-only refs (no back-reference) as not disconnectable. Sound: FKs are generated nullable by default, so `=== false` is the declared-intent signal.
- **Required-FK can't be left invalid by a disconnect.** UI hides the control; and even if `removeRelated` disconnect is invoked directly on a required FK, the DB constraint rejects the `update` atomically (caught → error result), so no invalid state is persisted.

### `{ removed }` distinct shape — PASS
Distinct from single-op `success` (mirrors `bulkDelete`'s `{ deleted, total }`), so the example's redirect-on-`success` `serverAction` wrapper doesn't hijack the in-place removal. Client `readRemoveOutcome` handles it and treats an unrecognised shape as failure (row never silently lost). Optimistic hide + `router.refresh()` on success; `refresh()` preserves URL-encoded filter/nav state.

### Scope boundary — PASS
Only the `relationship-table-remove` affordance is added. No inline cell edit (#737) or create drawer (#738); cells remain read-only (`CellRenderer` unchanged). The other named seams are left intact for those tickets.

### Example / schema regen — PASS
Adds `Note` (owned-child, `delete` opt-in) beside `posts` (`disconnect` default), modelling both modes for e2e. **Verified the schema is legitimate generator output:** re-ran `pnpm generate` in `examples/starter-auth` → zero diff; the `Note` model mirrors `Post`'s generator pattern exactly (`ownerId String? @map("owner")`, `@@index([ownerId])`, and no `createdAt`/`updatedAt` — Post lacks them too in this example). Not hand-edited. `Note` has proper access (`isOwner` scoping) + owner-assign hook. The pre-existing auth-schema drift (#753) is untouched, as expected.

### Type safety / a11y / theming — PASS
No exposed `any`/casts (`AccessContext<unknown>`; the two `any` uses are `ListConfig<any>`, matching the file's existing eslint-disable). Client props are serialisable; `context` is only passed to the server component, never the client — only `serverAction` (a server-action reference) crosses. `✕` control has `aria-label`, keyboard focus (Button), `stopPropagation` so removal doesn't also navigate; delete uses `ConfirmDialog variant="danger"` (supported); reason uses `role="alert"`. Theme tokens only (`text-destructive`, `text-muted-foreground`).

### Changeset & tests — PASS
Changeset is `minor` for `@opensaas/stack-core` + `@opensaas/stack-ui` — matches the packages actually changed (additive action/config + affordance). Core tests cover disconnect (to-one), delete, M2M disconnect payload, and access-denied silent failure. UI unit tests cover default/delete/none parsing, required-FK + list-only ⇒ not disconnectable, and client happy-path+refresh / revert-with-reason (`role="alert"`) / confirm-before-delete / no-control. e2e (`07-relationship-row-removal.spec.ts`) exercises both disconnect (post survives) and delete (note gone).

### Minor / non-blocking
1. **PR description says it "adds ADR-0018", but ADR-0018 already exists in `main`** (added by #740, commit `7ddd97a`); this PR modifies no `docs/adr/` file. Code implements the existing ADR correctly — just a description/framing inaccuracy that may send reviewers looking for the ADR in the diff.
2. **`parentListKey` is a dead prop** on `RelationshipTableClient` — declared in the props interface ("used for accessible labelling") and passed from `RelationshipTable`, but never destructured/used in the body (the `aria-label` uses `title`). Remove it or use it.
3. **Footer count is briefly stale** between optimistic hide and the `router.refresh()` settling (e.g. "2 rows" with 1 visible). Cosmetic, self-heals. Optional.
4. **Test gap (optional):** delete-mode access-denied silent failure isn't unit-tested at the core level (only disconnect-denied is), though the path is identical to `bulkDelete`'s covered delete path.

Nice work — the access-control gating, silent-failure semantics, and the deliberate distinct-shape decision are all handled correctly.


---

PR #757, opened 2026-07-20
feat(core,ui): pre-linked create drawer for relationship tables (#738)

borisno2:
Implements #738. Part of #728.

## What

Adds a **"+ Add" affordance** to the read-only Relationship tables on the item view (delivered by #734, row-removal by #739). It opens a **drawer** hosting the related list's create form, with the back-reference to the current record **preset and hidden**. On submit the new row is created through the secured context already linked to the parent, then the drawer closes and the table refreshes in place — preserving nav/filter state.

Why a drawer (not an inline draft row): the related list's **full validation and required fields** must be enforced, including required fields that are not table columns — a deliberate deviation from the design mock, recorded in the spec.

## How

- **`@opensaas/stack-core`** — new `createRelated` generic server action. It runs `db[relatedList].create` through the secured context and returns a **distinct `{ created, id?, error?, fieldErrors? }` shape** (never a single-op `success`), so a UI wrapper that redirects on `success` (the item-form pattern) can't hijack the in-place create — the same rationale as `bulkDelete`'s `{ deleted, total }` (#733) and `removeRelated`'s `{ removed }` (#739).
- **`@opensaas/stack-ui`** — `RelationshipCreateDrawer` mounted on the **existing `relationship-table-toolbar` seam** #734 reserved. It reuses the shared `useItemForm` engine + the field-component registry (no `switch` on field type), and a Radix `Dialog` for the focus trap / `Escape` / dialog aria. Named `data-slot` parts: `relationship-table-add`, `relationship-create-drawer`, `relationship-create-form`, `relationship-create-error`. The server component (`RelationshipTable`) gates the control and prepares the (serialisable) create fields via `prepareItemForm` with the back-reference stripped.

## Create-and-link access semantics (ADR-0018)

- The create runs on the **RELATED list**, so the related list's own **create access + hooks + field-level access** apply — **never the parent's**.
- The back-reference is set **on the server** from `field`/`parentId` — never trusted from the client `data` payload — so the row links to exactly the record being edited and the client cannot re-target the link.
- **Gating**: "+ Add" is shown only when (a) a back-reference exists to preset the link (list-only refs have none → hidden, mirroring #739's disconnect) and (b) the related list's `create` access is **not statically denied** (hard-deny-hides, `isOperationPotentiallyAllowed`). A filter/function-scoped denial is treated as "potentially allowed", so the control shows and a row-level denial surfaces at commit time as a **generic error** (no denied-vs-absent leak — Silent-failure contract).
- Serialisable client boundary: only the `serverAction` reference + plain serialised field configs/relationship options cross to the client — never `context` or closures.

## Scope

Only the pre-linked create drawer on the `relationship-table-toolbar` seam. No inline cell editing (#737). Integrates cleanly with existing row removal (#739) on the `relationship-table-row` seam — not duplicated.

## Tests (all green locally)

- **core** (`packages/core`): `createRelated` — happy path with the to-one back-reference preset, to-many (many-to-many) back-ref connect, access-denied → `{ created: false }` (create never attempted), and DB field-errors passthrough into the drawer shape.
- **ui** (`packages/ui`): `RelationshipCreateDrawer` — opens/creates-and-links/closes+refreshes, denial keeps the drawer open with the reason, per-field errors surface, and the required non-column field is present; plus `RelationshipTableClient` toolbar-seam gating tests.
- **e2e** (`e2e/starter-auth/08-relationship-prelinked-create.spec.ts`): adds a post pre-linked to the user via the drawer (including a required non-column `slug`), asserts it appears in the parent's Posts table AND is actually linked, and that "+ Add" is absent on tables the session cannot create on. Full suite: **59 passed** (no cross-feature regressions).
- `pnpm lint`, `pnpm format`, `pnpm manypkg fix`, `tsc` — clean.

Changeset: `minor` for `@opensaas/stack-core` + `@opensaas/stack-ui`.

## Reviewer attention

- **Access-control path**: confirm the server-side back-reference injection + the create-access gating match ADR-0018 (related list governs, parent never does).
- **Serialisable client boundary**: the drawer receives only serialised field configs + relationship options + the server-action reference.

---


borisno2 (commented):
## Review: pre-linked create drawer for relationship tables (#738)

**Verdict: APPROVE** (posted as a Comment because GitHub blocks a formal Approve from the authoring identity). The two highest-risk mechanisms — server-side back-reference injection and create-access gating — are implemented correctly and match ADR-0018. No blocking issues. Minor/non-blocking recommendations below.

### What it does
Adds a "+ Add" affordance on the read-only relationship tables (#734) that opens a Radix `Dialog` drawer hosting the related list's create form, with the back-reference preset and hidden. On submit a new `createRelated` server action creates the row on the **related** list through the secured context, already linked to the parent, then the drawer closes and the table refreshes in place.

### Top concern — create-and-link access correctness & no-leak: PASS
- `createRelated` runs `model.create` where `model = db[dbKey]` (`packages/core/src/context/index.ts:453`) — the **access-controlled delegate**, not raw prisma. The related list's own operation-level create access, hooks, and field-level access apply; the parent's never do. No prisma bypass.
- Back-reference is set **on the server**: `const data = { ...props.data }` then `data[props.field] = { connect: ... }`. Because the injection happens **after** the spread, a client-supplied `data[field]` is always overwritten — the client cannot re-target the link. To-one → `{ connect: { id } }`, to-many → `{ connect: [{ id }] }`, mirroring `removeRelated`'s disconnect logic. Correct.
- Access-denied create returns `null` → `{ created: false, error: 'Access denied or operation failed' }` (generic, no denied-vs-absent leak). Silent-failure contract honoured.

### Create-access gating: PASS
`resolveCreateForm` (`packages/ui/src/components/RelationshipTable.tsx`) hides the control unless (a) `section.backReferenceField` exists (list-only refs have none → hidden, mirroring #739) and (b) `isOperationPotentiallyAllowed(relatedListConfig.access?.operation, 'create', …)` is true. That helper denies-by-default on a missing rule and on `false`/throw, but treats a filter/function as "potentially allowed" — so a scoped denial shows the control and reverts at commit as a generic error. Matches the spec and #739's pattern.

### Distinct result shape: PASS
Returns `{ created, id?, error?, fieldErrors? }` — no `success` key. Verified against the real wrapper in `examples/starter-auth/app/admin/[[...admin]]/page.tsx:16`, which redirects only on `'success' in result && result.success`, so `createRelated` passes through and never triggers a redirect. `readCreateOutcome` maps `created:true`→success and treats any **unrecognised shape as failure** — a create is never falsely reported as succeeded.

### Serialisable client boundary: PASS
`RelationshipCreateDrawer` receives only `serverAction` (a server-action reference), serialised field configs (`prepareItemForm`), plain relationship options, and strings. No `context`, closures, or field-builder functions cross the boundary. The back-reference field is stripped from `formFields` before serialisation, so it never appears in the client form or props.

### Reuse / scope / a11y: PASS
- Reuses the shared `useItemForm` engine + `FieldRenderer` registry (no `switch` on field type) and the existing Radix `Dialog` primitive — not a parallel form stack. Radix provides focus trap, Escape, focus return, dialog aria; `role="alert"` on the error; named `data-slot` parts throughout. Splitting into `RelationshipCreateForm` so it remounts on each open (resetting `useItemForm`) is a clean touch.
- Scope is clean: only the `relationship-table-toolbar` seam. No inline cell editing (#737). Row removal (#739) is untouched and integrated alongside, not duplicated.
- Changeset correctly scoped: core minor + ui minor.

### Non-blocking recommendations
1. **(Test gap — security property) Add a test that the client cannot re-target the link.** The core happy-path test sends `data: { title, content }` with no `author`, so it proves the server *sets* the back-ref but not that it *overwrites* a hostile one. Add a case passing `data: { title, author: 'evil-id' }` and assert `create` is called with `author: { connect: { id: parentId } }`. This is the single most security-relevant property and is currently guaranteed by code but not locked by a test.
2. **(Minor perf) `resolveCreateForm` eagerly runs `prepareItemForm`** (bounded relationship-option fetches for the related list's own relationship fields) on every parent item-view render when create is allowed — even if the drawer is never opened. Bounded/take-limited, so not a blocker; consider deferring to drawer-open if it surfaces in profiling.
3. **(Nit) Field-level create access is enforced at commit, not in which fields the drawer displays** — a field the session can't write is shown but stripped by the write pipeline. Consistent with the existing standalone create form (no leak, enforced server-side); flagging only to confirm it's intended.
4. **(Nit) `createRelated` doesn't validate that `field` is a relationship field**, and skips injection if only one of `field`/`parentId` is supplied. Unreachable from the UI (the drawer only mounts when `backReferenceField` exists and always sends both), and any malformed direct call degrades to a normal access-governed create + generic DB error — harmless, but a defensive guard would make the generic action more robust.

Tests cover the create-and-link happy path (to-one + to-many), access-denied hidden-vs-commit, field-error passthrough, drawer open/close/refresh, and an e2e that opens the drawer, creates with a required non-column field, and asserts the row appears **and** is actually linked. Solid coverage.

_Reviewed strictly from the PR diff, its description, and the surrounding code it touches._

---

PR #760, opened 2026-07-20
feat(core,ui): inline cell editing in relationship tables (#737)

borisno2:
Implements #737. Part of #728.

## What

Makes individual cells in a to-many **Relationship table** (item view, #734) editable in place, on the reserved `relationship-table-cell` seam. Integrates alongside — not on top of — row removal (#739/#756) and the pre-linked create drawer (#738/#757).

- Click an editable cell → an in-place editor opens (Enter/Space on the focused display trigger also works).
- **Enter or blur commits, Escape cancels.**
- Each commit is a **single-field update on the RELATED row through the secured context**; the update is optimistic and the table refreshes on success.
- A Silent failure (access denied / row gone) or a validation error **reverts the cell to its original value with a visible reason** (`role="alert"`), including inline field errors.
- Committed values **re-render through the Cell registry**, so select cells stay coloured badges.
- Non-editable cells keep click-to-navigate; main list tables are unchanged (Relationship-table only).

## How

**`@opensaas/stack-core`**
- New `updateRelated` server action with a **distinct result shape** `{ updated, error?, fieldErrors? }` (never a single-op `success`), mirroring `removeRelated`/`createRelated` so a redirect-on-`success` wrapper can't hijack an in-place edit. It runs `db[relatedList].update({ data: { [field]: value } })` — one field, through the secured delegate.
- `checkFieldAccess` is re-exported from `@opensaas/stack-core/internal` so the UI reuses the **single canonical** field-access evaluator rather than a parallel one.

**`@opensaas/stack-ui`**
- `RelationshipTable` (server) computes `editableColumns` from the **related list's own** operation-level update access (table gate, mirroring `isOperationPotentiallyAllowed`) and each field's update field-access (`isFieldPotentiallyWritable`), excluding relationships, virtual/computed fields, passwords and system columns.
- `RelationshipTableClient` gains an additive `editableColumns` prop.
- New `RelationshipTableCell` reuses the **Cell registry** (`CellRenderer`) for its read display and the **field-component registry** (`FieldRenderer`, `mode="edit"`) for the editor — no bespoke per-type input, no switch on field type. Named Slots: `relationship-table-cell-display`, `relationship-table-cell-editor`, `relationship-table-cell-edit-trigger`, `relationship-table-cell-error`. Theme tokens only; focus management + Enter/Escape/blur keyboard handling; blur into a portalled overlay (e.g. a select dropdown) does not prematurely commit.

## Edit access semantics

Every edit is an operation on the **related list**, scoped to that list's access (ADR-0018), never the parent's:

- **Field-level:** a field the session cannot update (statically-false field access) renders read-only, no affordance. Filter/item-dependent field access is treated as "potentially writable" — the affordance shows and a row-level denial surfaces at commit as a revert (no denied-vs-absent leak).
- **Operation-level:** a hard `update: false` on the related list makes the whole table non-editable.
- **Row/filter-scoped denials:** surface at commit via the revert path with a generic message.

## Tests

- **Core** (`packages/core/tests/context.test.ts`): `updateRelated` happy path (one field updated), access-denied → `{ updated: false }` generic reason, validation error surfaced.
- **UI** (`RelationshipTableCell.test.tsx` + `RelationshipTableClient.test.tsx`): commit-on-Enter + optimistic refresh, revert-with-reason on Silent failure, inline field-error, Escape cancel, no-op unchanged commit, not-writable → read-only (no affordance), editable cell edits instead of navigating while non-editable cells still navigate, select re-renders as a badge.
- **e2e** (`e2e/starter-auth/08-relationship-inline-edit.spec.ts`): real inline commit persists through the secured context (verified on the Post list), and a validation-rejected edit reverts with a reason while the DB is unchanged. Updated the #734 nav test to the new click-to-edit behaviour.

### Local results
- `packages/core` + `packages/ui` unit suites: **all green** (853 + 491).
- Full root **e2e suite: green** (61 passed; the only 2 flakes were pre-existing `06-filter-builder` URL-timing flakes, unrelated, passed on retry).
- `pnpm lint`, `prettier`, `manypkg fix`, and `tsc` (core+ui build, and the starter-auth `next build`) all clean.

## Changeset

`minor` for `@opensaas/stack-core` and `@opensaas/stack-ui` (additive server-action shape + additive UI prop/component/Slots).

## Notes for reviewers / risks
- **Field-level access path:** editability is decided server-side per column via `checkFieldAccess` with no `item` (throws → treated as potentially-writable so item-dependent access stays editable and is re-checked at commit). Worth a look.
- **Serialisable boundary:** the client cell receives only the server-action reference + serialised field/cell configs + primitive value/flags — never `context` or closures.
- To keep the suite robust as it accumulates >50 signups (the user list page cap), both #737 and #738 e2e specs now open the user's item view directly via the session user id instead of hunting the paginated `/admin/user` row; this also removed an `a:has-text("Edit")` collision with post titles.
- Shared-file edits (`packages/ui/src/components/RelationshipTableClient.tsx`) are additive, per the concurrent #736 lane note.

---

### DIFF
--- a/.changeset/inline-cell-editing-relationship-tables.md
+++ b/.changeset/inline-cell-editing-relationship-tables.md
@@ -1,0 +1,31 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Add inline cell editing to admin Relationship tables
+
+Cells in a to-many Relationship table on the item view are now editable in place.
+Click a cell to edit it, commit with Enter or blur, cancel with Escape. Each
+commit is a single-field update on the **related** row through the secured
+context, so the related list's own operation- and field-level update access plus
+its hooks/validation apply — never the parent's. The update is optimistic and
+reverts, with a visible reason, on a Silent failure (access denied / row gone) or
+a validation error (inline field errors surface too). Committed values re-render
+through the Cell registry, so select cells stay coloured badges.
+
+A field the session cannot write — or a table whose related-list update access is
+statically denied — renders read-only with no edit affordance; row-level
+(filter-scoped) denials surface at commit as a revert. Non-editable cells keep
+click-to-navigate; main list tables are unchanged (this is Relationship-table
+only).
+
+- `@opensaas/stack-core`: the generic server action gains a distinct
+  `updateRelated` result shape (`{ updated, error?, fieldErrors? }`), and
+  `checkFieldAccess` is exposed on `@opensaas/stack-core/internal` so the UI can
+  decide the edit affordance without a parallel field-access evaluator.
+- `@opensaas/stack-ui`: `RelationshipTableClient` accepts `editableColumns`; the
+  editable cell reuses the field-component registry for its editor and the Cell
+  registry for its display (new Slots: `relationship-table-cell-display`,
+  `relationship-table-cell-editor`, `relationship-table-cell-edit-trigger`,
+  `relationship-table-cell-error`).
--- a/e2e/starter-auth/05-item-view.spec.ts
+++ b/e2e/starter-auth/05-item-view.spec.ts
@@ -85,7 +85,13 @@ test.describe('Item view layout', () => {
     await expect(sessions.locator('[data-slot="relationship-table-empty"]')).toBeVisible()
   })
 
-  test('clicking a Relationship-table row navigates to the related record', async ({ page }) => {
+  test('clicking an editable relationship-table cell edits in place instead of navigating', async ({
+    page,
+  }) => {
+    // Inline cell edit (#737) supersedes row-click navigation on EDITABLE cells:
+    // a Post's columns are all editable for its author, so clicking one opens the
+    // inline editor and the item view stays put (main list tables keep
+    // click-to-navigate; this is Relationship-table-only).
     const testUser = generateTestUser()
     await signUp(page, testUser)
     await createPost(page, { title: 'Nav Post', slug: 'nav-post', viewCount: 1 })
@@ -95,9 +101,16 @@ test.describe('Item view layout', () => {
     await page.locator('tr', { hasText: testUser.email }).locator('a:has-text("Edit")').click()
     await page.waitForURL(/\/admin\/user\/[^/]+$/, { timeout: 10000 })
 
-    await page.locator('[data-slot="relationship-table-row"]', { hasText: 'Nav Post' }).click()
-    await page.waitForURL(/\/admin\/post\/[^/]+$/, { timeout: 10000 })
-    await expect(page.locator('input[name="title"]')).toHaveValue('Nav Post')
+    const posts = page.locator('[data-slot="relationship-table"]', {
+      has: page.getByRole('heading', { name: 'Posts', exact: true }),
+    })
+    await posts
+      .locator('[data-slot="relationship-table-cell-edit-trigger"]', { hasText: 'Nav Post' })
+      .click()
+
+    // The inline editor opens and the URL is unchanged (no navigation).
+    await expect(posts.locator('[data-slot="relationship-table-cell-editor"] input')).toBeVisible()
+    await expect(page).toHaveURL(/\/admin\/user\/[^/]+$/)
   })
 
   test('a list with no to-many relationship renders a single details card', async ({ page }) => {
--- a/e2e/starter-auth/08-relationship-inline-edit.spec.ts
+++ b/e2e/starter-auth/08-relationship-inline-edit.spec.ts
@@ -1,0 +1,124 @@
+import { test, expect, type Page } from '@playwright/test'
+import { signUp, generateTestUser } from '../utils/auth.js'
+
+/**
+ * Relationship-table inline cell edit (issue #737, ADR-0018).
+ *
+ * The User item view renders a `posts` Relationship table whose cells are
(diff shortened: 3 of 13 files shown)


borisno2 (commented):
## Review: inline cell editing in relationship tables (#737)

**Verdict: APPROVE** (posted as a Comment review — the integration identity authored the PR, so a formal Approve is blocked; the verdict is explicit here.)

The security-critical property — **render-gate as a cosmetic affordance, real enforcement at commit through the secured context** — holds. I traced it end to end rather than trusting the description.

### Field-level access path (top concern) — verified sound

- `updateRelated` runs `model.update(...)` where `model = db[dbKey]` (`context/index.ts` ~L428) — the **access-controlled delegate**, not raw prisma. So the RELATED list's operation-level update access, field-level update access, hooks and validation all apply — never the parent's. ✓
- **Commit-time re-check is genuine, not just a render gate.** Field-level update denial routes through `filterWritableFields`, which per #568 **throws `ValidationError('Cannot update "<field>": field-level access denied.')`** rather than silently dropping the field. `updateRelated`'s catch converts that to `{ updated: false, error, fieldErrors }` → the cell reverts with a visible reason. A statically-false field access never reaches here because `resolveEditableColumns` omits the column (no affordance). An item-dependent field access throws in `isFieldPotentiallyWritable` (no `item` passed) → treated as potentially-writable → affordance shown → **re-checked per row at commit**. This is exactly the fail-safe behaviour required. ✓
- **Operation-level `update: false`** → `isOperationPotentiallyAllowed` returns false → `resolveEditableColumns` returns `[]` → whole table read-only. ✓
- **Filter/row-scoped denial** → secured `model.update` returns `null` → `{ updated: false, error: 'Access denied or operation failed' }` (generic) → no denied-vs-absent leak. ✓
- The throw-handling **asymmetry** is correct and intentional: operation-gate throw → deny (fail closed at table level); field-gate throw → potentially-writable (item-dependent access legitimately needs the row, re-checked at commit). Both fail safe.
- **No privilege escalation via a crafted client call**: even if a client invokes `updateRelated` with a non-listed field (e.g. `password`), the secured delegate still enforces that list's field access + Zod + hooks — the same authority as the normal `update` action. `editableColumns` is purely a UI affordance; the server is authoritative. Correct posture.

### Secured single-field update & result shape — verified
- Updates exactly one field (`data: { [props.field]: props.value }`); test asserts `post.update` called with `{ id }` and `delete` not called.
- Distinct `{ updated, error?, fieldErrors? }` shape (never `{ success }`), so the redirect-on-`success` item-form wrapper can't hijack an in-place edit. `readUpdateOutcome` treats any unrecognised shape as failure (never a false success). ✓

### Serialisable boundary / reuse / scope — verified
- Client cell receives only the server-action reference + serialised `field`/cell configs + primitive `value`/flags — no `context`, no closures. ✓
- Display via the #729 Cell registry (`CellRenderer`), editing via the field-component registry (`FieldRenderer`, `mode="edit"`) — no bespoke input, no `switch` on field type. Unregistered field type → read-only fallback (nice touch). Enter/blur commit, Escape cancel, optimistic + revert, blur-into-portal guarded (`isInOverlay`). ✓
- Scope is inline-edit only; row removal (#739) and create drawer (#738) are untouched (additive `editableColumns` prop; non-editable cells keep click-to-navigate, covered by `RelationshipTableClient.test.tsx`). ✓
- `checkFieldAccess` exposure via `internal.ts` reuses the single canonical evaluator; type-safe (no `any`, internal narrowing only). Theme tokens + named data-slots present. Changeset is core minor + ui minor. ✓

### Test-infra change — legitimate, not a regression-hider
The switch in `08-relationship-prelinked-create` and `05-item-view` to open the user's item view **directly by session user id** (via `/api/auth/get-session`) instead of hunting the paginated `/admin/user` row is a real robustness fix: the list caps at 50 rows and the suite accumulates >50 signups, plus it removed the `a:has-text("Edit")` collision with post titles. Assertions are unchanged — the specs still verify the actual create/edit behaviour (and persistence is re-confirmed via the Post list filtered by unique slug). Not a workaround masking a regression. ✓

### Minor (non-blocking) observations
1. **No-op detection is reference equality** — `commit()` skips the round-trip only when `draft === displayValue` (`RelationshipTableCell.tsx`). For scalar fields (text/select/number) this is correct, but `json` columns are *not* excluded from `editableColumns`, and their draft is a fresh object reference, so an unchanged `json` cell will always round-trip on Enter/blur. Harmless (the secured update just rewrites the same value) but a wasted request; a value-equality/normalised compare, or excluding `json` from inline-edit for now, would tidy it.
2. **Lost e2e coverage of row-click navigation** — `05-item-view` previously asserted a relationship-row click *navigates*; it now asserts an editable cell edits in place (correct, since all Post columns are editable for the author). Row navigation for a non-editable cell is still covered at unit level (`RelationshipTableClient.test.tsx`), so this is acceptable, just noting the e2e trade-off.
3. **CI**: `e2e` is green; the `test` (unit) job was still in progress when I reviewed — worth a glance that it landed green (author reports 853 + 491 local). The flagged `06-filter-builder` flake is a pre-existing URL-timing issue, unrelated to this diff.

None of the above blocks merge. The field-level enforcement — the property that most deserved scrutiny — is correctly implemented as commit-time re-check backed by a throwing `filterWritableFields`, with the UI gate purely advisory. Nice work.

---

Issue #752, opened 2026-07-19
Admin UI: bound relationship-table row fetch with a limit + "showing N of M" footer


borisno2:
## Parent

Follow-up from #728 (Generic reusable admin UI). Surfaced during review of #751 (item view layout, #734) — non-blocking, filed so #751 could merge.

## Problem

The read-only Relationship tables introduced in #734 fetch related rows with **no `take`/limit**: `buildItemViewInclude` includes the to-many rows unbounded, and the totals footer's count is `rows.length` (all included rows). A record with many related rows (e.g. a long-lived User's sessions/posts/orders) loads and renders **every** related row each time its edit page opens — memory, serialized-payload, and DOM cost that the main list view avoids via pagination. The relationship tables do not paginate.

This was acceptable for #734's read-only happy path but is a scalability risk worth closing before these tables see large datasets.

## What to build

- Add a bounded fetch (`take`) to the relationship-table include, with a sensible default limit (and optional per-relationship override in item-view config).
- Surface a "showing N of M" affordance in the totals footer (M = access-scoped count via the secured context), keeping the row count always shown.
- Consider pagination / "view all" navigation to the related list pre-filtered to the parent, if cheap; otherwise document the cap.
- Counts and rows must remain access-scoped through the secured context.

## Secondary note (from same review)

Default relationship-table columns come from the related list's curation, which for the auth `Account` list includes token fields (`accessToken`/`refreshToken`/`idToken`) — field-level access control still governs the actual values, but the starter-auth example surfaces these columns by default on the User edit page. Worth curating the example's Account columns (or documenting the default) so the example doesn't model exposing token columns.

## Acceptance criteria

- [ ] Relationship-table related-row fetch is bounded by a default `take` (per-relationship overridable)
- [ ] Footer shows "showing N of M" with M as the access-scoped count; row count still always shown
- [ ] Rows and counts remain access-scoped via the secured context
- [ ] starter-auth example Account relationship columns curated (or the default documented) so token columns aren't surfaced by default
- [ ] Tests cover the bounded fetch and the N-of-M footer; changeset included

## Blocked by

None — #734 is (being) merged.

Issue #732, opened 2026-07-19
Admin UI: to-many relationship columns — count cells, count sort, count filter


borisno2:
## Parent

Spec: #728

## What to build

A to-many relationship field used as a list column shows the access-visible related count (fetched within the same query), sorts by relation count when its header is clicked, and is filterable with numeric comparisons on the count (e.g. `orders:>5`) via its Filter spec. Virtual fields render through their Cell but expose no sort affordance and no Filter spec.

## Acceptance criteria

- [ ] Count cells show only what the session may see, with no extra per-row queries
- [ ] Header sort orders by relation count; virtual field headers show no sort affordance
- [ ] Count comparisons work in the Filter builder and in shared URLs
- [ ] Virtual fields are excluded from filtering and sorting everywhere (no error paths, affordances absent)
- [ ] E2e covers count display, count sort, and count filter on an example list; changeset included

## Blocked by

- #729
- #730

PR #764, opened 2026-07-20
feat(core,ui): to-many relationship count columns — count cells, count sort, count filter (#732)

borisno2:
Implements #732. Part of #728.

## What

To-many relationship fields used as list columns are now first-class:

- **Count cells** — a to-many column renders the **access-visible related count** (e.g. a User's post count), fetched in the *same* secured query.
- **Count sort** — clicking the column header sorts rows by relation count (`orderBy: { posts: { _count } }`).
- **Count filter** — the field's Filter spec offers numeric comparisons on the count (`posts:>5`), working in the filter builder and in shared URLs.
- **Virtual fields** render via their Cell but are excluded from sorting and filtering (no sort affordance, no Filter spec) — so a `?sort=<virtual>` can no longer reach Prisma.

To-one relationship columns are unchanged (linked Item label; label search filter).

## How

- **Access-scoped counts (single query, no per-row queries):** counts come from a filtered Prisma `_count` — `include: { _count: { select: { posts: { where: <related query access> } } } }`. New core helper `buildRelationshipCountSelect` folds each related list's operation-level `query` access into the count's `where`, mirroring how `buildIncludeWithAccessControl` scopes relation includes. A fully-denied related list is omitted → the cell shows `0` (never a leak). This also **removes the previous unbounded per-row fetch** of every related row for to-many columns (aligns with the #752 bounding concern).
- **Count filter:** Prisma cannot compare a relation `_count` in a `where` (verified empirically against Prisma 7). So the to-many relationship's `getFilterSpec` emits a structured marker (`RELATIONSHIP_COUNT_FILTER_KEY`) instead of a where fragment; `resolveRelationshipCountFilters` resolves it — through the **secured context** — into an access-scoped `{ id: { in } }` before the list query runs. Zero-count parents are handled correctly (fetched with count `0` and filtered in memory), and a denied related list resolves to count `0` with no query.
- **Count sort** passes `orderBy: { <rel>: { _count: dir } }` straight through the secured context.
- New core exports: `buildRelationshipCountSelect`, `resolveRelationshipCountFilters`, `isToManyRelationshipField`, `RELATIONSHIP_COUNT_FILTER_KEY` (+ `RelationshipCountFilterMarker`). Cell/registry edits are additive; no `switch` on field type — the count Cell resolves through the existing registry (`ui.cell` override still wins).

## Access-scoping

Every count (cell, sort input, and filter) flows through the secured context. The count's `where` is the related list's own `query` access, so a session only ever counts rows it may read; a closed related list yields `0`, not a leak. Count **sort** orders by the unfiltered relation `_count` (Prisma cannot filter an `orderBy` count) — a documented limitation that affects only row ordering, never displayed values.

## Bounding

Count cells and count sort issue **no** extra queries (the `_count`/`orderBy` ride the row query). A count **filter** issues exactly one extra aggregate read per count token (not per row) to resolve matching ids; that read fetches parent ids + counts under parent access. This is the deliberate trade for an access-correct count filter given Prisma's lack of relation-count `where` support.

## Tests

- **Core** (`relationship-count.test.ts`, `filter.test.ts`): filtered `_count` select folds in related access; denied list omitted; markers resolve to access-scoped `{ id: { in } }` incl. zero-count and denied cases; to-many Filter spec emits count comparisons (to-one still label search).
- **UI** (`ListView.test.tsx`, `Cells.test.tsx`, `CellRenderer.test.tsx`): to-many column resolves to a count via `_count` (never a row include); count sort builds nested `orderBy`; virtual sort ignored; count Cell renders count/zero; per-field `ui.cell` override still wins.
- **e2e** (`09-relationship-count-columns.spec.ts`): count display, count sort (asc/desc), count filter via shared URL and via the filter builder, on the User list's `posts` column.

Local: `pnpm lint` clean, `tsc`/`pnpm build` clean, core 865 + ui 496 unit tests pass, full e2e suite 66/66 pass (4 new + no regressions).

Changeset: `minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`.

---

### DIFF
--- a/packages/core/src/access/relationship-count.ts
+++ b/packages/core/src/access/relationship-count.ts
@@ -1,0 +1,266 @@
+import type { Session, AccessContext, PrismaFilter } from './types.js'
+import type { OpenSaasConfig, ListConfig, FieldConfig } from '../config/types.js'
+import type { FilterOperator, RelationshipCountFilterMarker } from '../filter/types.js'
+import { RELATIONSHIP_COUNT_FILTER_KEY } from '../filter/types.js'
+import { checkAccess, getRelatedListConfig } from './engine.js'
+import { getDbKey } from '../lib/case-utils.js'
+
+/**
+ * Access-scoped to-many relationship counts for the admin list view (issue
+ * #732).
+ *
+ * A to-many relationship column shows the count of the related rows the session
+ * may see — fetched in the SAME query as the row via Prisma's filtered
+ * `_count` (`_count: { select: { orders: { where: <access filter> } } }`), so no
+ * per-row query is issued and the count can never include rows the session
+ * cannot read. This module is the single place the related list's
+ * operation-level `query` access is folded into that `_count`, mirroring how
+ * `buildIncludeWithAccessControl` folds it into relation includes.
+ *
+ * It also resolves the count Filter spec's markers: Prisma cannot compare a
+ * relation count in a `where`, so a to-many relationship's Filter spec emits a
+ * {@link RELATIONSHIP_COUNT_FILTER_KEY} marker that
+ * {@link resolveRelationshipCountFilters} turns into an access-scoped
+ * `{ id: { in } }` before the query runs — never leaking counts of related rows
+ * the session cannot see.
+ */
+
+type CountArgs = {
+  session: Session | null
+  context: AccessContext
+}
+
+/**
+ * Whether a field is a to-many relationship — the only field kind that carries a
+ * relationship count (a to-one relationship has at most one related row).
+ */
+export function isToManyRelationshipField(field: FieldConfig | undefined): boolean {
+  return (
+    field?.type === 'relationship' &&
+    'many' in field &&
+    field.many === true &&
+    'ref' in field &&
+    typeof field.ref === 'string' &&
+    field.ref.length > 0
+  )
+}
+
+/** The per-relation entry the count `_count.select` uses for one relationship. */
+type CountAccessEntry =
+  | { kind: 'all' } // related list fully readable → count every related row
+  | { kind: 'scoped'; where: PrismaFilter } // count only rows matching the access filter
+  | { kind: 'denied' } // related list not readable at all → count is always 0
+
+/**
+ * Resolve the related list's operation-level `query` access for one to-many
+ * relationship into the entry its `_count` select needs. Returns `null` when the
+ * field is not a resolvable to-many relationship.
+ */
+async function relationshipCountAccessEntry(
+  field: FieldConfig | undefined,
+  args: CountArgs,
+  config: OpenSaasConfig,
+): Promise<CountAccessEntry | null> {
+  if (!field || !isToManyRelationshipField(field) || !('ref' in field)) {
+    return null
+  }
+  const ref = field.ref
+  if (typeof ref !== 'string') return null
+  const related = getRelatedListConfig(ref, config)
+  if (!related) return null
+
+  const queryAccess = related.listConfig.access?.operation?.query
+  const result = await checkAccess(queryAccess, { session: args.session, context: args.context })
+
+  if (result === false) return { kind: 'denied' }
+  if (typeof result === 'object') return { kind: 'scoped', where: result }
+  return { kind: 'all' }
+}
+
+/**
+ * Build the object to place at `include._count.select` for a list's to-many
+ * relationships, with each related list's `query` access folded into the
+ * per-relation `where` so the returned counts are access-scoped. A relationship
+ * whose related list is fully denied is omitted (its count renders as 0). Returns
+ * `undefined` when the list has no countable to-many relationships.
+ */
+export async function buildRelationshipCountSelect(
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
+  listConfig: ListConfig<any>,
+  args: CountArgs,
+  config: OpenSaasConfig,
+): Promise<Record<string, unknown> | undefined> {
+  const select: Record<string, unknown> = {}
+  for (const [fieldName, field] of Object.entries(listConfig.fields)) {
+    const entry = await relationshipCountAccessEntry(field, args, config)
+    if (entry === null || entry.kind === 'denied') continue
+    select[fieldName] = entry.kind === 'scoped' ? { where: entry.where } : true
+  }
+  return Object.keys(select).length > 0 ? select : undefined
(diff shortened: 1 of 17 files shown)


packages/core/src/access/relationship-count.ts
```diff
@@ -0,0 +1,266 @@
...
+  // Related list fully denied → every parent's access-visible count is 0.
+  if (entry.kind === 'denied') {
+    return matchesCount(0, marker.operator, marker.value) ? {} : { id: { in: [] } }
+  }
+
+  const delegate = asCountDelegate(args.context.db[getDbKey(listKey)])
+  if (!delegate) return {}
+
+  const countSelect = entry.kind === 'scoped' ? { where: entry.where } : true
+  const rows = await delegate.findMany({
```
borisno2, line 194: Nit (efficiency, non-blocking): this `findMany` has no `select`, so resolving the count filter pulls **every scalar column of every parent row** for the whole list, only to read `id` + `_count`. A `select: { id: true, _count: { select: { [fieldName]: countSelect } } }` would fetch just what's used. The unbounded full-list scan itself is a documented, accepted trade (Prisma has no relation-count `where`); this only trims per-row width. Note core's read pipeline warns-and-ignores `select` in some paths — verify the secured context honours it here before switching, else leave as-is.

---

packages/core/src/access/relationship-count.ts
```diff
@@ -0,0 +1,266 @@
...
+  }
+
+  const resolvedMembers: Array<Record<string, unknown>> = []
+  for (const member of members) {
+    const found = findMarker(member)
+    if (!found) {
+      resolvedMembers.push(member)
+      continue
+    }
+    resolvedMembers.push(
```
borisno2, line 256: Robustness (low, non-blocking): a matched member is **replaced** wholesale by the resolved `{ id: { in } }`. This is only safe because the filter engine guarantees each AND member (and the no-AND single object) carries exactly one field condition, so a marker never shares a member with sibling keys. That invariant holds today and is tested, but it's an implicit cross-module coupling — if the engine ever merged multiple conditions into one object, the siblings here would be silently dropped. Consider spreading the non-marker keys of `member` into the result (or asserting single-key) to harden against that. Not required for this PR.

---

borisno2 (commented):
## Review — PR #764: to-many relationship count columns (#732)

**Verdict: APPROVE** (submitted as a Comment review — a formal Approve is blocked because the review identity authored the PR). No blocking correctness or security issues found. Three non-blocking notes below (two inline).

### What it does
To-many relationship list columns now render an **access-scoped related count**, are sortable by relation `_count`, and filterable by numeric count comparisons (`posts:>5`). Virtual and to-one columns lose their sort affordance (no orderable scalar). Additive across core + ui, no `switch` on field type.

### Access-scoping of the display count — the load-bearing property ✅
Traced `buildRelationshipCountSelect` → `relationshipCountAccessEntry` → `checkAccess`. For each to-many field it folds the **related** list's operation-level `query` access into the per-relation `_count.where`:
- object filter → `{ where: <access filter> }` (scoped)
- `true` → bare `true` (fully readable)
- `false`/**absent access block** → **denied → omitted** from the select

I verified in `engine.ts` that `checkAccess(undefined)` returns `false` (default-deny), so a related list with no `query` access is denied → its `_count` is omitted → `readRelationshipCount` returns `0`. A session can never count rows it may not read, and a closed list shows `0`, never a true total. The count rides the **same secured `context.db.findMany`** as the row (`include._count = { select: countSelect }`) — no raw prisma total, no per-row fetch. This also removes the previous unbounded per-row include of every related row. Correct and secure.

### Count filter through the secured context ✅
`resolveRelationshipCountFilters` runs as a pre-pass: it detects `_countFilter` markers (only honoured on genuine to-many fields — `isToManyRelationshipField` guards the read, so the marker can't be injected onto a scalar), and for each resolves via **one** secured `findMany` per token computing access-scoped `_count` per parent, keeping ids that satisfy the comparison → `{ id: { in } }`. Both parent access (secured `findMany`) and related access (folded `_count.where`) are respected. Zero-count rows and the fully-denied case (count `0`, **no query issued**) are handled and tested. One aggregate read per token, not per row — the cost note is accurate.

### Count sort ✅
`orderBy: { <rel>: { _count: dir } }` through the secured context. The orderBy count is **unfiltered** (Prisma limitation) — but the **displayed** value always comes from the filtered `include._count`, a separate computation, so ordering can differ from a scoped count without ever leaking a total. Documented in the PR body and changeset. Correct separation.

### Virtual / to-one sort suppression ✅
`isSortableField` (server) replaces the old `sort.field in listConfig.fields` check, closing the latent `?sort=<virtual>` → Prisma-throw path (a virtual field *is* in `fields`, so the old check let it through). `isSortableColumn` (client) mirrors it and `virtual` is now serialized. Only virtual + to-one are suppressed; scalars and to-many (count) stay sortable.

### Other checks ✅
- Registry precedence: count Cell is a branch inside `RelationshipCell` resolved via the type registry; per-field `ui.cell` override still wins (tested in `CellRenderer.test.tsx`). No `switch`.
- Serializable boundary: `_count` deleted from the item, to-many resolved to a plain number before crossing to the client.
- No `any`/casts exposed externally — `ListConfig<any>` and test-only `as any` are eslint-disabled with justification, matching existing patterns; production code uses `unknown` + narrowing.
- Theme tokens (`text-primary`, `hover:bg-muted/70`, `tabular-nums`, `text-right`), named slot `cell-relationship-count`.
- Changeset present: `@opensaas/stack-core` minor + `@opensaas/stack-ui` minor.
- Scope contained: list-view count columns + the sort fix only; item-view relationship tables untouched.

### Non-blocking findings
1. **Efficiency (inline, minor):** the count-filter `findMany` has no `select`, over-fetching all scalar columns of every parent row to read only `id` + `_count`. `select: { id: true, _count: {...} }` would trim it (verify the secured context honours `select` first).
2. **Robustness (inline, low):** the marker member is replaced wholesale — safe only under the filter engine's one-condition-per-member invariant; a defensive spread/assert would harden the cross-module coupling.
3. **Process (not code):** CI on this PR head shows only Vercel (docs) + GitGuardian (skipped) — there is **no unit/e2e test gate running on the PR**, so the "core 865 / ui 496 / e2e 66/66" green rests on the author's local run and isn't independently verifiable here. Recommend confirming the test workflow runs before merge. `mergeable_state` is `blocked` (self-approval / required review).

Solid, well-tested, security-correct work. Approving.

---

Issue #765, opened 2026-07-20
Minor hardening: relationship count-filter over-fetch + marker-member replace robustness


borisno2:
Non-blocking nits from the review of #732 (PR #764). The feature is correct and secure; these are small efficiency/robustness tidy-ups in `packages/core/src/access/relationship-count.ts`.

## 1. Count-filter pre-pass over-fetches (efficiency, minor)

`resolveRelationshipCountFilters` runs a secured `findMany` per count-filter token to compute access-scoped `_count` per parent, but issues it with **no `select`** — so it materialises all scalar columns of every parent row just to read `id` + `_count`.

- [ ] Add `select: { id: true, _count: { select: { <rel>: { where: <access> } } } }` to trim it — first verify the secured `context.db` `findMany` honours `select` (and that field-level read access still applies through it). Keeps the access-scoping identical, just narrows the projection.

## 2. Marker member replaced wholesale (robustness, low)

When resolving a `_countFilter` marker, the matched AND-member is **replaced** wholesale by the resolved `{ id: { in } }`. This is safe only because the filter engine currently guarantees each AND member (and the no-AND single object) carries exactly one field condition, so a marker never shares a member with sibling keys. That invariant holds today and is tested, but it's an implicit cross-module coupling — if the engine ever merged multiple conditions into one object, the sibling keys here would be silently dropped.

- [ ] Harden by spreading the non-marker keys of the member into the result (or asserting single-key) so a future engine change can't silently drop sibling conditions.

## Notes
Both low priority, non-blocking for #728. No security impact — the access-scoping of the count (display, sort, and filter) was verified correct in review.

PR #771, opened 2026-07-20
fix(core): harden relationship count-filter over-fetch + marker-member replace

borisno2:
Implements #765. Part of #728.

Two non-blocking hardening items from the #732 review, both in `packages/core/src/access/relationship-count.ts`.

## 1. Count-filter over-fetch (efficiency)

The issue asked to add `select: { id: true, _count: { select: { <rel>: { where: <access> } } } }` to `resolveRelationshipCountFilters`' secured per-token `findMany`, *first verifying the secured `context.db` `findMany` honours `select`*.

**It does not.** The secured `findMany` accepts `select` only to make its no-op visible — it `warns-and-ignores` it (`warnIfSelectIgnored` in `context/index.ts`, and the "Narrowing Reads" note in `packages/core/CLAUDE.md`) and never forwards it to the underlying Prisma call; only `include`/fragment `query` reach Prisma, and neither can drop scalar columns. Forcing a `select` here would therefore be a silent no-op that also trips the ignore-warning, not a real projection.

Per the issue's instruction, the projection is **left as-is** and the reasoning is documented in a code comment (and here). Access-scoping is unchanged — it lives entirely in the filtered `_count.where` include. A real fix would require the read pipeline to honour `select` first, which is out of scope.

## 2. Marker-member replace robustness

When resolving a `_countFilter` marker, the matched AND-member was replaced **wholesale** by the resolved `{ id: { in } }`. This is safe only under the filter engine's current one-condition-per-member invariant. Hardened by spreading the member's non-marker sibling keys into the result via a small `mergeResolvedMember` helper, so a future engine change that merges multiple conditions into one member can't silently drop siblings. Behaviour is byte-identical for the current single-condition case (siblings empty → returns the resolved fragment). On the contrived chance a sibling shares the resolved `id` key, both are ANDed rather than one overwriting the other.

## Tests

Extended `relationship-count.test.ts`:
- the secured read is issued with only the filtered `_count` include and **no** `select` projection (access-scoped counts preserved — existing denied→0 and zero-count tests still pass);
- a sibling condition co-present in the same member as the marker is preserved (not dropped);
- a co-present sibling `id` condition is ANDed with the resolved id constraint.

## Verification
- `pnpm build` (tsc) — clean
- core vitest — 871 passed (42 files), incl. 12 in `relationship-count.test.ts`
- `pnpm lint` — 0 errors (3 pre-existing warnings, unrelated)
- `pnpm manypkg fix` + `pnpm format` — clean
- Changeset: `@opensaas/stack-core` **patch** (bug-fix/hardening)

No security impact — counts/reads stay access-scoped through the secured `context.db`; no raw Prisma introduced.

---

### DIFF
--- a/.changeset/quiet-otters-hum.md
+++ b/.changeset/quiet-otters-hum.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Harden relationship count-filter resolution: preserve any sibling conditions co-present in a `_countFilter` marker's AND-member instead of replacing it wholesale, and document why the secured count read intentionally keeps its full projection (`context.db` does not honour `select`).
--- a/packages/core/src/access/relationship-count.test.ts
+++ b/packages/core/src/access/relationship-count.test.ts
@@ -199,4 +199,71 @@ describe('resolveRelationshipCountFilters', () => {
     )
     expect(resolved).toEqual({ AND: [{ name: { contains: 'ada' } }, { id: { in: ['u2'] } }] })
   })
+
+  it('reads with only the filtered `_count` include — no over-fetching `select` projection', async () => {
+    const config = makeConfig()
+    const findMany = vi.fn((_args: unknown) =>
+      Promise.resolve([{ id: 'u2', _count: { posts: 7 } }]),
+    )
+    await resolveRelationshipCountFilters(
+      marker('gt', 5),
+      config.lists.User,
+      'User',
+      { session: null, context: makeContext(findMany) },
+      config,
+    )
+    // The secured read is issued with the access-scoped `_count` include and NO
+    // `select` — the secured findMany does not honour `select`, so forcing one
+    // would be an ignored no-op. Access-scoping lives entirely in `_count.where`.
+    expect(findMany).toHaveBeenCalledTimes(1)
+    const callArg = findMany.mock.calls[0][0]
+    expect(callArg).not.toHaveProperty('select')
+    expect(callArg).toEqual({
+      include: { _count: { select: { posts: { where: { status: { equals: 'published' } } } } } },
+    })
+  })
+
+  it('preserves a sibling condition co-present in the same member as the marker', async () => {
+    const config = makeConfig()
+    const findMany = vi.fn(async () => [{ id: 'u2', _count: { posts: 7 } }])
+    // A single AND-member carrying BOTH a scalar condition and the count marker.
+    // The marker resolution must keep the co-present sibling, not replace the
+    // member wholesale (guards against a future filter-engine change that merges
+    // conditions into one member).
+    const where = {
+      AND: [
+        {
+          name: { contains: 'ada' },
+          posts: { [RELATIONSHIP_COUNT_FILTER_KEY]: { operator: 'gt', value: 5 } },
+        },
+      ],
+    }
+    const resolved = await resolveRelationshipCountFilters(
+      where,
+      config.lists.User,
+      'User',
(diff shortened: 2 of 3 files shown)


borisno2 (commented):
## Code review — PR #771 (relationship count-filter hardening)

**Verdict: APPROVE** (posted as a Comment review because the reviewing identity is also the PR author, so GitHub blocks a formal Approve.)

Reviewed strictly from the PR diff plus the code it touches. Both hardening items are correct, well-scoped, and the tests meaningfully exercise the new behaviour. No blocking issues found.

### 1. "Leave projection as-is" — claim verified ✅
The PR claims the secured `context.db.findMany` does **not** honour `select`, so adding one would be a no-op. Confirmed against source:
- `packages/core/src/context/index.ts` — `warnIfSelectIgnored` (L107–122) is invoked by both `findUnique` (L904) and `findMany` (L1042); the `select` arg is typed with the comment *"not honoured — accepted only so the no-op can be made visible"* and is never forwarded to Prisma.
- `packages/core/CLAUDE.md` "Narrowing Reads (`select` is not honoured)" corroborates.

So omitting `select` on the count read (`relationship-count.ts` L194–196) is correct — **not** a missed optimization — and the explanatory comment (L194–205 region) is accurate. The residual scalar over-fetch is a genuine, documented limitation of the read pipeline, correctly deferred as out of scope.

### 2. Marker-member robustness — verified ✅
`mergeResolvedMember` (L228–235) + the sibling-spread at the call site (L280–298):
- **One-condition-per-member (today's invariant):** `siblings` is empty → helper returns `resolved` unchanged. Byte-identical to the previous wholesale `resolvedMembers.push(resolved)`. ✅
- **Co-present non-`id` sibling:** preserved via `{ ...siblings, ...resolved }` instead of being dropped. Covered by the "preserves a sibling condition" test. ✅
- **Sibling shares `id`:** collision detected (`Object.keys(resolved).some(k => k in siblings)`) and both are ANDed rather than one clobbering the other. Covered by the "ANDs the resolved id constraint" test. ✅

### 3. Access-scoping preserved — verified ✅
`resolveOneCountFilter` still reads through the secured `args.context.db[getDbKey(listKey)]` (L190) — no raw Prisma introduced. The `_count.where` still comes from the related list's `query` access via `relationshipCountAccessEntry` (L193, L59–78). Denied→0 (L186–188) and zero-count / empty-`in` paths are untouched. The hardening did not weaken security scoping.

### 4. Conventions / tests / changeset — verified ✅
- No new `any` and no exposed casts. The two `eslint-disable ... no-explicit-any` are pre-existing on `ListConfig<any>`; new code uses `Record<string, unknown>` confined internally, consistent with the repo's typing rules.
- Changeset `.changeset/quiet-otters-hum.md` present, `@opensaas/stack-core: patch`, correctly scoped as a bug-fix/hardening (not a feature). Description matches the change.
- Tests are real assertions on resolved output, not placeholders; the "no `select` projection" test also pins the exact `include` shape, guarding the access-scoping.

### Minor / non-blocking observations
- **Multiple markers in one member would leak.** If the one-condition-per-member invariant were ever violated by *two markers* co-present in a single member, `findMarker` resolves only the first and the second would now survive as an unresolved marker sibling passed to Prisma. This is strictly outside the documented invariant (and the old wholesale-replace would have silently dropped it instead) — so not a regression and not reachable today, but worth a line in the helper's doc if the invariant is ever relaxed. No action required.
- Comment quality is excellent — the rationale is captured where a future reader will look.

Nice, disciplined hardening PR. Approving.

---

PR #772, opened 2026-07-20
test(auth): lock generated Session/Account user FK shape (parity, not a regression)

borisno2:
Implements #753 · Part of #728

## Investigation outcome: intended behaviour, not a regression

#753 asked whether the generator's auth-schema output — which (a) adds `onDelete: Cascade` on `Session.user`/`Account.user` and (b) drops `@@index([userId])` on those models — is correct.

**Both are deliberate and pre-existing**, introduced in #679 → PR #698 (`cf8b4bd`) to mirror better-auth's own generated schema. They are declared in `userRelationshipDb` (`packages/auth/src/config/derive-auth-lists.ts`):

- `isIndexed: false` on the `Session.user` / `Account.user` relationships suppresses the default FK `@@index`.
- A field-level `extendPrismaSchema` rewrites the relation line to add `onDelete: Cascade`.

The goal is **clean-diff adoption**: better-auth's schema emits no separate `userId` index and cascades user deletes, so matching it lets a project adopt a live better-auth database without a spurious `DROP/CREATE INDEX` migration or a referential-action change.

### On the "silent performance regression" concern
The missing `@@index([userId])` is a deliberate parity tradeoff, not an accident. Session lookups run through the unique `token` index and verification through `identifier`; the `userId` column is unindexed by default, exactly as better-auth ships. A project that wants the index can override `isIndexed` on the relationship. The generic FK-index path is untouched — ordinary app relationships (e.g. `Post.author`) still emit `@@index` as before (locked by the contrast test here).

### Reproduction
`pnpm generate` in `examples/starter-auth` on current `main` produces **zero diff** against the committed schema. The drift reported in #753 was a previously-stale committed file, already fixed when #751 merged the fresh regeneration.

## What this PR does (capture the decision, no code fix)
Since the output is correct, no generator change is warranted. Instead:

- **ADR-0007**: added a decision bullet documenting the auth FK-shape parity (no `@@index`, `onDelete: Cascade`), the performance tradeoff, and the `isIndexed` escape hatch.
- **New test** `packages/auth/tests/generated-fk-shape.test.ts`: end-to-end run of the auth config through the real generator, asserting Session/Account emit `onDelete: Cascade` with **no** `@@index([userId])`, plus a contrast case proving a non-auth relationship FK still gets the default `@@index`. This closes the drift-detection gap (existing tests covered the config inputs and the generic mechanism separately, but nothing asserted the assembled auth schema text).
- Adds `@opensaas/stack-cli` as an auth devDependency so the generator can run in-test.

## Testing
- `packages/auth` tests: 9 files, 140 passed (incl. 2 new).
- `pnpm lint`: 0 errors (pre-existing warnings only).
- `pnpm manypkg check`: success.
- `examples/starter-auth` regenerate: no diff.

Recommendation: with the decision now documented + locked, #753 can be closed as intended-behaviour.

---


packages/auth/tests/generated-fk-shape.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      current = await plugin.beforeGenerate(current)
+    }
+  }
+  return generatePrismaSchema(current)
+}
+
+/** Slice out a single `model X { ... }` block from generated schema text. */
+function modelBlock(schema: string, modelName: string): string {
+  const start = schema.indexOf(`model ${modelName} {`)
+  if (start === -1) throw new Error(`model ${modelName} not found in generated schema`)
```
borisno2, line 31: Nit (non-blocking): `generateSchema` re-implements the CLI generate sequence (`config()` + a `beforeGenerate` loop) by hand rather than calling the shared pipeline. For the FK-shape assertion this is faithful — the field-level `extendPrismaSchema` that adds `onDelete: Cascade` and the `isIndexed` suppression are both applied *inside* `generatePrismaSchema`, not in a later stage — so the lock is sound today. The caveat is that if the real pipeline ever grows a step that mutates schema text after `beforeGenerate` (e.g. an `afterGenerate` transform), this helper would silently diverge from what the CLI emits. A short comment pointing at the canonical pipeline function (or reusing it) would harden this against future drift. Not required for merge.

---

packages/auth/tests/generated-fk-shape.test.ts
```diff
@@ -0,0 +1,81 @@
+import { describe, it, expect } from 'vitest'
+import { config, list } from '@opensaas/stack-core'
+import { text, relationship } from '@opensaas/stack-core/fields'
+import type { OpenSaasConfig } from '@opensaas/stack-core'
+import type { Plugin } from '@opensaas/stack-core/extend'
+import { generatePrismaSchema } from '@opensaas/stack-cli/generator/prisma'
+import { authPlugin } from '../src/config/plugin.js'
```
borisno2, line 7: Nit (non-blocking): this imports `@opensaas/stack-cli/generator/prisma`, whose package export resolves to `dist/` (not source). So this test requires `@opensaas/stack-cli` to be built before `packages/auth`'s vitest run — running the auth suite in isolation on a clean tree without a prior `pnpm build` of cli would fail to resolve the import. Fine under CI (which builds first) and the PR reports the suite green, but worth being aware of as an ordering coupling. The auth plugin itself is imported from source (`../src/config/plugin.js`), so the mismatch is only for the cli dep.

---

borisno2 (commented):
## Review — PR #772 (test: lock generated Session/Account user FK shape)

**Verdict: APPROVE** (posted as a Comment review; verdict stated explicitly here). This is a well-scoped, docs + test + devDep change. I verified the investigation's central conclusion against the actual code and it holds up.

### What this PR does
Investigation outcome for #753: the auth generator's `onDelete: Cascade` and missing `@@index([userId])` on `Session.user`/`Account.user` are **intended better-auth parity**, not a regression. No generator change; instead it (a) documents the decision in ADR-0007, (b) adds an end-to-end regression test over the real generator, and (c) adds `@opensaas/stack-cli` as an auth devDep so the test can run the generator.

### Verification (I traced the claims, did not take them on faith)

**1. "Intended, not a regression" — CONFIRMED.** In `packages/auth/src/config/derive-auth-lists.ts`:
- `Session.user` (L195) and `Account.user` (L223) both set `isIndexed: false` on the relationship — an explicit, deliberate opt-out of the default FK index.
- `userRelationshipDb` (L118-127) attaches a field-level `extendPrismaSchema` that rewrites the relation line to inject `onDelete: Cascade`.
Both are declared on purpose with an explanatory doc comment (L109-116) citing #679. Not incidental.

**2. Generic FK-index path is untouched — CONFIRMED.** `packages/cli/src/generator/prisma.ts` L330-336 emits `@@index([fk])` per-model, driven by each field's own `foreignKeyIndex.indexType`. Only the two auth `user` relations opt out via `isIndexed: false`; there is no global suppression. Ordinary app relationships (e.g. `Post.author`, and the test's `Widget.owner`) still get `@@index`.

**3. The test genuinely locks the behaviour — CONFIRMED.** `packages/auth/tests/generated-fk-shape.test.ts` runs the real `generatePrismaSchema` (from `@opensaas/stack-cli/generator/prisma`) over an actual `authPlugin` config and asserts on assembled schema text:
- positive: `Session`/`Account` blocks contain `@relation(onDelete: Cascade, fields: [userId], references: [id])` and do **not** contain `@@index([userId])`;
- contrast: a non-auth `Widget.owner` FK still emits `@@index([ownerId])`.
These are real assertions, not placeholders. The exact `@relation(onDelete: ..., fields: [...], references: [id])` output format matches what the field-level `extendPrismaSchema` mechanism already produces (existing `prisma.test.ts` covers the identical shape with `SetNull`), so the assertion strings are faithful to generator output. `extendPrismaSchema` is applied inside `generatePrismaSchema`, so the manual pipeline in the test captures it correctly.

**4. ADR-0007 edit — accurate, no overclaiming.** The added bullet correctly frames it as parity (clean diff vs a live better-auth DB), states the perf tradeoff honestly (unindexed `userId`, relying on `token`/`identifier` indexes), and documents the `isIndexed` escape hatch. Matches the code.

**5. Dependency + lockfile — benign.** `@opensaas/stack-cli: workspace:*` is a **devDependency** only (not runtime), shared-package version aligned via workspace protocol, and there is **no cli→auth cycle** (cli has no auth dependency). The lockfile churn beyond the one added `link:../cli` entry is pnpm re-normalizing `eslint-import-resolver-typescript` peer-dependency path strings — cosmetic, no version changes, matches the author's note.

**6. No shipped-code behaviour change — CONFIRMED.** Diff touches only: changeset, ADR doc, `packages/auth/package.json` (devDep), the new test, and `pnpm-lock.yaml`. No runtime `src` changes. No `any`/type-casts introduced in the test. Changeset is correctly scoped: `@opensaas/stack-auth: patch`.

### Findings by severity
- **Blocking:** none.
- **Minor / nits (inline):** two non-blocking notes on the test — (a) it hand-rolls the CLI generate sequence rather than reusing a shared pipeline, so it could drift if a future post-`beforeGenerate` schema transform is added; (b) the `@opensaas/stack-cli` import resolves to `dist/`, coupling the auth suite to cli being built first. Neither affects correctness today.

### Bottom line
The "intended behaviour, not a regression" conclusion is genuinely correct and well-supported by the code, and the new test really does lock both the auth FK shape and the contrast that ordinary FKs keep their index. Recommend merge; #753 can be closed as intended-behaviour once merged.

_Reviewed by Claude Code (`/review`). Posted as a Comment review; treat as an approval._

---

Issue #758, opened 2026-07-20
Test hardening: lock the server-side back-reference overwrite in createRelated (relationship create drawer)


borisno2:
## Context

Non-blocking follow-up from the review of #757 (pre-linked create drawer, #738). The code is correct; this adds test coverage for a security-critical invariant that is currently guaranteed by code but not locked by a test.

## The gap

`createRelated` (`packages/core/src/context/index.ts`) sets the back-reference on the server by spreading client `data` and THEN overwriting `data[field] = { connect: { id: parentId } }`, so a client-supplied `data[field]` can never re-target the link to a different parent/relationship. The current happy-path test sends `data: { title, content }` with no back-reference field — so it proves the server *sets* the link, but NOT that it *overwrites* a hostile client-supplied one.

## What to add

- [ ] A core test that calls `createRelated` with `data` containing a hostile back-reference (e.g. `data: { title, author: 'evil-id' }`) and asserts the underlying secured `create` is called with the back-reference connecting to the trusted `parentId` (`author: { connect: { id: parentId } }`), NOT `'evil-id'`. Cover both to-one and to-many back-references.

## Also worth considering (minor, from the same review)

- [ ] Defensive guard in `createRelated`: validate that `field` is actually a relationship field, and handle a lone `field`/`parentId` (only one supplied) explicitly. Unreachable from the UI today (the drawer only mounts with a valid back-reference and always sends both), and a malformed direct call already degrades to a normal access-governed create + generic error — so this is hardening, not a fix.
- [ ] (Perf, optional) `resolveCreateForm` eagerly runs `prepareItemForm` (bounded relationship-option fetches) on every parent item-view render when create is allowed, even if the drawer is never opened — consider deferring to drawer-open if it shows up in profiling.

Non-blocking; #757 is merged. This is test/robustness hardening for the create-and-link path.

PR #773, opened 2026-07-20
test(core): lock the createRelated back-reference overwrite (#758)

borisno2:
Implements #758. Part of #728.

## Summary

`createRelated` (`packages/core/src/context/index.ts`) sets the parent back-reference on the **server** by spreading the client `data` and **then** overwriting `data[field] = { connect: { id: parentId } }`, so a hostile client-supplied `data[field]` can never re-target the link. This invariant was only implicitly guaranteed by code — this PR locks it with tests and adds a small defensive guard.

### Primary — tests (the load-bearing deliverable)

Added to `packages/core/tests/context.test.ts`:

- **To-one back-ref:** calls `createRelated` with hostile `data: { title, author: 'evil-id' }` and asserts the secured `create` receives `author: { connect: { id: 'u1' } }` (the trusted `parentId`), not `'evil-id'`.
- **To-many / M2M back-ref:** calls `createRelated` with hostile `data: { title, teachers: { connect: [{ id: 'evil-id' }] } }` and asserts the secured `create` receives `teachers: { connect: [{ id: 't1' }] }` (trusted parent by id).

Both run through the **secured** delegate (`db[dbKey]`), never raw prisma.

### Secondary — defensive guard (hardening, not a fix)

Minimal guard in `createRelated`, unreachable from the drawer (which always sends both `field` and `parentId` with a valid relationship back-reference), so happy-path behaviour is unchanged:

- Reject a malformed direct call that supplies only one of `field`/`parentId` (rather than silently skipping injection and letting `data[field]` slip through).
- Validate that `field` names a relationship field before injecting the connect.

Each guarded case has a test. `backRefField.type` is a plain `string` on `BaseFieldConfig`, so no casts were needed.

### Deferred

The perf note (`resolveCreateForm` eager `prepareItemForm`) is explicitly deferred per the issue — not touched here.

## Verification

- `packages/core` vitest: **875 passed (42 files)**, including the 4 new tests.
- `pnpm lint`: 0 errors (only pre-existing warnings).
- Core `tsc` build: clean.
- `pnpm manypkg fix` + `pnpm format`: clean.
- Changeset: `patch` for `@opensaas/stack-core` (shipped guard code changed).

---

### DIFF
--- a/.changeset/quick-owls-guard.md
+++ b/.changeset/quick-owls-guard.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Harden `createRelated` server action: reject malformed calls that supply only one of `field`/`parentId`, and validate that the back-reference names a relationship field before injecting the parent connect.
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -620,11 +620,35 @@ export function getContext<
     // as `{ created: false }` with a generic reason (no denied-vs-absent leak).
     if (props.action === 'createRelated') {
       try {
+        // Defensive guard (hardening; unreachable from the drawer, which always
+        // mounts on a valid back-reference and sends BOTH field and parentId).
+        // A lone field/parentId is a malformed direct call — reject it rather
+        // than silently skip back-reference injection and let a client-supplied
+        // data[field] slip through unguarded to the create.
+        if (!!props.field !== !!props.parentId) {
+          return {
+            created: false,
+            error: 'createRelated requires both field and parentId, or neither',
+          }
+        }
         const data: Record<string, unknown> = { ...props.data }
         if (props.field && props.parentId) {
           const backRefField = listConfig.fields[props.field]
-          const backRefIsMany =
-            !!backRefField && 'many' in backRefField && backRefField.many === true
+          // The back-reference must name a relationship field on this list; a
+          // non-relationship field would otherwise receive a nonsensical
+          // { connect } value. Also hardening — the drawer only ever passes a
+          // real relationship back-reference here.
+          if (!backRefField || backRefField.type !== 'relationship') {
+            return {
+              created: false,
+              error: `Field "${props.field}" on list "${props.listKey}" is not a relationship field`,
+            }
+          }
+          const backRefIsMany = 'many' in backRefField && backRefField.many === true
+          // The back-reference is set on the SERVER from the trusted parentId,
+          // OVERWRITING any client-supplied data[field] spread in above, so a
+          // hostile client can never re-target the link (a to-many back-ref
+          // connects the parent by id).
           data[props.field] = backRefIsMany
             ? { connect: [{ id: props.parentId }] }
             : { connect: { id: props.parentId } }
--- a/packages/core/tests/context.test.ts
+++ b/packages/core/tests/context.test.ts
@@ -671,6 +671,137 @@ describe('getContext', () => {
         expect(createArg.data.teachers).toEqual({ connect: [{ id: 't1' }] })
       })
 
+      // Security-critical invariant (#758): the server sets the back-reference by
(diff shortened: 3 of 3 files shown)


borisno2 (commented):
## Review — PR #773 `test(core): lock the createRelated back-reference overwrite (#758)`

**Verdict: APPROVE** (posted as a Comment review — submitting a formal Approve is avoided since the PR originates from the Claude Code integration branch.)

Reviewed strictly from the PR diff + the surrounding `packages/core/src/context/index.ts` and `tests/context.test.ts`. This is a focused, well-executed hardening PR. The security invariant is genuinely locked, the guard is correct and non-regressing, and the changeset is scoped right. No blocking issues.

### What it does
Adds security-property tests that prove `createRelated` overwrites a hostile client-supplied back-reference with the trusted `parentId`, plus a small defensive guard that rejects malformed direct calls (only one of `field`/`parentId`, or a `field` that isn't a relationship) before any create is attempted.

### Verification against the review criteria

**1. Tests lock the invariant — CONFIRMED**
- To-one: calls `createRelated` with hostile `data: { title, author: 'evil-id' }`, asserts the underlying `create` receives `author: { connect: { id: 'u1' } }` (trusted `parentId`) and `.not.toBe('evil-id')`. Real assertions on `mockPrisma.post.create.mock.calls[0][0]`.
- To-many / M2M: hostile `data: { teachers: { connect: [{ id: 'evil-id' }] } }`, asserts `create` receives `teachers: { connect: [{ id: 't1' }] }`. The full field value is overwritten, so the hostile connect is discarded — asserted on the actual create args.
- Both go through the secured delegate: the handler uses `model = db[dbKey]` (context/index.ts:505), never raw prisma. The create is `model.create({ data })` (line 632), so the mock captured at the prisma boundary reflects the value emitted *after* the secured write pipeline ran.

**2. Defensive guard correct and safe — CONFIRMED**
- XOR reject: `if (!!props.field !== !!props.parentId)` returns `{ created: false, error }` and never reaches the create. Test confirms `create` not called.
- Relationship-field validation: `if (!backRefField || backRefField.type !== 'relationship')` rejects before injecting the connect. Test with scalar `field: 'title'` confirms rejection + no create.
- Happy path unchanged: the drawer always sends both `field` and `parentId` with a valid relationship back-ref, so both guards are inert on that path; the `neither` case (unlinked plain create) is also preserved. No behavioural change.
- `field.type` access is genuinely cast-free: `backRefField = listConfig.fields[props.field]` is a `FieldConfig`, and `.type` is the discriminant on `BaseFieldConfig`. Notably the adjacent `relationshipOptions` handler *does* cast (`as { type?: string; ref?: string }`), and the new code deliberately does not — good.
- No new denied-vs-absent leak: the guard errors describe a malformed/misconfigured call (schema-shaped), not record existence. The "not a relationship field" message matches the wording already used by `relationshipOptions`, so it's consistent with convention.

**3. No `any`/exposed casts; changeset scoped — CONFIRMED**
- Shipped guard code introduces no `any` and no casts.
- The `any` in the new M2M test is a mock, gated by an `eslint-disable`, matching the pre-existing `connects a to-many back-reference…` test in the same file — established test-only convention.
- Changeset `@opensaas/stack-core: patch` is correct: shipped guard code changed, and a defensive hardening is a patch.

### Minor / non-blocking
- The new M2M hostile test re-declares the `m2mPrisma`/`m2mConfig` fixtures inline, duplicating the pre-existing happy-path M2M test (context.test.ts:610-672). A shared helper would cut ~50 lines of duplication, but this mirrors the file's current style and isn't worth blocking on.
- The guard's `!!props.field !== !!props.parentId` XOR is terse but correct and well-commented.

Solid work — the invariant is now proven rather than implicit, and the guard closes the one path where a lone `field`/`parentId` could have slipped an unguarded `data[field]` through.

---

PR #774, opened 2026-07-20
feat(ui): bound item-view relationship tables with take + "showing N of M" footer

borisno2:
## Summary

Implements #752. Part of #728.

The read-only item-view Relationship tables (#734) previously fetched every related row with **no `take`/limit** — a record with many related rows loaded and rendered every one on each edit-page open. This bounds the fetch and surfaces the full access-scoped total in the footer.

### What changed

- **Bounded fetch:** each to-many Relationship table is now fetched with a `take` — default `DEFAULT_ITEM_VIEW_TAKE` (10), overridable per relationship via `ui.itemView.take`. Rows remain access-scoped through the secured context (`context.db` include), never raw prisma.
- **"Showing N of M" footer:** the totals footer now reads `Showing N of M rows`, where **N** is the rendered (bounded) count and **M** is the full **access-scoped** total. M comes from a filtered `_count` fetched in the *same* secured `findUnique`, folding each related list's own `query` access into the count (reusing #732's `buildRelationshipCountSelect`). A fully-denied related list is omitted from the count → reads `Showing 0 of 0` and **never leaks a true total**. The row count is always shown, including the zero-column footer path (#734's guarantee kept).
- **Core:** `mergeIncludeWithAccessControl` now preserves a caller-supplied `take` on a to-many relation include. It only ever *narrows* the fetched rows and can never widen past the access `where`, so it rides on top of the per-relation access filter — this is what lets the secured `findUnique` bound related-row reads without dropping row-level access.
- **Example (secondary note):** curated starter-auth's auth-derived `Account`/`Session` Relationship-table columns (`ui.itemView.columns`) so credential/token columns (`accessToken`/`refreshToken`/`idToken`, and Session's `token`) are **not** surfaced by default on the User edit page. Field-level access already governs the values; this keeps the example from *modelling* token-column exposure.

### Access-scoping of M (reviewer note)

M is fetched only through the secured context via `buildRelationshipCountSelect`, which folds the related list's `query` access into the `_count` `where` and omits fully-denied lists. `readAccessScopedTotal` falls back to the bounded row count (0 for a denied list) when a field is absent from `_count`, so a denied related list can never expose a real total. `checkAccess(undefined)` denies by default (ADR-0013), so closed auth lists (Account/Session in the example) correctly report `0`.

### Scope

Item-view relationship-table bounding + N-of-M footer + example column curation only. No list-view or edit/create/remove-affordance changes beyond adding the footer/take.

## Tests

- **Unit (core):** `mergeIncludeWithAccessControl` preserves a caller `take` and AND-combines it with the access `where`; drops it for a denied relation. Full core suite: **877 passed**.
- **Unit (ui):** `deriveItemView` take default + override + malformed-fallback; `buildItemViewInclude` applies take (default + override) alongside nested includes; `readAccessScopedTotal` (present / zero / denied→fallback / no-leak); footer "Showing N of M" incl. bounded, fallback, zero-count, and zero-column paths. Full ui suite: **520 passed**.
- **e2e (`e2e/starter-auth`):** full suite **68 passed** — the User edit page shows `Showing 2 of 2 rows` (access-scoped), the denied Session table shows `Showing 0 of 0 rows`, and the Account/Session tables no longer surface token columns.
- `pnpm lint` (0 errors), `pnpm format` (clean), `pnpm build` (all packages typecheck) — clean.

## Changeset

`minor` for `@opensaas/stack-core` and `@opensaas/stack-ui`.

---

### DIFF
--- a/.changeset/spicy-rivers-bound.md
+++ b/.changeset/spicy-rivers-bound.md
@@ -1,0 +1,34 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': minor
+---
+
+Bound the admin item-view Relationship tables with a `take` and a "showing N of M" footer
+
+The read-only Relationship tables on a record's edit page (issue #734) previously
+fetched every related row unbounded. They now fetch a bounded page of related rows
+and surface the full access-scoped total in the footer.
+
+- **Bounded fetch:** each to-many Relationship table fetches at most a default cap
+  of related rows (`DEFAULT_ITEM_VIEW_TAKE`, 10), overridable per relationship via
+  `ui.itemView.take`. Rows are still fetched through the secured context, so only
+  access-visible rows come back.
+- **"Showing N of M" footer:** the totals footer now reads `Showing N of M rows`,
+  where N is the rendered (bounded) count and M is the full access-scoped total,
+  fetched via a filtered `_count` that folds the related list's own `query` access
+  in (mirroring the list view's count columns). A fully-denied related list reads
+  `Showing 0 of 0` and never leaks a true total. The row count is always shown,
+  including the zero-column footer path.
+
+```typescript
+sessions: relationship({
+  ref: 'Session.user',
+  many: true,
+  // Cap this table at 5 rows; the footer still shows the full access-scoped total.
+  ui: { itemView: { take: 5 } },
+})
+```
+
+Core: `mergeIncludeWithAccessControl` now preserves a caller-supplied `take` on a
+to-many relation include (it only narrows the fetch, never widening past the access
+`where`), so the secured `findUnique`/`findMany` include can bound related-row reads.
--- a/e2e/starter-auth/05-item-view.spec.ts
+++ b/e2e/starter-auth/05-item-view.spec.ts
@@ -72,9 +72,11 @@ test.describe('Item view layout', () => {
     await expect(table.getByText('View Post A')).toBeVisible()
     await expect(table.getByText('View Post B')).toBeVisible()
 
-    // Footer: row count always, plus the summed numeric column (5 + 10 = 15).
+    // Footer: "showing N of M" (issue #752) with M the access-scoped total —
+    // here the fetch is under the default cap, so N === M === 2 — plus the
+    // summed numeric column (5 + 10 = 15).
     const footer = table.locator('[data-slot="relationship-table-footer"]')
-    await expect(footer).toContainText('2 rows')
+    await expect(footer).toContainText('Showing 2 of 2 rows')
     await expect(footer).toContainText('15')
 
     // Rows show only access-visible data: the closed Session list (ADR-0013,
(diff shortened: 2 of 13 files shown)


borisno2 (commented):
## Review — verdict: APPROVE

(Submitted as a Comment review; stating the verdict explicitly since a formal Approve can be blocked when the integration identity is involved.)

Thorough read of the diff plus the load-bearing surrounding code (`access-filter.ts` full merge, `relationship-count.ts` `buildRelationshipCountSelect`, and the `context/index.ts` read pipeline). This is a careful, security-conscious change. The two properties that most deserved scrutiny — the access-scoping of M and the take-vs-access-filter interaction — both hold.

### Access-scoping of M (security-critical) — sound
- M is fetched only through the secured `findUnique` via `_count: { select: buildRelationshipCountSelect(...) }`. `buildRelationshipCountSelect` folds each related list's operation-level `query` access into the per-relation `where` and **omits a fully-denied list entirely** (`relationshipCountAccessEntry` returns `denied` → `continue`).
- `readAccessScopedTotal` reads the value off `_count` only when present-and-numeric; a denied (omitted) list has no key → falls back to `rows.length`, which is itself `0` for a denied list because `buildIncludeWithAccessControl` also drops that relation from the include (`accessResult === false → continue`). Net result "Showing 0 of 0" — no true-total leak. Confirmed by the e2e assertion on the denied Session table.
- Crucially, the include `where` and the count `where` are derived from the **same** `checkAccess(queryAccess, ...)` call in both `buildIncludeWithAccessControl` and `relationshipCountAccessEntry`, so M and N can never diverge in scope (M ≥ N always holds, and both share the exact same access filter). Zero-count is handled.
- `_count` is a non-declared-relationship key, so it passes through `mergeIncludeWithAccessControl` unchanged (the `!isDeclaredRelationship` branch) and survives `stripVirtualFieldsFromInclude` — it is not re-scoped or dropped. Correct, since it's already access-scoped at construction.

### take vs. access filter (`mergeIncludeWithAccessControl`) — only narrows, never widens
- `asEntryObject` now carries a numeric `take`; the merge re-attaches `callerEntry.take` **on top of** the merged `entry` that already contains the access `where` (AND-combined) and the recursively-merged nested include. It does not replace or bypass the access filter. Prisma applies `where` then `take`, so the bound rides strictly on top of the per-relation access scope.
- A denied relation is still dropped wholesale (`accessValue === undefined → continue`) — the `take` cannot resurrect it. Both behaviours are locked by the new core unit tests (preserve+AND-combine; drop-on-denied). The read pipeline routes the item-view caller include through this merge on both `findUnique` and `findMany` (context/index.ts lines 990/1124), confirming the path is live.

### Bounded fetch correctness — sound
- `DEFAULT_ITEM_VIEW_TAKE = 10`, per-relationship override via `ui.itemView.take`. `readPositiveInteger` guards with `Number.isInteger(value) && value > 0`, falling back to the default for `0` / negative / fractional / `NaN` / non-numeric — so no `take: 0`/negative/NaN can reach Prisma. Override plumbs through `deriveItemView` → `RelationshipTableSection.take` → `buildItemViewInclude` entry, covered by unit tests.

### "Row count always shown" (#734 guarantee) — preserved
- `formatCountLabel` falls back to `count` when `total` is undefined, and the zero-column footer branch also renders `countLabel`. The updated `RelationshipTableClient` tests cover the zero-column path ("Showing 3 of 12 rows") and the no-total fallback.

### Example column curation — example-only, no schema change
- `examples/starter-auth` re-declares `accounts`/`sessions` with the same `ref`/`many`, changing only `ui.itemView.columns` + `removeAction`. Generated Prisma schema is unaffected (schema is driven by `ref`/`many`, not `ui`), and field-level access still governs the values — the curation only stops token columns being *modelled* as default headers. e2e asserts the token headers are absent. Good.

### Quality
- No exposed `any`/casts in production code (the `any` uses are eslint-disabled in tests only); `total` client prop is a serialisable number and minimal; footer uses existing `text-muted-foreground` token. Changeset correctly scoped `minor` for `@opensaas/stack-core` + `@opensaas/stack-ui`.
- Test trade-off is acceptable: the N<M truncation is unit-tested (`readAccessScopedTotal` present/zero/denied-fallback/no-leak + footer "Showing 2 of 42"), while the **security** property (denied → "0 of 0") is covered in e2e. Not a coverage gap for the access-scoping.

### Nits (non-blocking, no change required)
- `formatCountLabel` pluralises on M, so a `0`-of-`1` case would read "Showing 0 of 1 row". Harmless; M is the meaningful quantity.
- `buildItemViewInclude`/`readAccessScopedTotal` are now exported from `ItemForm.tsx` purely for unit tests — consistent with existing patterns in this codebase, fine to keep.

Verdict: **APPROVE** — the access-scoping of M and the take-on-top-of-access-filter interaction are both correct and well-tested.

---

Issue #761, opened 2026-07-20
Minor hardening: relationship-table inline-edit json no-op + custom-bulk-action error fallback


borisno2:
Non-blocking nits surfaced during review of #737 (PR #760) and #736 (PR #759). Both features are correct and merged/merging; these are small robustness/tidy-ups.

## 1. Inline-edit no-op detection is reference equality (from #737 review)

`RelationshipTableCell.commit()` skips the server round-trip only when `draft === displayValue`. For scalar fields (text/select/number) that's correct, but `json` columns are **not** excluded from `editableColumns`, and a json draft is a fresh object reference — so an unchanged `json` cell always round-trips to the secured `updateRelated` on Enter/blur. Harmless (it rewrites the same value) but a wasted request.

- [ ] Either exclude `json` from inline-editable columns for now, or use a normalised value-equality compare for the no-op check.

## 2. Custom-bulk-action handler error fallback can surface an internal message (from #736 review)

In the `bulkAction` branch of `packages/core/src/context/index.ts`, an unexpected error thrown inside a custom handler (not `ValidationError`/`DatabaseError`) is passed to `parsePrismaError` and its `.message` returned to the client. For a genuine non-Prisma bug this could surface an internal message to the client. Handler authors own their handlers, so it's minor.

- [ ] Add a generic fallback message for the non-Prisma, non-known-error case (return a generic "Action failed" to the client and log the real error server-side).

## Notes
Both are low priority and non-blocking for #728. No security impact (the access-control paths in both features were verified correct in review).

PR #775, opened 2026-07-20
fix(core,ui): harden relationship-table inline-edit and custom bulk actions (#761)

borisno2:
Implements #761. Part of #728.

Two small, independent hardening items surfaced during review of #737 and #736. Both are additive and low-risk.

## Item 1 — inline-edit json no-op round-trip (`@opensaas/stack-ui`)

`RelationshipTableCell.commit()` skips the server round-trip only when `draft === displayValue` (reference equality). That is correct for scalar fields, but `json` columns were **not** excluded from `editableColumns`, and a json draft is always a fresh object reference — so an unchanged json cell always round-tripped to the secured `updateRelated` on Enter/blur (harmless, but a wasted request).

**Chosen approach (the issue's preferred option): exclude `json` from inline-editable columns.** In `resolveEditableColumns` (`packages/ui/src/components/RelationshipTable.tsx`), `json` now joins the existing excluded types (relationship / virtual / password / system columns). This is cleaner than a normalised value-equality compare, matches the reference-equality assumption the cell's no-op guard already relies on, and json fields need a proper structured editor anyway (out of scope here). No change to `RelationshipTableCell` itself.

- Test: `packages/ui/tests/components/RelationshipTableEditableColumns.test.ts` — a json column is not editable; existing scalar/relationship/virtual/password/system exclusions stay intact.

## Item 2 — custom-bulk-action error fallback (`@opensaas/stack-core`)

In the `bulkAction` branch of `packages/core/src/context/index.ts`, an unexpected error thrown inside a custom handler (not a `ValidationError`/`DatabaseError`) was passed to `parsePrismaError`, whose `.message` was returned to the client — potentially surfacing an internal message.

Now: recognised Prisma errors (a `DatabaseError` from `parsePrismaError`) keep their user-safe, translated message; anything else is logged server-side via `console.error` and the client receives a generic `"Action failed"`. `ValidationError`/`DatabaseError` behaviour is unchanged. The handler still runs through the **secured** context — no change to access-scoping or silent-failure semantics.

- Test: `packages/core/tests/context.test.ts` — a handler throwing a plain `Error` returns `"Action failed"` (never the internal text) and logs the real error server-side.

## Verification (local)

- Unit: `@opensaas/stack-core` 878 passed; `@opensaas/stack-ui` 522 passed.
- e2e (throwaway config, non-default port 3123, local Chromium; not committed): `05-item-view`, `08-relationship-inline-edit`, `08-custom-bulk-actions` — 6/6 passed. No regression to inline-edit or custom bulk actions.
- `pnpm build` (typecheck) clean for both packages; `pnpm lint` (no new warnings/errors), `pnpm manypkg fix`, `pnpm format` clean.

## Changesets

- `@opensaas/stack-ui`: patch (json inline-edit exclusion)
- `@opensaas/stack-core`: patch (bulk-action generic error fallback)

---

### DIFF
--- a/.changeset/quiet-foxes-run.md
+++ b/.changeset/quiet-foxes-run.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Return a generic "Action failed" message (and log the real error server-side) when a custom bulk-action handler throws an unexpected non-Prisma error, instead of surfacing its internal message to the client
--- a/.changeset/tidy-owls-hide.md
+++ b/.changeset/tidy-owls-hide.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+Exclude json columns from relationship-table inline editing so an unchanged json cell no longer wastes a secured update round-trip
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -564,7 +564,15 @@ export function getContext<
           return { bulkAction: false, error: error.message }
         }
         const dbError = parsePrismaError(error, listConfig)
-        return { bulkAction: false, error: dbError.message }
+        // A recognised Prisma error carries a user-safe, translated message.
+        if (dbError instanceof DatabaseError) {
+          return { bulkAction: false, error: dbError.message }
+        }
+        // Anything else is an unexpected handler bug whose raw `.message` could
+        // leak internal detail to the client — log it server-side and return a
+        // generic client-facing message instead.
+        console.error(`Bulk action "${props.key}" on list "${props.listKey}" failed:`, error)
+        return { bulkAction: false, error: 'Action failed' }
       }
     }
 
--- a/packages/core/tests/context.test.ts
+++ b/packages/core/tests/context.test.ts
@@ -469,6 +469,41 @@ describe('getContext', () => {
         // The handler never ran.
         expect(mockPrisma.post.update).not.toHaveBeenCalled()
       })
+
+      it('returns a generic message (not the internal text) and logs when the handler throws a plain Error', async () => {
+        // An unexpected, non-Prisma, non-known-error bug inside the handler must
+        // not surface its internal message to the client (#761).
+        const consoleError = vi.spyOn(console, 'error').mockImplementation(() => {})
+        try {
+          const context = await getContext(
+            configWithPublishAction({
+              handler: async () => {
+                throw new Error('secret internal detail: connection string leaked')
+              },
+            }),
+            mockPrisma,
+            { userId: 'u1' },
+          )
+          const result = await context.serverAction({
+            listKey: 'Post',
+            action: 'bulkAction',
+            key: 'publish',
+            ids: ['p1', 'p2'],
+          })
+
+          // Client sees the generic fallback — never the internal message.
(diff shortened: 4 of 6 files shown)


borisno2 (commented):
## Review: PR #775 — harden relationship-table inline-edit and custom bulk actions

**Verdict: APPROVE** (posted as a Comment review — GitHub blocks a formal Approve because the review identity authored the PR).

Two small, independent, additive hardening items. Both are correct, well-scoped, and covered by meaningful tests. No blocking issues found.

### Item 1 — exclude `json` from inline-editable columns (`@opensaas/stack-ui`)
- `resolveEditableColumns` now `continue`s on `field.type === 'json'`, joining the existing relationship / password / virtual / system exclusions. Verified the exclusion is type-gated and does **not** touch other scalars (text/integer/select/checkbox/timestamp remain editable) — confirmed by the second test asserting `['title', 'count']` survive.
- The change is purely to *which* columns are editable; `RelationshipTableCell.commit()` and its `draft === displayValue` no-op guard are untouched. Correct rationale: a json draft is always a fresh reference so the reference-equality guard never holds, wasting a secured `updateRelated` round-trip on an unchanged object.
- Exporting `resolveEditableColumns` is a harmless test-seam widening, not a behaviour change.
- Test asserts json is not editable AND the other exclusions stay intact. Good.

### Item 2 — generic error fallback for custom bulk actions (`@opensaas/stack-core`)
- Verified the full error ladder in the `bulkAction` branch:
  - `ValidationError` / `DatabaseError` thrown directly → keep their user-safe `.message` (unchanged, no regression).
  - `parsePrismaError` returns a `DatabaseError` for recognised Prisma errors → user-safe translated message preserved (`dbError instanceof DatabaseError`).
  - Anything else — a plain `Error` (returned as-is by `parsePrismaError`) or an unknown value — is `console.error`-logged server-side and the client gets a generic `"Action failed"`, never the raw `.message`. This closes the leak.
- The handler still runs through the **secured** context (`action.handler({ ..., context })` with the same request context) and `hasAccess` is re-checked before invocation — access-scoping / silent-failure / no-leak semantics are unchanged.
- Test throws a plain `Error('secret internal detail...')`, asserts the result is exactly `{ bulkAction: false, error: 'Action failed' }`, that the message does not contain the internal text, and that `console.error` was called with an `Error` argument. Meaningful, not a placeholder.

### General
- No `any` and no exposed casts introduced. The only cast is `as unknown as AccessContext<unknown>` in the UI test's `makeContext` helper — internal test plumbing via `unknown`, which is acceptable per the repo's typing rules.
- Changesets correctly scoped: `@opensaas/stack-ui` patch (json exclusion) + `@opensaas/stack-core` patch (generic error fallback). Both are hardening/bug-fixes → patch is right.
- `console.error` matches the existing logging convention in this file.

### Non-blocking observation (pre-existing, out of scope)
- The "other Prisma errors" branch of `parsePrismaError` wraps `prismaError.message` in a `DatabaseError`, so an unrecognised Prisma error code can still surface Prisma's own message to the client via the preserved-`DatabaseError` path. This is unchanged behaviour, consistent with the create/update server-action path, and outside this PR's stated scope (non-Prisma handler bugs). Worth a future look if bulk-action handlers do raw DB work, but not a reason to hold this PR.

LGTM. 👍

---

Issue #767, opened 2026-07-20
storage-vercel: honour `public: false` with full private-access support


borisno2:
## What to build

Setting `public: false` on the Vercel Blob storage config actually produces private blobs, and the framework can still read and serve them.

Today the option is dead code — the provider always uploads with `access: 'public'` regardless of the config, so the documented "Whether files should be publicly accessible" option silently does nothing. `@vercel/blob` v2 supports `access: 'private'`, but private blobs are not fetchable via their plain URL, so honouring the option also means fixing the read path.

- Upload: `public: false` maps to `access: 'private'`; the default stays public.
- Download: fetch private blobs through the SDK's authorized read path instead of a plain `fetch(url)`; public blobs keep working as today.
- Implement the optional `getSignedUrl()` provider method so private files can be served through developer-controlled routes (the pattern the storage docs already describe for private S3 files).
- Docs: how-to and reference storage pages describe private mode and the serving pattern.

## Acceptance criteria

- [ ] Upload with `public: false` creates a blob with private access; default remains public
- [ ] `download()` returns content for both public and private blobs
- [ ] `getSignedUrl()` returns a time-limited URL that serves a private blob
- [ ] Metadata stored at upload time remains usable by the serving pattern for private blobs
- [ ] Tests cover upload/download/signed-URL for both access modes
- [ ] Docs updated for private mode and serving private files
- [ ] Minor changeset for `@opensaas/stack-storage-vercel`

## Blocked by

- #766 — rewrites the same `download()` path on top of the corrected not-found semantics and truthful test doubles.

PR #779, opened 2026-07-21
fix(storage-vercel): getUrl() now returns a resolvable Vercel Blob URL

### LINKED ISSUE #768: storage-vercel: getUrl() returns a URL that can never resolve
## What to build

`getUrl(filename)` on the Vercel Blob provider returns a URL that actually serves the blob.

Today it fabricates a host without the store subdomain, which always 404s — real blob URLs embed the store ID and access mode in the host (`https://<storeId>.<access>.blob.vercel-storage.com/<pathname>`). The store ID is derivable from the read-write token the same way the SDK derives it internally, so construct the URL from store ID + access mode + pathname. The token format is undocumented (parity-with-SDK, not a stable API), so a token that doesn't yield a store ID must fail loudly rather than produce a silently wrong URL.

## Acceptance criteria

- [ ] `getUrl()` output for a public blob matches the URL the SDK returns at upload time
- [ ] Private-mode stores produce the private-host form (servable via the signed/authorized paths from the private-access ticket)
- [ ] A token that doesn't yield a store ID produces a descriptive error, not a fabricated URL
- [ ] Tests cover both access modes and the malformed-token case
- [ ] Patch changeset for `@opensaas/stack-storage-vercel`

## Blocked by

- #767 — the URL host embeds the access mode introduced there.

borisno2:
## Summary
- `getUrl()` on the Vercel Blob provider fabricated a host without the store subdomain (`https://blob.vercel-storage.com/...`), which always 404'd.
- It now derives the store ID from the configured `token`/`storeId` (or the `BLOB_READ_WRITE_TOKEN`/`BLOB_STORE_ID` env vars), mirroring the same precedence and parsing `@vercel/blob` uses internally, and constructs `https://<storeId>.<access>.blob.vercel-storage.com/<pathname>` — matching the URL the SDK returns at upload time.
- Private-mode stores (`public: false`) now produce the private-host form, servable via the signed/authorized paths added in #767.
- A token that doesn't yield a store ID throws a descriptive error instead of producing a silently wrong URL, since the token format is undocumented (parity-with-SDK, not a stable API).

## Test plan
- [x] Added tests covering: public URL derived from token, private-host URL, `pathPrefix`, special characters in filename, store ID from explicit `storeId` (OIDC), from `BLOB_STORE_ID` env, from `BLOB_READ_WRITE_TOKEN` env, token-over-storeId precedence, and descriptive errors for malformed token / no credentials
- [x] `pnpm test` in `packages/storage-vercel` — 48/48 passing
- [x] `pnpm build` (via turbo, including `@opensaas/stack-core` and `@opensaas/stack-storage` deps) — passes
- [x] `pnpm lint` (eslint on changed files) — no errors
- [x] `pnpm manypkg fix` / `prettier --check` — no changes needed
- [x] Changeset added (`patch` for `@opensaas/stack-storage-vercel`)

Closes #768

---
_Generated by [Claude Code](https://claude.ai/code/session_01UnqgPzigsjAVNEwzhazSxh)_

### DIFF
--- a/.changeset/fix-vercel-blob-get-url.md
+++ b/.changeset/fix-vercel-blob-get-url.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-storage-vercel': patch
+---
+
+Fix `getUrl()` returning a fabricated host that never resolves. It now derives the store ID from the configured token/store ID (the same way the SDK does internally) and embeds it along with the access mode, matching the real URL the SDK returns at upload time. A token that doesn't yield a store ID now throws a descriptive error instead of producing a silently wrong URL.
--- a/packages/storage-vercel/src/index.ts
+++ b/packages/storage-vercel/src/index.ts
@@ -4,6 +4,24 @@ import { randomBytes } from 'node:crypto'
 import type { StorageProvider, UploadOptions, UploadResult } from '@opensaas/stack-storage'
 
 /**
+ * Derives the store ID embedded in a Vercel Blob read-write token
+ * (`vercel_blob_rw_<storeId>_<secret>`), the same way `@vercel/blob` derives
+ * it internally to build blob URLs. This format is undocumented and not a
+ * stable public API — it's parity with the SDK's internal parsing, not a
+ * contract Vercel guarantees. Returns `''` if the token doesn't match the
+ * expected shape.
+ */
+function parseStoreIdFromToken(token: string): string {
+  const segments = token.split('_')
+  return segments.length >= 4 ? segments[3] : ''
+}
+
+/** Strips the `store_` prefix some store ID values are given with. */
+function normalizeStoreId(storeId: string): string {
+  return storeId.startsWith('store_') ? storeId.slice('store_'.length) : storeId
+}
+
+/**
  * Configuration for Vercel Blob storage
  */
 export interface VercelBlobStorageConfig {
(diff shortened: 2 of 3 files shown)


borisno2:
## Self-review

**Overview:** `getUrl()` on the Vercel Blob provider previously returned `https://blob.vercel-storage.com/<pathname>`, a host missing the store subdomain that always 404s. The fix derives the store ID (from the configured `token`/`storeId`, or the `BLOB_READ_WRITE_TOKEN`/`BLOB_STORE_ID` env vars — same precedence and parsing `@vercel/blob` uses internally) and builds `https://<storeId>.<access>.blob.vercel-storage.com/<pathname>`, matching the URL the SDK returns at upload time.

**Correctness**
- Store ID resolution precedence mirrors `@vercel/blob`'s `resolveBlobAuth`: an explicit `token` wins (its embedded store ID is authoritative, even over an explicit `storeId`), then `storeId`/`BLOB_STORE_ID`, then `BLOB_READ_WRITE_TOKEN`. Verified against the real SDK source (`chunk-CIIQSN42.js`, `parseStoreIdFromReadWriteToken`/`constructBlobUrl`).
- A token that doesn't parse to a store ID throws a descriptive `Error` instead of silently returning a broken URL, per the acceptance criteria.
- Private-mode stores get the `.private.` host segment — consistent with the private-access pattern from #767 (`getSignedUrl` already uses `access: 'private'`).
- Pushed a follow-up commit (91aaf0c) trimming an explicit `storeId` config value the same way the SDK trims it — the initial version only trimmed the env var fallback.

**Test coverage:** public/private URL construction, `pathPrefix`, special characters in filenames, store ID from explicit `storeId` (OIDC path), from `BLOB_STORE_ID` env, from `BLOB_READ_WRITE_TOKEN` env, token-over-explicit-storeId precedence, and descriptive errors for a malformed token and for no credentials at all. 48/48 tests passing.

**Non-blocking note:** when both `token` and `storeId` are set and the token turns out to be malformed, the current implementation throws rather than falling back to the explicit `storeId`. This matches the SDK's own precedence (token always wins), so I left it as-is rather than deviating from parity — flagging in case a friendlier fallback is preferred.

**Verified:** `pnpm test` (48/48), `pnpm build` via turbo (including `stack-core`/`stack-storage` deps), `eslint` on changed files, `prettier --check`, `pnpm manypkg fix` (no changes needed).

---

PR #786, opened 2026-07-21
fix(storage-vercel): add allowOverwrite so stable-filename uploads can replace existing blobs

### LINKED ISSUE #769: storage-vercel: overwrites throw for stable filenames — add allowOverwrite with a smart default
## What to build

Re-uploading to the same pathname works when unique filename generation is disabled.

The Blob API rejects overwrites by default, and the provider never sends `allowOverwrite`, so with `generateUniqueFilenames: false` any upload to an existing pathname throws "blob already exists". That makes stable-filename mode unusable for replace workflows (e.g. a field configured with `cleanupOnReplace`).

- Add `allowOverwrite?: boolean` to the Vercel Blob storage config.
- When unset, default to `true` iff `generateUniqueFilenames === false` (stable names imply replace semantics); an explicit setting always wins.
- Docs describe the option and its default.

## Acceptance criteria

- [ ] Upload to an existing pathname succeeds when `generateUniqueFilenames: false` and `allowOverwrite` is unset
- [ ] Explicit `allowOverwrite: false` restores reject-on-overwrite even with stable names; explicit `true` permits overwrite with generated names
- [ ] Tests model the real API default (reject overwrite unless the flag is sent) and cover the default matrix
- [ ] Docs describe the option and its default
- [ ] Minor changeset for `@opensaas/stack-storage-vercel`

## Blocked by

None — can start immediately.

borisno2:
## Summary
- The Vercel Blob API rejects uploads to an existing pathname unless `allowOverwrite` is sent, but the provider never sent it — so with `generateUniqueFilenames: false`, re-uploading to the same pathname (e.g. a field configured with `cleanupOnReplace`) always threw "blob already exists".
- Added `allowOverwrite?: boolean` to `VercelBlobStorageConfig`. When unset, it defaults to `true` iff `generateUniqueFilenames === false` (stable names imply replace semantics); an explicit setting always wins in either direction.
- Updated the package README and `docs/content/reference/storage.md` to describe the option and its default.
- Added a minor changeset for `@opensaas/stack-storage-vercel`.

## Test plan
- [x] New tests cover the default matrix (unique filenames → no `allowOverwrite` sent, matching the real API's reject-by-default behavior; stable filenames → defaults to `true`) and explicit overrides in both directions
- [x] `pnpm vitest run` in `packages/storage-vercel` (53 tests passing)
- [x] `pnpm lint` passes
- [x] `pnpm turbo build` for `@opensaas/stack-core`, `@opensaas/stack-storage`, `@opensaas/stack-storage-vercel` passes

Closes #769

---
_Generated by [Claude Code](https://claude.ai/code/session_017TS31DXc6CAK1U5SfwL8vv)_

### DIFF
--- a/.changeset/tame-otters-overwrite.md
+++ b/.changeset/tame-otters-overwrite.md
@@ -1,0 +1,13 @@
+---
+'@opensaas/stack-storage-vercel': minor
+---
+
+Add `allowOverwrite` config option to `vercelBlobStorage`. The Vercel Blob API rejects uploads to an existing pathname unless this is sent, which made stable-filename replace workflows (`generateUniqueFilenames: false`, e.g. a field with `cleanupOnReplace`) throw "blob already exists". `allowOverwrite` now defaults to `true` when `generateUniqueFilenames` is `false`, and `false` otherwise; an explicit setting always wins.
+
+```typescript
+vercelBlobStorage({
+  token: process.env.BLOB_READ_WRITE_TOKEN,
+  generateUniqueFilenames: false,
+  allowOverwrite: false, // opt back into reject-on-overwrite with stable names
+})
+```
--- a/docs/content/reference/storage.md
+++ b/docs/content/reference/storage.md
@@ -264,6 +264,24 @@ storage: {
 }
 ```
 
+The Blob API rejects uploads to an existing pathname unless `allowOverwrite` is
+sent. With `generateUniqueFilenames: false` (stable filenames), `allowOverwrite`
+defaults to `true` since re-uploading to the same pathname is the expected
+replace workflow (e.g. a field configured with `cleanupOnReplace`). With
+generated filenames, collisions aren't expected, so it's left at the API's
+default of rejecting overwrites. Set `allowOverwrite` explicitly to override
+either default:
+
+```typescript
+storage: {
+  avatars: vercelBlobStorage({
+    token: process.env.BLOB_READ_WRITE_TOKEN,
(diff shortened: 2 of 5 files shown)


borisno2:
## Self-review

**What this does:** Adds `allowOverwrite?: boolean` to `VercelBlobStorageConfig`. `resolveAllowOverwrite()` returns the explicit config value if set; otherwise `true` when `generateUniqueFilenames === false`, else `undefined` (key omitted, so the Blob API's own reject-by-default behavior applies unchanged for the generated-filename path).

**Correctness:** Matches the issue's acceptance criteria exactly — stable-filename uploads to an existing pathname now succeed by default, an explicit `false` restores reject-on-overwrite with stable names, an explicit `true` permits overwrite with generated names, and the default case for generated filenames sends no `allowOverwrite` key at all (verified against `@vercel/blob@2.6.1`'s real `PutCommandOptions` type, which defaults to reject).

**Tests:** New `describe('allowOverwrite', ...)` block covers the full default matrix plus both explicit-override directions, and a case modeling the real API's reject error for the default (no-override) path. All 53 tests pass.

**Docs/changeset:** Package README and `docs/content/reference/storage.md` both describe the option and its default; changeset added as `minor` (new config option) for `@opensaas/stack-storage-vercel`.

No issues found. Local verification: `pnpm vitest run` (packages/storage-vercel), `pnpm turbo build` (core, storage, storage-vercel), and `pnpm lint` all pass clean.


---

PR #792, opened 2026-07-21
fix(storage-vercel): extension-less filenames and cacheControlMaxAge: 0

### LINKED ISSUE #770: storage-vercel: small fixes — extension-less filenames, cacheControlMaxAge: 0, type-only import
## What to build

Three small correctness/hygiene fixes to the Vercel Blob provider:

- Unique filename generation treats an extension-less original (e.g. `README`) as all-extension, producing names like `<timestamp>-<id>README`. Use the same extension semantics as the local provider (empty extension when none — `path.extname` behaviour).
- `cacheControlMaxAge: 0` is dropped by a truthiness check; an explicit zero should be passed through to the SDK.
- The SDK's put-options type is imported as a value; make it a type-only import per the repo's ESM conventions.

## Acceptance criteria

- [ ] Extension-less filenames generate `<timestamp>-<id>` with no trailing original-name fragment; real extensions are preserved
- [ ] `cacheControlMaxAge: 0` reaches the SDK call
- [ ] SDK types are imported type-only; `pnpm lint` passes
- [ ] Tests cover the extension-less filename and zero max-age cases
- [ ] Patch changeset for `@opensaas/stack-storage-vercel`

## Blocked by

None — can start immediately.

borisno2:
## Summary
- Unique filename generation now uses `path.extname` semantics (matching the local storage provider), so an extension-less original like `README` produces a clean generated name with no trailing original-name fragment, while real extensions are still preserved.
- `cacheControlMaxAge: 0` is now passed through to the `@vercel/blob` `put()` call instead of being dropped by a truthiness check.
- Verified the SDK's put-options type (`PutCommandOptions`) is already imported type-only on `main`, so no change was needed there.

## Test plan
- [x] Added tests covering the extension-less filename case and the `cacheControlMaxAge: 0` case (both the explicit-zero and omitted-by-default paths)
- [x] `pnpm test` in `packages/storage-vercel` (56 passed)
- [x] `pnpm lint` passes
- [x] `pnpm build` for `@opensaas/stack-core`, `@opensaas/stack-storage`, `@opensaas/stack-storage-vercel`
- [x] Patch changeset added for `@opensaas/stack-storage-vercel`

Closes #770

---
_Generated by [Claude Code](https://claude.ai/code/session_012Vg1ySDW5c2NpX77ZxHQsP)_

### DIFF
--- a/.changeset/quiet-otters-jump.md
+++ b/.changeset/quiet-otters-jump.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-storage-vercel': patch
+---
+
+Fix unique filename generation to use `path.extname` semantics (matching the local provider) so extension-less originals no longer get the whole name appended, and pass through an explicit `cacheControlMaxAge: 0` instead of dropping it.
--- a/packages/storage-vercel/src/index.ts
+++ b/packages/storage-vercel/src/index.ts
@@ -1,6 +1,7 @@
 import { put, del, get, issueSignedToken, presignUrl, BlobNotFoundError } from '@vercel/blob'
 import type { PutCommandOptions } from '@vercel/blob'
 import { randomBytes } from 'node:crypto'
+import path from 'node:path'
 import type { StorageProvider, UploadOptions, UploadResult } from '@opensaas/stack-storage'
 
 /**
@@ -96,7 +97,7 @@ export class VercelBlobStorageProvider implements StorageProvider {
       return originalFilename
     }
 
-    const ext = originalFilename.substring(originalFilename.lastIndexOf('.'))
+    const ext = path.extname(originalFilename)
     const uniqueId = randomBytes(16).toString('hex')
     const timestamp = Date.now()
     return `${timestamp}-${uniqueId}${ext}`
@@ -196,7 +197,7 @@ export class VercelBlobStorageProvider implements StorageProvider {
       ...this.authOptions(),
     }
 
-    if (this.config.cacheControlMaxAge) {
+    if (this.config.cacheControlMaxAge !== undefined) {
       uploadOptions.cacheControlMaxAge = this.config.cacheControlMaxAge
     }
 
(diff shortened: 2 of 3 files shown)


borisno2:
## Self-review

**Overview:** Two small correctness fixes to `@opensaas/stack-storage-vercel`'s `VercelBlobStorageProvider`, per issue #770.

**Changes:**
- `generateFilename`: switched from `originalFilename.substring(originalFilename.lastIndexOf('.'))` to `path.extname(originalFilename)`, matching `packages/storage/src/providers/local.ts`. Previously an extension-less name like `README` had `lastIndexOf('.')` return `-1`, and `substring(-1)` clamps to `0`, so the *entire original filename* was appended as a bogus "extension" (e.g. `1690000000-<hash>README`). Now it correctly resolves to `''`, so generated names for extension-less files are clean (`<timestamp>-<hash>`), while real extensions (`.txt`, `.jpg`, …) are preserved unchanged.
- `upload()`: changed `if (this.config.cacheControlMaxAge)` to `if (this.config.cacheControlMaxAge !== undefined)` so an explicit `cacheControlMaxAge: 0` reaches the `@vercel/blob` `put()` call instead of being silently dropped by the truthiness check.
- Checked the third item in the issue (type-only import of `PutCommandOptions`) — it's already type-only on `main` (fixed incidentally in an earlier PR), so no change was needed there.

**Tests added:**
- Extension-less original (`README`) produces `${timestamp}-${hash}` with no trailing fragment.
- `cacheControlMaxAge: 0` is passed through to `put()`.
- `cacheControlMaxAge` is omitted entirely from the options object when not configured (guards against a regression back to always-set).

**Verification:**
- `pnpm test` in `packages/storage-vercel` — 56/56 passing.
- `pnpm build` for `@opensaas/stack-core`, `@opensaas/stack-storage`, `@opensaas/stack-storage-vercel` — clean.
- `pnpm lint` — 0 errors (3 pre-existing warnings unrelated to this change).
- `pnpm manypkg fix` / `pnpm format` — no changes needed.
- Patch changeset added for `@opensaas/stack-storage-vercel`.

No issues found — change is minimal, scoped, and consistent with existing conventions (mirrors the local provider's extension handling exactly).


---

PR #794, opened 2026-07-21
fix(core): validate multi-column fields before splitting into physical columns

### LINKED ISSUE #789: Hook Pipeline: validate multi-column fields before splitting into physical columns
## What to build

Today, a field that implements the multi-column extension point (`getColumnNames`/`assembleColumns`/`splitColumns` — currently only `@opensaas/stack-storage`'s `image()`/`file()` in `db.columns: 'keystone'` mode) has its `splitColumns` run **inline inside `executeFieldResolveInputHooks`** (Hook Pipeline Phase 1.5), which deletes the logical field key and replaces it with per-part physical columns *before* list `validate` → field `validate` → `validateFieldRules` (Phases 2–3) ever run.

The practical effect: a field's `resolveInput` hook can have a documented "unknown/unrecognised value — return as-is and let validation catch it" fallback branch, but in multi-column mode that promise is false. The unrecognised value gets split by `splitColumns` (which treats anything that isn't a valid shape as equivalent to `null`) into `null`/`undefined` physical columns, `validateFieldRules` never sees the logical field at all (it was already deleted), and the write succeeds silently with corrupted/absent data instead of throwing a validation error.

Reorder the Hook Pipeline / Write Pipeline so that:

- Field `resolveInput` still runs at Phase 1.5, but returns its resolved value under the **logical field key** — no split yet.
- List `validate` → field `validate` → `validateFieldRules` (Phases 2–3) run against that logical-keyed value, exactly as they do for any single-column field — so a field's Zod schema (`getZodSchema`) actually gets a chance to reject an unrecognised shape.
- Only **after** `validateFieldRules` passes does a new step call the field's `splitColumns`, replacing the logical key with its physical per-part columns, before `filterWritableFields` (Phase 5) runs.

This has three call sites to update in lockstep, since each runs the same resolveInput → validate → validateFieldRules → filterWritableFields sequence independently:

- The top-level Write Pipeline (`packages/core/src/context/write-pipeline.ts`, via `packages/core/src/context/hook-pipeline.ts`)
- Nested relationship writes on **create** (`packages/core/src/context/nested-operations.ts`)
- Nested relationship writes on **update** (`packages/core/src/context/nested-operations.ts`)

The existing field-level write-access gate on the split (a denied field drops its logical key and contributes none of its per-part columns — see the `checkFieldAccess` call currently inside the split branch of `executeFieldResolveInputHooks`) must be preserved exactly as-is; only the *timing* relative to validation changes, not the access semantics.

## Acceptance criteria

- [ ] A multi-column field whose `resolveInput` returns an unrecognised, non-nullish value now causes the write to throw a `ValidationError` (from the field's `getZodSchema` via `validateFieldRules`), for both `create` and `update`, at the top-level write path.
- [ ] The same holds through at least one nested-relationship write path (create and update).
- [ ] The three legitimate shapes a multi-column field can receive — `null`/`undefined`, an already-shaped metadata object, and (for storage's fields) a `File`-like upload trigger — continue to write identically to today (no behavior change for the happy path).
- [ ] Field-level write-access gating on the split (deny → drop the field, write none of its columns; sudo bypasses) is unchanged.
- [ ] `packages/core/src/access/multi-column-read-write.test.ts` is updated so its generic multi-column fixture (and any new cases) exercises the new phase ordering — re-targeted to whichever function/step now owns the split, without weakening the existing access-gate and null-split coverage.
- [ ] `pnpm test` passes in `packages/core`.
- [ ] A changeset is added per the `pr-changeset` skill (patch — bug fix, no public API change).

## Blocked by

None — can start immediately.

borisno2:
## Summary

- Moves the `splitColumns` step for multi-column fields (currently only `@opensaas/stack-storage`'s `image()`/`file()` in `db.columns: 'keystone'` mode) so it runs **after** `validateFieldRules` instead of inline inside `executeFieldResolveInputHooks` (Hook Pipeline Phase 1.5, before validation).
- Previously, an unrecognised/non-nullish value returned by a field's `resolveInput` "let validation catch it" fallback was split into `null`/`undefined` physical columns and written silently — `validateFieldRules` never saw the logical field because it had already been deleted from `resolvedData`.
- `executeFieldResolveInputHooks` (`packages/core/src/hooks/index.ts`) now only resolves each field's value under its logical key — no split. A new `splitMultiColumnFields` function performs the split, gated by the exact same field-level write-access check (`checkFieldAccess`) that previously ran inline, so a denied field still drops its logical key and contributes none of its per-part columns (sudo still bypasses).
- Updated the three call sites in lockstep, since each independently runs the same `resolveInput → validate → validateFieldRules → filterWritableFields` sequence:
  - `packages/core/src/context/hook-pipeline.ts` (top-level Write Pipeline) — added as Phase 4, after Phase 3 (`validateFieldRules`).
  - `packages/core/src/context/nested-operations.ts` — `processNestedCreate` and `processNestedUpdate` (nested relationship writes).
- Updated `CONTEXT.md`'s Hook Pipeline glossary entry and comments in `field-access.ts` that referenced the old gate location.

## Test plan

- [x] `packages/core/src/access/multi-column-read-write.test.ts` retargeted: the split-behavior tests now drive the new `splitMultiColumnFields` function directly (access-gate and null-split coverage preserved unweakened); added a new describe block proving `executeFieldResolveInputHooks` no longer splits and passes unrecognised values through unsplit under the logical key.
- [x] Added `packages/core/tests/multi-column-validation-ordering.test.ts` — full `context.db` integration tests proving:
  - top-level `create`/`update` with an unrecognised multi-column value throws `ValidationError` and never reaches the DB;
  - the same holds through a nested relationship `create` and `update`;
  - the legitimate shapes (`null`/`undefined`, a valid shaped object) continue to split and persist identically to before.
- [x] `pnpm test` passes in `packages/core` (44 files / 898 tests) and `packages/storage` (7 files / 148 tests, unaffected).
- [x] `pnpm build` (core), `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all clean.
- [x] Changeset added (patch — bug fix, no public API change).

Closes #789

---
_Generated by [Claude Code](https://claude.ai/code/session_01QPX69aoqSuBMAwD3bDCSWk)_


borisno2:
## Self-review

### Overview
Moves the `splitColumns` step for multi-column fields (storage `image()`/`file()` in `db.columns: 'keystone'` mode) so it runs **after** `validateFieldRules` instead of inline inside `executeFieldResolveInputHooks` (which ran before validation). `executeFieldResolveInputHooks` now only resolves each field's value under its logical key; a new `splitMultiColumnFields` (in `packages/core/src/hooks/index.ts`) performs the split + field-access gate, called from all three write paths that independently run `resolveInput → validate → validateFieldRules → filterWritableFields`: the top-level Hook Pipeline (`hook-pipeline.ts`, new Phase 4) and nested create/update (`nested-operations.ts`).

### Correctness
- Verified exactly 3 call sites for the new split step (`hook-pipeline.ts` ×1, `nested-operations.ts` ×2) — matches the issue's "three call sites... in lockstep" requirement, confirmed by grep.
- Field-level write-access gate is preserved byte-for-byte: same `checkFieldAccess` call, same args, same "drop logical key on deny" behavior, sudo bypass intact.
- `connectOrCreate`'s create branch reuses `processNestedCreate`, so it inherits the fix without a separate call site.
- Confirmed via `tsc --noEmit` that the `resolvedData` type threading through `let resolvedData = ...` reassignments across `executeResolveInput` → `executeFieldResolveInputHooks` → `splitMultiColumnFields` stays consistent.

### Minor (non-blocking) observation
`applyCreateDefaults` (Phase 1.75) now runs **before** the split instead of after (previously split was at old Phase 1.5, before defaults). This means a multi-column field with a `defaultValue` that's omitted from input will now actually get its default value split into physical columns, whereas previously it wouldn't have been split at all (and would likely have broken the write, since the raw JSON-shaped logical key isn't a real Prisma column in multi-column mode). This looks like an incidental improvement rather than a regression, but it's not explicitly covered by a test — worth keeping in mind if `image()`/`file()` ever gain `defaultValue` support in multi-column mode.

### Test coverage
- `multi-column-read-write.test.ts` retargeted: split-behavior tests now drive `splitMultiColumnFields` directly; access-gate and null-split coverage preserved unweakened; added coverage proving `executeFieldResolveInputHooks` no longer splits and passes unrecognised values through under the logical key.
- New `multi-column-validation-ordering.test.ts`: full `context.db` integration tests proving an unrecognised value throws `ValidationError` before touching the DB, for top-level create/update AND nested create/update, plus happy-path (null + valid shape) parity.
- `pnpm test` in `packages/core`: 44 files / 898 tests pass. `packages/storage`: 7 files / 148 tests pass (unaffected — its tests exercise `resolveInput` directly, not the removed split logic).
- `pnpm build`, `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all clean.

### Security
No change to the access-control surface — the field-level write-access gate around the split is identical to before, just relocated. No new undeclared-key or bypass paths introduced (`filterWritableFields`'s existing `splitColumnOwners` gate for direct raw-column writes is untouched).

No blocking issues found.


---

PR #798, opened 2026-07-22
fix(ui): strip synthetic _count from item-view details data

### LINKED ISSUE #797: Strip `_count` from item-view data so edit-form Save works on any to-many-relationship list
## What to build

In `@opensaas/stack-ui`'s derived item view (`ItemForm.tsx` → `ItemViewLayoutView`, issue #734/#752), the record fetch adds a synthetic `_count` select (for the relationship-table "Showing N of M" footer) alongside the real fields. That `_count` key is never stripped before the record is split into "details form data," so it rides through `prepareItemForm` → `ItemFormClient` → `useItemForm`'s `transformItemFormData` and ends up as a raw key in every update payload. The server correctly rejects it (`Cannot update "_count": it is not a field of this list`), so **Save silently fails on the edit page for any list with at least one non-`picker`-mode to-many relationship field** — a severe regression affecting most of the generic-CRUD admin surface.

The fix has two parts, delivered together in one PR:

1. **Direct fix** — stop `_count` from ever reaching `detailsItemData`/`initialData`/client form state. `ItemViewLayoutView` already reads the count separately (`itemData?._count`) for the footer payload, so the key doesn't need to survive the split.
2. **Hardening (defense-in-depth)** — `transformItemFormData` (`useItemForm.ts`) should only emit keys that exist in the `fields` map it's given, dropping anything else on submit. This guards against any other non-field key (not just `_count`) accidentally reaching the client/server boundary in the future, and turns a future recurrence into a silent no-op removal instead of a full save failure — which is why it ships alongside the direct fix rather than as a substitute for it.

End-to-end: editing and saving a record on a list with a `many: true` relationship (in the default item-view display mode, not `picker`) succeeds and persists the change, for lists with one or more to-many relationship fields, including self-referential relationships.

## Acceptance criteria

- [ ] Editing and saving a record on a list with a `many: true` relationship field (default item-view mode) succeeds — no `_count`-related validation error, and the field change persists.
- [ ] `_count` is stripped before `detailsItemData` is built in `ItemViewLayoutView`, not merely worked around downstream.
- [ ] `transformItemFormData` drops any key from `formData` that has no corresponding entry in the `fields` map, verified by a unit test that injects a bogus non-field key and asserts it's absent from the transformed submission payload.
- [ ] A regression test (component or e2e) covers fetch → edit → save for a list with a to-many relationship, reproducing the original failure prior to the fix.
- [ ] The existing `picker`-mode workaround continues to work unchanged (no regression for lists already using `ui.itemView.displayMode: 'picker'`).
- [ ] Changeset added per the `pr-changeset` skill (patch — bug fix).

## Blocked by

None — can start immediately.

borisno2:
## Summary

- The derived item view's record fetch (`ItemViewLayoutView` in `ItemForm.tsx`) adds a synthetic `_count` select alongside the real fields, for the Relationship table's "Showing N of M" footer. That key was never stripped before the record was split into "details form data," so it rode through `prepareItemForm` → `ItemFormClient` → `useItemForm`'s `transformItemFormData` and ended up as a raw key in every update payload — which the server correctly rejected (`Cannot update "_count": it is not a field of this list`), silently breaking Save on the edit page for any list with at least one non-`picker`-mode to-many relationship field.
- **Direct fix**: extracted the details-data filtering into a new exported `buildDetailsItemData(itemData, layout)` helper that excludes both the section fields and `_count`, so `_count` never reaches `detailsItemData`/`initialData`/client form state.
- **Hardening (defense-in-depth)**: `transformItemFormData` (`useItemForm.ts`) now drops any key from `formData` that has no corresponding entry in the `fields` map it's given, turning any future non-field key that reaches this boundary into a silent no-op removal instead of a full save failure.

## Test plan

- [x] Unit test: `transformItemFormData` drops a bogus non-field key (`_count`) from the submit payload.
- [x] Regression test threading the real `buildDetailsItemData` → `prepareItemForm` → `transformItemFormData` pipeline for a list with a `many: true` relationship table, asserting `_count` never survives into the final submit payload (reproduces the original failure prior to the fix).
- [x] Existing `picker`-mode item-view tests pass unchanged (no regression).
- [x] `pnpm test` (full `@opensaas/stack-ui` suite — 529 tests) passes.
- [x] `pnpm build` (typecheck) passes.
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` run clean.
- [x] Changeset added (patch — bug fix).

Closes #797

---
_Generated by [Claude Code](https://claude.ai/code/session_01V8boyfgPM1q3xt9m2VHKMx)_

### DIFF
--- a/.changeset/thirty-cobras-shout.md
+++ b/.changeset/thirty-cobras-shout.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+Fix Save failing on the edit page for any list with a `many: true` relationship table: strip the synthetic `_count` payload before it reaches the details form data, and harden `transformItemFormData` to drop any submitted key with no matching field.
--- a/packages/ui/src/components/ItemForm.tsx
+++ b/packages/ui/src/components/ItemForm.tsx
@@ -104,13 +104,35 @@ export function readAccessScopedTotal(
 }
 
 /**
+ * Strip a fetched item-view record down to the details-card fields (issue
+ * #797): every Relationship-table section's field (rendered separately as a
+ * read-only table) AND the synthetic `_count` payload the fetch adds
+ * alongside the real fields (for {@link readAccessScopedTotal}'s "showing N of
+ * M" footer). Neither belongs in the details form's data — `_count` in
+ * particular is not a field of the list, so if it survived into the submit
+ * payload the server would reject the whole update.
+ */
+export function buildDetailsItemData(
+  itemData: Record<string, unknown>,
+  layout: ItemViewLayout,
+): Record<string, unknown> {
+  const detailsItemData: Record<string, unknown> = {}
+  for (const [key, value] of Object.entries(itemData)) {
+    if (key !== '_count' && !layout.sections.some((section) => section.fieldName === key)) {
+      detailsItemData[key] = value
+    }
+  }
+  return detailsItemData
+}
+
+/**
  * Edit-mode item view whose layout is DERIVED from the list shape (issue #734):
  * scalar/to-one fields (and picker-demoted relationships) in a details card
  * with the existing whole-form Save/Cancel, and each to-many relationship as a
  * read-only Relationship table. The arrangement follows from the number of
  * tables — one → two-column split, several → stacked.
  */
-async function ItemViewLayoutView({
+export async function ItemViewLayoutView({
   context,
   config,
   listConfig,
@@ -191,12 +213,7 @@ async function ItemViewLayoutView({
     ...listConfig,
     fields: Object.fromEntries(detailsFieldEntries),
   }
-  const detailsItemData: Record<string, unknown> = {}
-  for (const [key, value] of Object.entries(itemData)) {
-    if (!layout.sections.some((section) => section.fieldName === key)) {
-      detailsItemData[key] = value
-    }
-  }
+  const detailsItemData = buildDetailsItemData(itemData, layout)
 
   const { serializableFields, initialData, relationshipData } = await prepareItemForm(
     context,
--- a/packages/ui/src/components/ItemForm.tsx
+++ b/packages/ui/src/components/ItemForm.tsx
@@ -132,7 +132,7 @@ export function buildDetailsItemData(
  * read-only Relationship table. The arrangement follows from the number of
  * tables — one → two-column split, several → stacked.
  */
-export async function ItemViewLayoutView({
+async function ItemViewLayoutView({
   context,
   config,
   listConfig,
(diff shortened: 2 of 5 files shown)


borisno2:
## Self-review

**Overview**

Fixes #797: `ItemViewLayoutView`'s item-view fetch adds a synthetic `include._count` select for the Relationship table's "Showing N of M" footer, and that key was riding, unfiltered, all the way into the update payload — the server rejected it as an undeclared field, silently breaking Save on any edit page for a list with a `many: true` relationship rendered as a table.

**Changes**

1. `packages/ui/src/components/ItemForm.tsx` — extracted the details-data filtering (previously inlined) into an exported `buildDetailsItemData(itemData, layout)`, which now also excludes `_count` in addition to relationship-table section fields. Follows the file's existing pattern of exporting small pure helpers (`buildItemViewInclude`, `readAccessScopedTotal`) off an otherwise-private async server component for unit testability.
2. `packages/ui/src/lib/useItemForm.ts` — `transformItemFormData` now drops any `formData` key with no matching entry in `fields`, as defense-in-depth against any other non-field key reaching the submit payload in the future.

**Code quality / correctness**

- Both fixes are minimal and targeted; no unrelated refactors.
- Verified system fields (`id`/`createdAt`/`updatedAt`) aren't broken by the `transformItemFormData` change: they were never in the `fields` map to begin with (confirmed against `packages/core/src/access/field-access.ts`, which explicitly skips them from its own undeclared-key check), so they're already stripped downstream regardless — now just stripped one layer earlier, which is harmless.
- Checked for multi-column/split-column fields that might rely on `formData` carrying keys absent from `fields` — the UI package has no client-side multi-column field support today, so the new `if (!fieldConfig) continue` guard has no legitimate case to break.
- Self-review caught one loose end: I'd initially exported `ItemViewLayoutView` for a full-render test, then switched the regression test to exercise `buildDetailsItemData` directly (RTL/happy-dom in this repo can't render nested async Server Components like `RelationshipTable`). The now-unused export was reverted in the follow-up commit — `ItemViewLayoutView` stays module-private.

**Test coverage**

- `useItemForm.test.ts`: unit test that `transformItemFormData` drops a bogus `_count` key.
- `ItemFormCountStrip.test.tsx` (new): threads the real `buildDetailsItemData` → `prepareItemForm` → `transformItemFormData` pipeline for a list with a to-many relationship table and asserts `_count` never survives to the submit payload — reproduces the original failure mode without needing full component rendering.
- Full `@opensaas/stack-ui` suite (529 tests) passes; existing `picker`-mode item-view tests are unaffected.

**Risks**

- Low. The change narrows what reaches the update payload; it can only fix previously-broken saves (any list with a to-many relationship table) or make already-inert keys (system fields) disappear one step earlier — no legitimate submitted field is dropped.

No changes requested.


---

PR #826, opened 2026-07-29
fix(e2e): derive relationship-count spec slugs per-attempt

### LINKED ISSUE #825: e2e: relationship-count spec uses fixed slugs, so Playwright retries can never pass

Found while investigating a red `e2e` check on #824 (a docs-only PR that cannot affect the admin UI).

## What happens

`e2e/starter-auth/09-relationship-count-columns.spec.ts` creates its posts with slug prefixes fixed per test — `count-display-0…N`, `count-sort-0…N`, `count-url-0…N`, `count-builder-0…N`. In `examples/starter-auth`, `Post.slug` is declared `isIndexed: 'unique'`, which is a **global** uniqueness constraint, not per-user.

Each test's `beforeEach` signs up a fresh user, so a first attempt writes `count-builder-0…3` successfully. On a Playwright retry the test signs up a *different* fresh user but attempts the *same* slugs, the creates collide with the rows the first attempt already wrote, and the retry's user ends up with **zero** posts. The assertions then fail against the retry's own data rather than against the behaviour under test.

## Observed

CI run [30443383817](https://github.com/OpenSaasAU/stack/actions/runs/30443383817), job `e2e` — 67 passed, 1 failed:

- **Attempt 1** — a genuine flake in the Filter builder path: after clicking Apply, `page.waitForURL(/search=posts:>3/)` timed out at 10 s.
- **Retry #1 and #2** — both got past the URL wait, then `expect(locator('tbody tr').filter({ hasText: user.email })).toHaveCount(1)` received `0`, consistently, for the full 5 s timeout.

The retries' `0` is the slug collision, not the flake. The sibling test that drives the same comparison through a shared URL (`?search=posts:%3E3`) passes, so the filter engine and the count filter itself are working.

## Why it matters beyond one test

Every test in the file shares this pattern, so the retry safety net is inert for the whole spec: any transient failure there escalates to a hard red, and the retry output actively misleads (a navigation flake presents as "the count filter returns no rows"). That is what made a docs-only PR look like it broke relationship-count filtering.

## Agent Brief

**Category:** bug
**Summary:** Make the relationship-count e2e spec's fixtures unique per attempt so retries exercise the behaviour instead of colliding with the previous attempt's rows.

**Current behavior:**
The spec's post-creation helper takes a fixed string prefix and derives slugs from it plus an index. Because the example's `Post.slug` is globally unique and each attempt signs up a new user, a retried test cannot create its posts and asserts against a user with none.

**Desired behavior:**
Each test attempt creates rows that cannot collide with any other attempt or test. Slugs should derive from something already unique per attempt — the generated test user's email/id is the natural source, since the spec already relies on it to isolate rows in assertions. A retry then goes through the same setup as a first attempt and fails only if the behaviour under test is actually broken.

**Key interfaces:**
- The spec's post-creation helper (currently taking a plain prefix) and its call sites.
- No change to `examples/starter-auth`'s `Post.slug` uniqueness — global uniqueness is correct for a slug; the test fixtures are what should be unique.

**Acceptance criteria:**
- [ ] Running the spec twice in a row against the same database passes both times
- [ ] Forcing a failure in one test and letting Playwright retry it shows the retry reaching the same assertion state as the first attempt (not `0` rows)
- [ ] No test in the file creates a slug that another test or attempt could also create
- [ ] The rest of the suite is unaffected

**Out of scope:**
- The underlying Filter-builder navigation flake (Apply not landing within 10 s) — worth its own investigation; this issue is only about the fixtures making it unrecoverable and misdiagnosable
- Changing retry counts or timeouts
- Any change to the count column, count sort, or count filter behaviour

borisno2:
## Summary
- `e2e/starter-auth/09-relationship-count-columns.spec.ts` built post slugs from a fixed prefix + index (`count-display-0…N`, etc.). `examples/starter-auth`'s `Post.slug` is globally unique, and each test attempt signs up a fresh user, so a Playwright retry hit the same slugs as the first attempt, its creates collided, and it ended up asserting against a user with zero posts — a navigation flake then presented as "the count filter returns no rows".
- `createPosts` now derives its slug prefix from the per-attempt test user's email (`generateTestUser` already makes this unique per attempt, including retries), so no two attempts or tests can ever produce colliding slugs.
- No change to `Post.slug`'s uniqueness, the count column/sort/filter behavior, retry counts, or timeouts — scope is limited to the test fixtures per the issue's "out of scope" list.

## Test plan
- [x] `pnpm lint` — clean (only 3 pre-existing, unrelated warnings)
- [x] `pnpm format` / `pnpm manypkg fix` — no changes needed
- [ ] `pnpm test:e2e` run of `09-relationship-count-columns.spec.ts` against `examples/starter-auth` (requires the dev server + browsers; not run in this environment — please confirm in CI)

Closes #825

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01D6Hey2wbdzPjRXhTGbp5Kz

---
_Generated by [Claude Code](https://claude.ai/code/session_01D6Hey2wbdzPjRXhTGbp5Kz)_

### DIFF
--- a/e2e/starter-auth/09-relationship-count-columns.spec.ts
+++ b/e2e/starter-auth/09-relationship-count-columns.spec.ts
@@ -37,8 +37,14 @@ test.describe('To-many relationship count columns', () => {
   }
 
   async function createPosts(page: Page, count: number, prefix: string) {
+    // Derive the slug from the per-attempt test user (its email is already
+    // unique per attempt, including Playwright retries — see generateTestUser),
+    // not just the fixed `prefix`. Post.slug is globally unique, so a retry
+    // that reused the first attempt's slugs would collide with rows the first
+    // attempt already wrote and end up asserting against its own empty state.
+    const uniquePrefix = `${prefix}-${user.email.split('@')[0]}`
     for (let i = 0; i < count; i++) {
-      await createPost(page, { title: `${prefix} ${i}`, slug: `${prefix}-${i}` })
+      await createPost(page, { title: `${prefix} ${i}`, slug: `${uniquePrefix}-${i}` })
     }
   }
 


borisno2:
## Review

**What this does:** Fixes #825 — `09-relationship-count-columns.spec.ts` built post slugs from a fixed prefix + index. Since `Post.slug` is globally unique and each test attempt signs up a fresh user, a Playwright retry (new user, same slugs as the first attempt) collided on create and ended up asserting against an empty state, misdiagnosing flakes as filter bugs. `createPosts` now folds the per-attempt test user's email (already unique per attempt via `generateTestUser`, including retries) into the slug prefix.

**Correctness**
- `user.email` is `test-${timestamp}-${random}@example.com`; `email.split('@')[0]` yields only lowercase letters, digits, and hyphens — safe as a slug segment, no sanitization needed.
- `user` is captured in the enclosing `describe` scope and reassigned per-`beforeEach`, so `createPosts` always picks up the current attempt's user — no stale-closure risk.
- Titles are intentionally left on the plain prefix; only the globally-unique-constrained `slug` needed to change. Matches the issue's "out of scope" list (no changes to count column/sort/filter behavior, retry counts, or timeouts).

**Acceptance criteria** — all four are satisfied by this change: reruns get fresh users → fresh slugs; a retry goes through the same setup as attempt 1 instead of colliding; no two tests/attempts can produce the same slug; nothing else in the file or suite is touched.

**Nit (non-blocking):** the collision probability isn't literally zero (birthday-bound on `Math.random().toString(36).substring(7)`), but that's the same assumption `generateTestUser` already relies on for email uniqueness elsewhere in this suite, so it's consistent with existing practice rather than a new risk.

No issues found — looks good to merge once CI is green.

---

PR #827, opened 2026-07-29
Fix serial relationship-options fetch in admin item form

### LINKED ISSUE #822: Admin UI item form fetches relationship options serially — N sequential round-trips before first paint
### Version

`@opensaas/stack-ui` 0.31.1 (`@opensaas/stack-core` 0.32.0)

### What happens

The Admin UI edit form for a list with several relationship fields is slow to load, and the delay scales linearly with the number of relationship fields on the list.

On a list with 8 relationship fields the form issues 8–16 sequential database round-trips before it can render.

### Cause

`packages/ui/src/lib/prepareItemForm.ts:74-96` — the relationship-options fetch is an `await` inside a `for` loop:

```ts
for (const [fieldName, fieldConfig] of Object.entries(listConfig.fields)) {
  if (fieldConfigAny.type === 'relationship') {
    ...
    relationshipData[fieldName] = await getRelationshipOptions(...)
  }
}
```

Each iteration blocks the next. The calls are fully independent — each reads a different related list, and nothing in the loop body depends on a previous result. `getRelationshipOptions` itself can issue two queries (the bounded window, plus a second for selected ids falling outside it), so the serial cost is up to 2N round-trips.

This sits on the critical path: the form is server-rendered, so the user waits the full sum before first paint.

### Suggested fix

Collect the relationship fields first, then `Promise.all` the `getRelationshipOptions` calls and assemble `relationshipData` from the results. That turns 2N sequential round-trips into 2 waves.

The existing per-field `try`/`catch` (which degrades a failed field to `[]`) maps onto `Promise.allSettled`, or a `.catch(() => [])` per promise, so error behaviour need not change.

### Note

The per-field fetch is already bounded and projected — take-limited, `id` + label field only — so this is purely about concurrency, not about over-fetching.

### Caveat

Diagnosed by reading the source rather than by profiling. I haven't isolated what share of the observed load time is this loop versus the item read itself.

borisno2:
## Summary

- `prepareItemForm` (`packages/ui/src/lib/prepareItemForm.ts`) fetched each relationship field's options with an `await` inside a `for` loop, causing up to 2N sequential DB round-trips before the item form's first paint (N = number of relationship fields on the list).
- Fetches now run concurrently via `Promise.all`, turning 2N sequential round-trips into 2 concurrent waves.
- Per-field error handling is preserved: a failed fetch still degrades that field to `[]` and logs the error, it just no longer blocks sibling fields from resolving.

## Test plan

- [x] Added a regression test (`fetches relationship options for multiple fields concurrently, not serially`) that uses two never-resolving mock `findMany` promises to assert both fetches are kicked off before either resolves — fails against the old serial `for` loop, passes against the fix.
- [x] `pnpm test` in `packages/ui` — all 530 tests pass
- [x] `pnpm build` in `packages/core` and `packages/ui`
- [x] `pnpm manypkg fix` / `pnpm format` — no changes needed beyond the diff
- [x] Changeset added (`@opensaas/stack-ui` patch)

Closes #822

---
_Generated by [Claude Code](https://claude.ai/code/session_01XpqqF63AJoZPWf2kEuVD97)_

### DIFF
--- a/.changeset/tame-rivers-jump.md
+++ b/.changeset/tame-rivers-jump.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+Fix the admin item form fetching relationship options serially (N sequential round-trips before first paint). Fetches now run concurrently via `Promise.all`.
--- a/packages/ui/src/lib/prepareItemForm.ts
+++ b/packages/ui/src/lib/prepareItemForm.ts
@@ -71,29 +71,32 @@ export async function prepareItemForm(
   // fetch that always unions the item's currently-selected id(s) so their
   // label renders even outside the window.
   const relationshipData: Record<string, Array<{ id: string; label: string }>> = {}
-  for (const [fieldName, fieldConfig] of Object.entries(listConfig.fields)) {
-    const fieldConfigAny = fieldConfig as { type: string; ref?: string; many?: boolean }
-    if (fieldConfigAny.type === 'relationship') {
-      const ref = fieldConfigAny.ref
-      if (ref) {
-        // Parse ref format: "ListName.fieldName"
-        const relatedListName = ref.split('.')[0]
+  const relationshipFields = Object.entries(listConfig.fields).filter(
+    ([, fieldConfig]) => (fieldConfig as { type: string }).type === 'relationship',
+  ) as Array<[string, { type: string; ref?: string; many?: boolean }]>
 
-        if (config.lists[relatedListName]) {
-          try {
-            relationshipData[fieldName] = await getRelationshipOptions(
-              context,
-              config,
-              relatedListName,
-              { selectedIds: extractSelectedIds(itemData[fieldName], fieldConfigAny.many) },
-            )
-          } catch (error) {
-            console.error(`Failed to fetch relationship items for ${fieldName}:`, error)
-            relationshipData[fieldName] = []
-          }
-        }
+  const relationshipResults = await Promise.all(
+    relationshipFields.map(async ([fieldName, fieldConfigAny]) => {
+      const ref = fieldConfigAny.ref
+      if (!ref) return null
+      // Parse ref format: "ListName.fieldName"
+      const relatedListName = ref.split('.')[0]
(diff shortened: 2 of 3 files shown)


borisno2:
## Code Review

**Overview:** Replaces the sequential `for` loop with a serial `await` per relationship field in `prepareItemForm.ts` with a `Promise.all` that kicks off all `getRelationshipOptions` fetches concurrently — exactly the fix suggested in #822. Small, surgical diff (impl + regression test + changeset).

### Correctness — Sound
- Filtering relationship fields first, then mapping to promises inside `Promise.all`, correctly launches every `getRelationshipOptions` call in the same tick instead of gating each on the previous one's resolution.
- The per-field `try`/`catch` is preserved (still wraps both `extractSelectedIds` and `getRelationshipOptions`, same scoping as before), so one field's failure degrades to `[]` without rejecting the whole `Promise.all` or blocking sibling fields.
- `null` results (missing `ref`, or `ref` pointing at a list absent from `config.lists`) are filtered out when reassembling `relationshipData` — matches the original's skip-silently behavior.
- Reassembly keys strictly off `result[0]` (the field name), so `Promise.all` resolution order doesn't affect correctness.

### Test coverage — Strong
- The new test (`fetches relationship options for multiple fields concurrently, not serially`) uses two never-resolving mock `findMany`s and asserts both are invoked before either resolves. This is deterministic (no timers/flakiness) thanks to JS's synchronous-until-first-await semantics — verified it fails against the old serial loop and passes against the fix.
- All pre-existing tests (bounded window, selected-id union for single/many relationships, create-mode no-op) still pass unmodified.

### Minor nit (non-blocking)
`relationshipData[result[0]] = [...result[1]]` spreads into a new array unnecessarily — `getRelationshipOptions` already returns a fresh array per call, so `relationshipData[result[0]] = result[1]` would be equivalent and simpler.

### Risk / security
None — purely a concurrency change to already access-controlled, take-limited, bounded reads. No new query surface.

**Verdict: Approve.** CI is green (test, e2e, coverage, GitGuardian). No blocking issues.


---

PR #828, opened 2026-07-29
fix(ui): render virtual fields read-only in the Admin UI item view

### LINKED ISSUE #821: Admin UI renders "Unsupported field type: virtual" for virtual fields
### Version

`@opensaas/stack-ui` 0.31.1 (`@opensaas/stack-core` 0.32.0)

### What happens

Any list with a `virtual()` field renders that field in the Admin UI as the literal text `Unsupported field type: virtual`, in place of the computed value. Both the list's item/edit view and any other surface going through the default field renderer are affected.

### Expected

The resolved virtual value, displayed read-only.

### Cause

`packages/ui/src/components/fields/registry.ts` — `fieldComponentRegistry` registers ten field types: `text`, `integer`, `checkbox`, `select`, `timestamp`, `password`, `relationship`, `json`, `file`, `image`. There is no entry for `virtual`.

`packages/ui/src/components/fields/FieldRenderer.tsx` then falls through to its "Unsupported field type" placeholder when the registry lookup returns `undefined`.

### Two defects, arguably

1. No component registered for `virtual`.
2. `FieldRenderer` renders virtuals in `mode='edit'` at all. Virtual fields are computed and unwritable, so an editable input for one is never correct — they should be skipped or forced read-only, the way `id` / `createdAt` / `updatedAt` already are earlier in the same component.

Fixing only (1) by registering an editable component would leave a form control bound to a field that cannot be written.

### Repro

1. Add a `virtual()` field to any list
2. Open that list's item view in the Admin UI
3. The field shows the placeholder text

### Workaround

`registerFieldComponent` is exported, so consumers can register a read-only renderer for `virtual` themselves. Per-field `ui.component` overrides also work. Both require every consumer to know about the gap.

borisno2:
## Summary
- Any list with a `virtual()` field rendered that field in the Admin UI item view (create, edit, singleton) as the literal placeholder text `Unsupported field type: virtual`, because the form-field component registry had no entry for `virtual` and `FieldRenderer` fell through to its unsupported-type fallback.
- Adds a default read-only `VirtualField` component, registered for the `virtual` field type, that renders the resolved value with the field's label using the shared field-shell (`FieldRoot`/`FieldLabel`/`FieldReadValue`) presentation.
- `FieldRenderer` now forces read-only presentation for any field config flagged `virtual`, regardless of the requested mode — a virtual field can never become an editable input, the same way `id`/`createdAt`/`updatedAt` are already excluded from editable forms.
- `getEditableFields` (the shared item-form field filter) now omits virtual fields entirely on **create** — there's no item yet, so there's no computed value to show — and keeps them (read-only) on **update**/singleton edit.
- `transformItemFormData` (the submit transform) drops virtual field values defensively before building the create/update payload. Core already strips virtual keys before they reach Prisma (#628), so this is defence-in-depth, not a live-break fix.
- A `null`/`undefined` virtual value renders the shared empty-value affordance (`-`); a non-string value (e.g. `Decimal`, a plain object) is JSON-stringified rather than rendering `[object Object]`.
- Per-field `ui.component` / `ui.fieldType` overrides continue to take priority over the new default, exactly as for every other field type.

Out of scope (per the issue): the Cell registry / list-table rendering of virtual columns (already correct via its plain-text fallback), and any change to core's virtual-field resolution or write-time stripping.

## Test plan
- [x] New tests: `VirtualField.test.tsx` (read-mode rendering, empty-value affordance, non-string value formatting, no editable control)
- [x] New tests: `FieldRenderer.test.tsx` (virtual field renders without the unsupported-type placeholder or `console.warn`, forces read-only even when asked for `mode="edit"`, `ui.component` override still wins)
- [x] New tests: `useItemForm.test.ts` (`getEditableFields` excludes virtual on `create`, keeps it on `update`; `transformItemFormData` drops virtual field values)
- [x] `pnpm exec vitest run` in `packages/ui` — 46 files / 542 tests passing
- [x] `pnpm lint` — no new errors (3 pre-existing unrelated warnings)
- [x] `pnpm manypkg fix` / `pnpm format`
- [x] Changeset added for `@opensaas/stack-ui` (patch)

Closes #821

---
_Generated by [Claude Code](https://claude.ai/code/session_015Y5op6CDtUKUQF9aenGsqC)_

### DIFF
--- a/.changeset/silent-otters-render.md
+++ b/.changeset/silent-otters-render.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-ui': patch
+---
+
+Fix virtual fields rendering "Unsupported field type: virtual" in the Admin UI item view — they now display their resolved value read-only, and are never offered as an editable control or included in create/update payloads.
--- a/packages/ui/src/components/fields/FieldRenderer.tsx
+++ b/packages/ui/src/components/fields/FieldRenderer.tsx
@@ -166,5 +166,11 @@ export function FieldRenderer(props: FieldRendererProps) {
     )
   }
 
-  return <FieldRendererInner {...props} Component={Component} />
+  // A virtual field is computed and never writable — force read-only
+  // presentation regardless of the requested mode (issue #821). This is a
+  // display-only guard: it never skips rendering (unlike id/createdAt/
+  // updatedAt above), it just never offers an editable input.
+  const effectiveMode = fieldConfig.virtual ? 'read' : mode
+
+  return <FieldRendererInner {...props} mode={effectiveMode} Component={Component} />
 }
--- a/packages/ui/src/components/fields/VirtualField.tsx
+++ b/packages/ui/src/components/fields/VirtualField.tsx
@@ -1,0 +1,47 @@
+'use client'
+
+import { FieldRoot, FieldLabel, FieldReadValue } from './field-shell.js'
+
+export interface VirtualFieldProps {
+  name: string
+  value: unknown
+  onChange: (value: unknown) => void
+  label: string
+  mode?: 'read' | 'edit'
+}
+
+/**
+ * Format a resolved virtual value for read-only display.
+ *
+ * A virtual field may declare any `outputType` (e.g. `Decimal`, a plain
+ * object) and crosses the server/client boundary through a JSON round-trip,
+ * so a non-primitive value must be stringified rather than rendered via
+ * `String()` (which would print `[object Object]`).
+ */
+function formatVirtualValue(value: unknown): string {
+  if (value === null || value === undefined) return '-'
+  if (typeof value === 'object') {
+    try {
+      return JSON.stringify(value)
+    } catch {
+      return String(value)
+    }
+  }
+  return String(value)
+}
+
+/**
+ * Default read-only renderer for `virtual` fields (issue #821). Virtual
+ * fields are computed and unwritable, so this component never offers an
+ * editable control — `FieldRenderer` forces `mode: 'read'` for any field
+ * config flagged `virtual`, but the component itself renders read-only
+ * regardless of the `mode` it's given as defence-in-depth.
+ */
+export function VirtualField({ label, value }: VirtualFieldProps) {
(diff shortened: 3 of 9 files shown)


borisno2:
## Self-review

**Overview:** Fixes #821 — a `virtual()` field rendered as the literal placeholder text `Unsupported field type: virtual` in the Admin UI item view because the form-field component registry had no entry for `virtual`. Adds a default read-only `VirtualField` component, forces read-only presentation for any field flagged `virtual` in `FieldRenderer` regardless of requested mode, excludes virtual fields from the create-mode field list (no item exists yet), and drops virtual field values from the submit payload defensively.

**Code quality / style**
- Follows the repo's field-type extensibility pattern (registry entry + component, no `switch` on field type in `FieldRenderer`).
- `VirtualField` reuses the shared field-shell primitives (`FieldRoot`/`FieldLabel`/`FieldReadValue`), consistent with every other field's read-mode presentation.
- `getEditableFields(fields, mode)`'s new `mode` param defaults to `'update'`, so the one pre-existing call site/test with a single argument is unaffected — no signature break for internal callers.

**Verification**
- Confirmed `RelationshipTable.tsx:130` already excludes virtual fields from the inline-editable table-column path, so forcing read-only in `FieldRenderer` is pure defence-in-depth there, not a behavior change.
- Confirmed `getEditableFields`/`useItemForm` aren't part of the package's public export surface (`src/index.ts`), so the signature addition is safe.
- `pnpm exec tsc --noEmit` in `packages/ui` — clean.
- `pnpm exec vitest run` in `packages/ui` — 46 files / 542 tests passing (including the new `VirtualField.test.tsx`, the new `FieldRenderer` virtual-field tests, and the new `useItemForm` tests for the create/update filtering and submit-payload exclusion).
- `pnpm lint` — no new errors (3 pre-existing, unrelated warnings).

**Risks:** None identified. The change is additive (new registry entry, new component) plus a narrowly-scoped mode-forcing guard in `FieldRenderer`; existing field types and forms are unaffected.

No changes requested.


---

Issue #844, opened 2026-07-31
0.33.0: reads inside `resolveOutput` recurse without bound — server OOMs on any access-scoped read


borisno2:
# 0.33.0: reads inside `resolveOutput` recurse without bound — server OOMs on any access-scoped read

**Regression in 0.33.0** (works on 0.32.0). Introduced by #838 / ADR-0022 (fix for #830).

Upgrading `@opensaas/stack-core` 0.32.0 → 0.33.0 makes every authenticated page in our app hang and drives the Next.js server to the V8 heap limit. In CI our Playwright server died ~3 minutes into the run:

```
[2815:0x1ee51000] 185775 ms: Mark-Compact 4073.9 (4105.2) -> 4062.2 (4106.0) MB, pooled: 0 MB,
                            769.07 / 0.01 ms (average mu = 0.304, current mu = 0.237) task;
                            scavenge might not succeed
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x74eae8 node::OOMErrorHandler(char const*, v8::OOMDetails const&) [next-server (v16.2.10)]
```

59 of 71 e2e tests failed, almost all as `ERR_CONNECTION_REFUSED` after the process was gone.

## Cause

Two changes in #838 interact:

**1. `access/access-filter.ts` — `buildIncludeWithAccessControl`**

0.32 bailed out entirely inside a `resolveOutput` context:

```ts
if (args.context._resolveOutputCounter.depth > 0) {
  return undefined   // no auto-include: the read fetched scalar columns only
}
```

0.33 replaced that with a flag that still emits a one-level include for every readable relation ([`access-filter.ts:126`](https://github.com/OpenSaasAU/stack/blob/main/packages/core/src/access/access-filter.ts#L126)):

```ts
const insideResolveOutput = args.context._resolveOutputCounter.depth > 0
// …
if (!insideResolveOutput && !visitedLists.includes(relatedListName)) {
  nested = await buildIncludeWithAccessControl(/* … */)
}
include[fieldName] = { where, nested }   // ← relation is included either way
```

**2. `access/field-visibility.ts` — `filterReadableFields`**

The `depth < MAX_DEPTH` guard on the relationship recursion was removed, so every returned relation row now runs *its own* `resolveOutput` hooks at any depth.

### Why the guards don't hold

[...]

Issue #830, opened 2026-07-30
Access control fails OPEN past include depth 5: caller includes pass through unscoped and nested rows skip field-read filtering


borisno2:
## Summary

Both halves of the read-side access-control pipeline stop at `MAX_DEPTH = 5`, and both **fail open** rather than closed. Past that depth a caller-supplied `include` is neither row-scoped nor field-filtered, so a relation returns every row of the related list with read-denied columns intact.

The two cut-offs are independent but line up, so they compound:

**1. Row scoping — `packages/core/src/access/access-filter.ts`**

`buildIncludeWithAccessControl` returns `undefined` at `depth >= MAX_DEPTH`. `mergeIncludeWithAccessControl` treats whole-object `undefined` as "nothing to merge against" and returns the caller's include unchanged. The behaviour is deliberate and documented in the source:

> The access-controlled include is recursive to `MAX_DEPTH` (see `buildIncludeWithAccessControl`); beyond that depth no auto-include exists, so **deeper caller selections pass through unscoped** — consistent with the existing auto-include behaviour.

The distinction the code draws — a whole-object `undefined` is a passthrough, an `undefined` *entry* is a denial — is reasonable for the `resolveOutput`/virtual-field and no-relationships cases it also covers. The `MAX_DEPTH` case is the one where passthrough is the wrong default, because it is the only one where a relation genuinely *should* have been scoped and wasn't.

**2. Field visibility — `packages/core/src/access/field-visibility.ts`**

`filterReadableFields` recurses only while `depth < MAX_DEPTH`. Past that, a relationship value falls through to `resolveReadableFieldValue`, which checks the read gate on the *relationship field itself* and then returns the nested object **as-is**. The related row's own per-field `access.read` gates are never evaluated, and its `resolveOutput` hooks never run.

## Why this matters

The failure is **silent and invisible**. There is no warning, no thrown error, and nothing in the returned shape that distinguishes "this relation was access-filtered" from "this relation was returned raw because it sat one level too deep". A reviewer reading the call site sees an ordinary `include` and reasonably assumes the engine scoped it — the framework's central promise.

It is also **discontinuous**: the same relation on the same list is filtered at depth 4 and unfiltered at depth 5. Adding one intermediate hop to an existing include — a refactor that looks purely additive — can silently switch off access control on the leaf. Nothing in the type system or at runtime flags it.

The consequences depend on what sits at the leaf:

[...]

PR #853, opened 2026-07-31
fix(core): a bare read fetches scalars, not relations (#848)

### LINKED ISSUE #848: Reads with no `include` auto-include every relation to depth 5 — Prisma returns scalars only, and this default is the root cause behind #566 / #830 / #844
## Summary

A read with **no `include` key** auto-includes *every* readable relationship of the list, recursing to `READ_INCLUDE_MAX_DEPTH` (5). Prisma's own semantics for the same call are "scalars only". The only way to get Prisma's behaviour through `context.db.*` is to pass an explicitly-empty `include: {}`, which is backwards from what every Prisma user expects.

I think this default is the shared root cause behind a run of already-fixed symptoms (#566, #830, 0.27.1's stack overflow, #844 / 0.34.0's cycle guard), and it just cost us a full day on the 0.33 → 0.34 upgrade.

## The code

`packages/core/src/context/index.ts` (`createFindUnique` ~L1010, `createFindMany` ~L1148):

```ts
include = args?.include
  ? mergeIncludeWithAccessControl(args.include, accessControlledInclude, listConfig.fields, config, listName)
  : toPrismaInclude(accessControlledInclude)   // ← no caller include ⇒ auto-include everything
```

`buildIncludeWithAccessControl` (`packages/core/src/access/access-filter.ts` L101) walks every declared relationship and recurses to depth 5.

So this:

```ts
await context.db.user.findUnique({ where: { id } })
```

…reads like "give me this row", but actually fetches a five-level-deep relation tree, access-filters every node in it, and runs every nested row's `resolveOutput` hooks.

## What it cost us on 0.34.0

0.34.0's `ResolveOutputCycleError` (#844) is a correct guard, but it converted a latent OOM into a hard 500 on **six** separate virtual fields in our app — all of the same shape, none of which had any interest in the relations being auto-included:

```ts
// User.name
name: virtual({
  type: 'string',
  hooks: {
    resolveOutput: async ({ item, context }) => {
      const [account] = await context.db.account.findMany({ where: { userId: item.id }, take: 1 })
      return `${account?.firstName} ${account?.surname} - ${item.provider}`
    },
  },
}),
```

`db.account` auto-includes `Account.user` → the nested `User` row is field-filtered → `User.name`'s hook runs again → `ResolveOutputCycleError: User.name → User.name`. Every page reading a `User` through a non-sudo context 500s.

Same loop, five more times: `LessonTerm.name` (via `db.lesson` / `db.term`), `Enrolment.name` (via `Student.enrolments`), `LessonSchedule.name` and `LessonAttendance.name` (via `Enrolment.lessonSchedules` / `.lessonAttendances`), `Bill.total` / `Bill.amountOwing` (via `BillItem.bill`).

The resolver only ever reads scalars off the row it fetched. The relation tree is pure cost, and the cost is not small — one of our seed writes exceeded Prisma's 5s interactive-transaction budget because a hook-issued `findMany` was dragging that tree along.

Our fix was to pass `include: {}` at all six sites. It works, but "pass an empty include to get Prisma's default" is a workaround, not an API.

## Why I think it's the root cause, not six unrelated bugs

Each of these was fixed by bounding a *consequence* of auto-include rather than the default itself:

- **#566** — a caller `include` REPLACED the access-controlled include, dropping per-relation access `where`s. Only reachable because there is an auto-include to be replaced.
- **#830** — access failed OPEN past include depth 5. Only reachable because a depth cap is needed at all.
- **0.27.1** — stack overflow re-descending a cyclic readable-relationship graph. Fixed by pruning at cycle back-edges in the auto-include walk.
- **#844 / 0.34.0** — unbounded recursion when a `resolveOutput` hook issues its own read. The loop runs through the hook↔read boundary precisely *because* the hook's read auto-includes the back-relation.

`buildIncludeWithAccessControl`'s own comment concedes the shape of the problem:

> A self-referential relation read from inside a hook is a NEW top-level read, invisible to this function's own recursion — the loop runs through the hook↔read boundary, not through repeated calls to `buildIncludeWithAccessControl`.

If a bare read fetched scalars only, there is no back-relation for the hook's read to pull in, and the cycle cannot form in the first place.

## Suggested direction

**Make a read with no `include` return the row's own columns (plus its virtuals), matching Prisma.** Relations come back only when the caller asks for them — at which point the existing `mergeIncludeWithAccessControl` path already applies access filtering correctly, and all the depth/cycle machinery only has to cover explicitly-requested trees.

I recognise the convenience the current default buys: admin UI label rendering gets `item.relation` without the caller knowing the relation graph. If that's the blocker, some options that don't tax every read in the app:

1. Opt **in** per read — `context.db.user.findUnique({ where, autoInclude: true })`.
2. Opt in per list — `ui.autoInclude` / an admin-surface-only read primitive, so it's scoped to the surface that actually needs it (in the spirit of #630's relationship-options primitive).
3. Keep the default but disable auto-include inside `resolveOutput` (`_resolveOutputChain.length > 0`) — narrowest fix, kills the whole cycle class outright, still leaves bare reads over-fetching everywhere else.

Happy to do any of these as a PR if you have a preference on which shape you'd take.

## Environment

`@opensaas/stack-core` 0.34.0, `@prisma/client` 7.9.1, Next.js App Router, Postgres (and PGlite in test).

borisno2:
## Summary

- A `context.db` read with no `include` (and no fragment `query`) used to auto-include **every** readable relationship of the list, recursing up to `READ_INCLUDE_MAX_DEPTH` (5) and evaluating every related list's operation-level `query` access along the way — Prisma's own semantics for the same call are "scalars only." This is identified as the shared root cause behind #566, #830, and #844.
- A bare read now returns the row's own columns plus its virtual fields only, matching Prisma, for `findUnique`, `findMany`, and a singleton's `.get()` alike — under sudo and under a session, uniformly. `.get()` also gains caller-`include` support it never had.
- Relations are still fetched, and access-scoped exactly as before, whenever a caller names them via `include` or a fragment `query` — the `mergeIncludeWithAccessControl` / `buildIncludeWithAccessControl` machinery (and the #566/#830 fixes it carries) is untouched; it's simply no longer invoked unprompted.
- Follows the decision recorded in ADR-0024 (`docs/adr/0024-a-read-with-no-include-fetches-scalars-not-relations.md`, #849, already merged).

### Implementation

- `packages/core/src/context/index.ts`: `createFindUnique`, `createFindMany`, and `createGet` (singleton) only call `buildIncludeWithAccessControl` when the caller supplied an `include`; a bare read sends `include: undefined` straight through. `createGet` gains the same fragment/`include`/sudo handling the other two reads already had.
- `packages/cli/src/generator/types.ts`: generates a `<List>GetArgs` type for singleton lists and types `get()` to accept it (`include`/`query`/`select`), mirroring `FindUniqueArgs`.
- Docs (`docs/content/concepts/{queries,access-control}.md`, `docs/content/reference/context-api.md`, `packages/core/CLAUDE.md`) updated to describe the new default; one example (`examples/blog/test-context-nested-virtual.ts`) fixed to pass an explicit `include` instead of relying on the old auto-include.
- Existing regression coverage for #566 and #830 (caller-supplied-`include` behavior) is **unmodified** — only bare-read call sites in tests were updated. The `#844` cycle-guard reproduction now demonstrates the cycle no longer occurs on the reporter's bare hook-issued read (with the mock's return value made to actually respect the `include` argument, so it faithfully models Prisma); the genuine two-list cycle regression is untouched and still throws.
- New test coverage: `packages/core/tests/bare-read-scalars.test.ts` (bare vs. explicit-include behavior, including "no related `query` access is evaluated on a bare read") and a new describe block in `packages/core/tests/singleton.test.ts` for `.get()`'s new `include` support.

## Test plan

- [x] `pnpm test` in `packages/core` — 923/923 passing
- [x] `pnpm test` in `packages/cli` — 323/323 passing
- [x] `pnpm build` for `@opensaas/stack-core` and `@opensaas/stack-cli` — clean
- [x] `pnpm lint` — clean (pre-existing unrelated warnings only)
- [x] `pnpm manypkg fix` — no changes
- [x] `pnpm format` — clean
- [x] Generated `examples/blog` types and typechecked the fixed example script against them
- [x] Changeset added (`.changeset/silent-relations-detour.md`, minor) leading with the silent nature of the break, both detection strategies, and the migration

Closes #848

https://claude.ai/code/session_01PifFTifspW7ud4ptUGK4Mx

---
_Generated by [Claude Code](https://claude.ai/code/session_01234uq5m5nrBAB6cQa95m9M)_

### DIFF
--- a/.changeset/silent-relations-detour.md
+++ b/.changeset/silent-relations-detour.md
@@ -1,0 +1,47 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
+---
+
+**This break is silent.** A `context.db` read with no `include` (and no fragment `query`) used to auto-include every readable relationship of the list, recursing up to 5 levels deep. It now returns the row's own columns plus its virtual fields only — matching Prisma's own semantics for a bare read — and relations arrive only when you name them. A read that used to return `post.author` now returns no `author` key at all: no error, no warning, just less data (ADR-0024). This applies uniformly to `findUnique`, `findMany`, and a singleton's `get()`, under sudo and under a session alike. Foreign-key columns (e.g. `authorId`) are unaffected and always returned, so a relation stays reachable by id without an `include`.
+
+**Detect call sites that need updating:**
+
+- Grep for bare reads: `context.db.*.find*` calls (or a singleton's `.get()`) with no `include` and no `query` argument, whose result is later used to access a relationship field.
+- Grep for `resolveOutput` hooks on `virtual` fields that read a relation off `item` (e.g. `item.author`, `item.posts`) — these silently degrade the same way, since a hook's own `context.db` read is subject to the same rule.
+
+**Migrate** by naming the relation explicitly, either via `include`:
+
+```typescript
+// Before — relied on the auto-include
+const post = await context.db.post.findUnique({ where: { id } })
+post.author // used to be populated
+
+// After — name it
+const post = await context.db.post.findUnique({
+  where: { id },
+  include: { author: true },
+})
+post.author // populated
+```
+
+or via a fragment `query`:
+
+```typescript
+const post = await context.db.post.findUnique({
+  where: { id },
+  query: postWithAuthorFragment,
+})
+```
+
+A `resolveOutput` hook that read `item.<relation>` should instead read through `context.db` with an explicit `include`, or its caller should pass one.
+
+**Singleton `get()` gains caller-`include` support** it never had — it can now be narrowed and widened like any other read:
+
+```typescript
+const settings = await context.db.settings.get({ include: { homepage: true } })
+```
+
+Bare reads also stop evaluating operation-level `query` access on related lists (that walk previously ran for every relation at every level before fetching anything), so an access function relied on for a side effect will no longer fire on a bare read.
+
+See `docs/adr/0024-a-read-with-no-include-fetches-scalars-not-relations.md` for the full rationale.
--- a/docs/content/concepts/access-control.md
+++ b/docs/content/concepts/access-control.md
@@ -23,7 +23,7 @@ Two defaults set the tone:
 
 Writes check operation-level access, filter writable fields, then persist. Reads are a **two-phase** pipeline:
 
-1. **Access Filter** (pre-query): the engine evaluates operation-level `query` access and merges the resulting filter into the Prisma `where`/`include` — rows and relations a session can't see never leave the database.
+1. **Access Filter** (pre-query): the engine evaluates operation-level `query` access and merges the resulting filter into the Prisma `where`/`include` — rows and relations a session can't see never leave the database. This only runs for relations the caller actually asked for: a read with no `include` (and no fragment `query`) fetches the row's own columns and virtual fields only, matching Prisma's own semantics — see [Queries & Fragments](/docs/concepts/queries).
 2. **Field Visibility** (post-query): on the returned rows, fields the session can't read are removed, `resolveOutput` hooks run, and virtual fields are computed.
 
 In order:
--- a/docs/content/concepts/queries.md
+++ b/docs/content/concepts/queries.md
@@ -97,6 +97,10 @@ if (!post) return notFound()
 `context.db` reads do **not** honour Prisma's `select` argument. Narrow a read with `include` (for relationships) or a fragment `query` instead. Passing `select` to `findUnique`/`findMany` is a no-op: it logs a runtime warning and the full, access-filtered record is still returned (field-level visibility is always enforced by [access control](/docs/concepts/access-control), regardless of `select`).
 {% /callout %}
 
+{% callout type="info" %}
+A `context.db` read with **no** `include` and **no** fragment `query` returns the row's own columns and virtual fields only — never relations, matching Prisma's own default. `post.author` is `undefined` on a bare `findUnique`/`findMany` unless you name it via `include` or a fragment. This also applies to a `virtual` field's `resolveOutput` hook issuing its own bare `context.db` read — it won't see relations either, so read through `context.db` for what you need or pass an `include`.
+{% /callout %}
+
 ### Via `runQuery` / `runQueryOne` (standalone helpers)
 
 When you don't have direct access to `context.db` (for example inside a hook or a shared utility), use the standalone helpers. They take the `context`, the PascalCase list key, the fragment, and query args:
--- a/docs/content/reference/context-api.md
+++ b/docs/content/reference/context-api.md
@@ -316,6 +316,10 @@ export async function getContext(session: Session = null) {
 
 All database operations are access-controlled and execute hooks in the correct order.
 
+{% callout type="warning" %}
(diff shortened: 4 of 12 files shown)


borisno2:
## Self-review

### Overview
Implements #848 / ADR-0024: a `context.db` read with no `include` (and no fragment `query`) now returns the row's own scalar columns plus its virtual fields only — it no longer auto-includes every readable relationship recursively. Applies uniformly to `findUnique`, `findMany`, and a singleton's `.get()` (which also gains caller-`include` support it never had), under sudo and under a session alike.

### Correctness
- The caller-`include` merge path (`buildIncludeWithAccessControl` + `mergeIncludeWithAccessControl`, carrying the #566/#830 fixes) is untouched — it is simply gated behind `args.include` being truthy instead of running unconditionally. `findFirst` is sugar over `findMany` and inherits the new behavior automatically; no separate change was needed there.
- Verified the two hot spots called out in the issue: (1) related-list `query` access is no longer evaluated on a bare read — covered by a spy assertion in `bare-read-scalars.test.ts`; (2) the reporter's `User → Account → Student` cycle no longer reproduces on the hook's bare `context.db.account.findMany(...)` call — the test's mock was changed to condition its return value on the actual `include` argument (mirroring real Prisma) rather than always returning the full relation tree, since the old mock would have masked the fix.
- `stripVirtualFieldsFromInclude(undefined, ...)` and `mergeIncludeWithAccessControl` with an empty caller `include: {}` were checked against existing behavior — both are unchanged from before this PR (an empty object was already truthy pre-PR, so `include: {}` already took the merge branch; that's pre-existing, not a regression here).

### Test coverage
- New: `packages/core/tests/bare-read-scalars.test.ts` (bare vs. explicit include, FK columns + virtuals still returned, related query access not evaluated, sudo parity).
- New: singleton `.get()` include-support tests in `singleton.test.ts`.
- Updated in place (not touched further than necessary): 3 bare-read assertions in `nested-access-and-hooks.test.ts` now pass an explicit `include` to keep testing what they were designed to test; the `#844` cycle reproduction and the concurrent-isolation test in `resolve-chain.test.ts` were adapted to the new default (see above).
- Left deliberately unmodified per the issue's acceptance criteria: the #566/#830 regression suites in `context.test.ts` and `access-relationships.test.ts`, and the genuine two-list cycle regression in `resolve-chain.test.ts` (still throws `ResolveOutputCycleError`, proving the cycle guard isn't weakened).

### Risks / things I considered and ruled out
- **Silent behavior change**: this is the core, accepted risk of the change (a read that returned `post.author` now returns no `author` key, no error) — called out explicitly as the lead of the changeset per ADR-0024, with both detection strategies and a migration example.
- **Generator surface**: `packages/cli`'s new `<List>GetArgs` type is only generated for `isSingleton` lists, and `get()`'s new `args` parameter is optional, so existing generated call sites (`get()` with no args) keep compiling and behaving the same.
- **MCP**: the `query` CRUD tool already issues a bare `findMany` with no `include` — it inherits the new default with no code change, matching the issue's explicit instruction not to add an `include` there.

### Verification run
- `packages/core`: 923/923 tests passing
- `packages/cli`: 323/323 tests passing
- `tsc` clean for both packages
- `pnpm lint` clean (pre-existing unrelated warnings only), `pnpm manypkg fix` no-op, `pnpm format` clean
- Regenerated `examples/blog` and typechecked the fixed example script against the real generated types

No further changes identified as needed.

---

PR #867, opened 2026-08-02
fix(auth): base-model plugin schema extension no longer clobbers derived Auth list db/access

### LINKED ISSUE #861: auth: a better-auth plugin extending a base model clobbers the derived Auth list's db + access config

## Summary

When a better-auth plugin contributes a `schema` extension for one of the four **base** models (`user`/`session`/`account`/`verification`), `authPlugin`'s `init` adds that extension as a **brand-new list** *before* the real derived Auth lists are added. The derived Auth list then loses to `extendList`, which merges only `fields`/`hooks`/`mcp` — so the derived list's `db` (`map`, `schema`, `timestamps`) and `access` are **silently dropped**.

For an adopting app (`adoptBetterAuthTables({ schema: 'auth' })`) this moves the auth user table out of its Postgres schema and drops two columns — a destructive migration produced purely by adding a plugin.

## Reproduction

Config: `adoptBetterAuthTables({ schema: 'auth' })` (so `user.modelName: 'AuthUser'`, `schema: 'auth'`), plus better-auth's `anonymous()` plugin — which declares `schema: { user: { fields: { isAnonymous: … } } }` — passed through `betterAuthPlugins`.

`opensaas generate`, with the plugin **absent**:

```prisma
model AuthUser {
  id            String   @id @default(cuid())
  name          String
  email         String   @unique
  emailVerified Boolean  @default(false)
  image         String?
  sessions      AuthSession[]
  accounts      AuthAccount[]
  createdAt     DateTime @default(now())
  updatedAt     DateTime @default(now()) @updatedAt

  @@map("AuthUser")
  @@schema("auth")
}
```

Same config with the plugin **present** — the only change:

```prisma
model AuthUser {
  id            String   @id @default(cuid())
  isAnonymous   Boolean  @default(false)
  name          String
  email         String   @unique
  emailVerified Boolean  @default(false)
  image         String?
  sessions      AuthSession[]
  accounts      AuthAccount[]

  @@schema("public")
}
```

Diffed the two generator outputs directly; the delta is exactly `+isAnonymous`, `-createdAt`, `-updatedAt`, `-@@map("AuthUser")`, `@@schema("auth")` → `@@schema("public")`.

`prisma migrate diff` against the live database therefore wants to **move `auth.user` to `public.AuthUser` and drop its `createdAt`/`updatedAt` columns**.

## Cause

`packages/auth/src/config/plugin.ts` (shipped `dist/config/plugin.js`), inside `init`:

1. The `betterAuthPlugins` loop runs **first**. `convertBetterAuthSchema(pluginSchema, baseModelKeys)` correctly resolves the extension's list key through the `modelName` remap (`user` → `AuthUser`) — that part works. But at this point `context.config.lists.AuthUser` does not exist yet, so it takes the `else` branch and calls `addList('AuthUser', listConfig)`. `convertTableToList` returns a bare `list({ fields })`: **no `db`, no `access`**.
2. The `authLists` loop runs **second**. `context.config.lists.AuthUser` now exists, so it takes the `extendList` branch — which merges `fields`/`hooks`/`mcp` only. The derived list's `db: { timestamps: true, map: 'AuthUser', schema: 'auth' }` and its `access` are never applied.

The `extendList` branch is correct for its intended case (the app declared `User` itself, and access/db belong to the app — ADR-0013). The bug is that a *plugin-contributed field extension* is being treated as list ownership.

Secondary effect: the `authPlugin({ access })` entry for that model is lost too. In our case `access.user` (`query: isAdmin`) silently became fully closed (stack-core is default-closed, so this fails safe rather than open — but it is still a silent functional regression, and it would be an *open* regression for anyone whose derived list carried a more permissive posture).

## Expected

A plugin's schema extension of a base model should merge into the derived Auth list, not pre-empt it. The derived list's `db` and `access` must survive.

## Suggested fix

Add the derived `authLists` **before** the `betterAuthPlugins` schema loop, so a base-model extension always hits the `extendList` path against the real derived list. Non-base plugin tables (`oauth_application`, `passkey`, …) still `addList` as they do today.

If ordering can't be swapped for another reason, the alternative is to make the plugin loop skip base-model keys and collect their fields into the derivation input instead.

Either way, a regression test asserting that `@@map`/`@@schema`/`timestamps` on a derived Auth list are unchanged by a base-model-extending plugin would lock this down.

## Workaround in use

None that is clean — an app can only re-assert `db`/`access` from a follow-on plugin that mutates `config.lists.AuthUser` in place (`extendList` cannot set `db` at all).

borisno2:
## Summary

- `authPlugin`'s `init` processed `betterAuthPlugins` schema extensions **before** registering the derived Auth lists. When a better-auth plugin declared a schema extension for a base model (`user`/`session`/`account`/`verification`), it resolved to the derived list key (e.g. `AuthUser`) and pre-empted it with a bare `list({ fields })` via `addList` — no `db`, no `access`. The real derived list, added moments later, then lost to the field-only `extendList` merge, silently dropping its `@@map`/`@@schema`/`timestamps` and `access` config.
- For an app adopting a separate-schema better-auth install (`adoptBetterAuthTables({ schema: 'auth' })`), this relocated the auth user table out of its Postgres schema and dropped the timestamp columns — a destructive migration produced purely by adding a plugin like `anonymous()`.
- Fixed by registering the derived Auth lists **before** processing the `betterAuthPlugins` schema loop, so a base-model extension always finds its key already occupied by the real derived list and takes the `extendList` (merge) path against it. Non-base plugin tables (`oauth_application`, `passkey`, …) still register via `addList` exactly as before.

## Test plan

- [x] Added a regression test (`packages/auth/tests/plugin-derived-keys.test.ts`) asserting that a base-model-extending better-auth plugin leaves the derived Auth list's `db.map`/`db.schema`/`db.timestamps` and `access` unchanged, while still merging in the plugin's fields
- [x] Verified the new test fails against the pre-fix code and passes after the fix (TDD red/green)
- [x] `pnpm --filter @opensaas/stack-auth test` — 147 tests pass
- [x] `pnpm --filter @opensaas/stack-auth build` — type-checks clean
- [x] `pnpm lint` — no new errors
- [x] Changeset added (`@opensaas/stack-auth`, patch)

Closes #861

---
_Generated by [Claude Code](https://claude.ai/code/session_01CVFf2Y9j9eByjLyo2aiVGT)_

### DIFF
--- a/.changeset/tame-otters-listen.md
+++ b/.changeset/tame-otters-listen.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-auth': patch
+---
+
+Fix a better-auth plugin's schema extension of a base model (`user`/`session`/`account`/`verification`) silently dropping the derived Auth list's `db` (`map`/`schema`/`timestamps`) and `access` config.
--- a/packages/auth/src/config/plugin.ts
+++ b/packages/auth/src/config/plugin.ts
@@ -62,7 +62,45 @@ export function authPlugin(config: AuthConfig): Plugin {
         verification: normalized.models.verification.modelName,
       }
 
-      // Extract additional lists from Better Auth plugins
+      // Add all auth lists FIRST, before any better-auth plugin schema
+      // extension is processed. This must happen before the betterAuthPlugins
+      // loop below so a base-model extension (e.g. a plugin's `schema: {
+      // user: { fields: … } }`) always finds the real derived list already
+      // registered under its key and takes the merge (`extendList`) path
+      // against it — instead of pre-empting that key with a bare
+      // `list({ fields })` that carries no `db`/`access` (see #861).
+      //
+      // The plugin only ever touches its OWN derived keys. When a developer
+      // renames the auth user model (e.g. user.modelName: 'AuthUser'), the
+      // derived key is 'AuthUser' and an app's separate 'User' list is left
+      // untouched — the plugin never extends/overwrites a list it didn't
+      // derive. Extending only kicks in when an existing list shares the
+      // derived key (e.g. the default 'User'), which is the intended
+      // "merge auth fields into my User" behaviour.
+      for (const [listName, listConfig] of Object.entries(authLists)) {
+        if (context.config.lists[listName]) {
+          // A list already exists under this derived key — merge auth fields
+          // in only. Access control belongs to whoever owns the list (the
+          // application declared it first), so the plugin never forwards its
+          // own access here — see ADR-0013.
+          context.extendList(listName, {
+            fields: listConfig.fields,
+            hooks: listConfig.hooks,
+            mcp: listConfig.mcp,
+          })
+        } else {
+          // Otherwise, add the auth list
+          context.addList(listName, listConfig)
+        }
+      }
+
+      // Extract additional lists from Better Auth plugins. Because the auth
+      // lists above are already registered, a schema extension of a base
+      // model (user/session/account/verification) always finds its resolved
+      // key occupied by the real derived list and merges via `extendList` —
+      // its `db`/`access` are preserved. Non-base plugin tables (e.g.
+      // `oauth_application`, `passkey`) still register as new lists via
+      // `addList`, same as before.
       for (const plugin of normalized.betterAuthPlugins) {
         if (plugin && typeof plugin === 'object' && 'schema' in plugin) {
           // Plugin has schema property - convert to OpenSaaS lists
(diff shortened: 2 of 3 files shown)


borisno2:
## Review

**Overview:** Fixes #861 — `authPlugin`'s `init` registered a better-auth plugin's schema extension of a base model (`user`/`session`/`account`/`verification`) *before* the derived Auth lists existed, so the extension pre-empted the list key with a bare `list({ fields })` (no `db`, no `access`). The real derived list then hit the field-only `extendList` merge path and silently lost `@@map`/`@@schema`/`timestamps`/`access`. The fix swaps the two loops so the derived Auth lists are always registered first.

**Correctness:**
- The reorder is exactly the minimal change needed: the `authLists` loop now runs before the `betterAuthPlugins` schema loop, so a base-model extension's `context.config.lists[listName]` check always finds the real derived list and takes the `extendList` (merge-only) branch — preserving `db`/`access`.
- Non-base plugin tables (`oauth_application`, `passkey`, …) are unaffected — they still resolve to keys the `authLists` loop never touches, so they still hit `addList` on first pass.
- Verified by tracing both pre-existing scenarios: (1) an app declares its own list under a derived key — unaffected, since the `authLists` loop's own `extendList`-vs-`addList` branch behavior for that key is unchanged, just runs earlier; (2) two plugins extending the same base model — unaffected, unrelated to ordering.
- No change to `ADR-0013` semantics: access is still never forwarded on any `extendList` call.

**Test coverage:**
- The new regression test in `plugin-derived-keys.test.ts` is well targeted — it reproduces the exact issue shape (`adoptBetterAuthTables`-style config + a base-model-extending better-auth plugin) and asserts `db.map`/`db.schema`/`db.timestamps` and `access.operation.query` all survive, while the plugin's field (`isAnonymous`) still merges in. Confirmed locally this test fails against the pre-fix code and passes after.
- Existing test suite (147 tests) and build both pass; `pnpm lint`/`format`/`manypkg fix` produced no changes.

**Style/process:**
- Comments were relocated and updated to reflect the new order/rationale rather than left stale — good.
- Changeset added at the correct `patch` level for `@opensaas/stack-auth`.

**Risk:** Low. This is a pure reordering of two independent loops within `init`; no new code paths, no API surface change. I don't see any remaining issues.

No changes requested.

---

PR #870, opened 2026-08-02
fix(auth): derived expiresAt and the session/account user FK are required

### LINKED ISSUE #863: auth: derived expiresAt and the session/account user FK are nullable; better-auth's own schema requires them

## Summary

Three columns on the derived Auth lists are emitted as **nullable**, but better-auth's own Prisma schema declares them **required**:

| Derived list | Column | Generated | better-auth's own schema |
| --- | --- | --- | --- |
| `AuthSession` | `expiresAt` | `DateTime?` | `DateTime` |
| `AuthVerification` | `expiresAt` | `DateTime?` | `DateTime` |
| `AuthSession` | `userId` + `user` relation | `String?` / `AuthUser?` | `String` / `User` |
| `AuthAccount` | `userId` + `user` relation | `String?` / `AuthUser?` | `String` / `User` |

There is no config knob to correct any of them, so a generated Auth schema can never match a live better-auth database on these columns.

## Detail

`packages/auth/src/config/derive-auth-lists.ts`:

```ts
expiresAt: timestamp({ db: fieldDb('expiresAt', f) }),        // no validation.isRequired / db.isNullable
user: relationship({ ref: `${keys.user}.sessions`, isIndexed: false, db: userRelationshipDb(f) }),
```

`fieldDb` only forwards a column `@map`, and `AuthModelConfig.fields` is `Record<string, string>` (column names), so there's no seam to pin nullability from the outside either.

Consequences:

- **Adopters** get a `prisma migrate diff` that wants to `DROP NOT NULL` on four columns — not data-destructive, but it silently relaxes referential integrity on the session/account FKs and makes an expired-session sweep (`where: { expiresAt: { lt: now } }`) unable to rely on the column being present.
- **Greenfield** users get a schema that diverges from what better-auth itself would have generated, so a later switch to (or comparison against) better-auth's own migrations is a mismatch.
- A `session` row genuinely cannot exist without a `userId` or an `expiresAt` — better-auth's adapter always writes both — so nullable is wrong on the merits, not just for parity.

## Proposed

Make the derivation match better-auth's schema:

```ts
expiresAt: timestamp({ validation: { isRequired: true }, db: { isNullable: false, ...fieldDb('expiresAt', f) } }),
user: relationship({ ref: `${keys.user}.sessions`, isIndexed: false, db: { ...userRelationshipDb(f), isNullable: false } }),
```

Note this would be a **schema change for existing greenfield users** (their generated columns are currently nullable), so it likely wants a minor-version note — or, if that's too sharp, a `strictNullability`-style opt-in on `AuthConfig` that adopters can turn on.

## Related

- The `updatedAt` columns on the derived `session`/`account` lists come out as `DateTime @default(now()) @updatedAt`, whereas better-auth's own schema has `@updatedAt` with no `now()` default. Non-destructive (`SET DEFAULT`), so filing it here as a note rather than separately — but it's the same "derived shape ≠ better-auth's shape" family.

borisno2:
## Summary

- Core: relationship fields gain `db.isNullable` on the FK-owning (single) side — when set to `false`, the generated FK column and its relation field lose their `?` together. Omitting it leaves every existing relationship byte-for-byte unchanged (still nullable by default). Misusing it on a `many: true` field throws at field-definition time, mirroring the existing `db.foreignKey` validation.
- Auth: the derived Auth lists now use `db.isNullable: false` so `Session.expiresAt`, `Verification.expiresAt`, and the `Session.user`/`Account.user` foreign keys generate as **required**, matching better-auth's own Prisma schema. Previously these four columns were nullable with no config seam to correct them, so a generated Auth schema could never match a live better-auth database on these columns.
- Regenerated `examples/starter-auth/prisma/schema.prisma` to reflect the new required columns.
- Docs: documented the new `db.isNullable` relationship option in `docs/content/reference/fields-api.md`.
- Added a changeset (minor on both `@opensaas/stack-core` and `@opensaas/stack-auth`) with a migration note — existing greenfield adopters will see a migration that adds `NOT NULL` to these four columns after regenerating.

## Test plan

- [x] Unit tests: `relationship()` constructor accepts `db.isNullable` on a single relationship and throws when combined with `many: true` (`packages/core/tests/field-types.test.ts`)
- [x] Generator tests: FK column/relation field emit without `?` when `db.isNullable: false`, for both bidirectional and list-only refs; default (omitted) behavior unchanged (`packages/cli/src/generator/prisma.test.ts`)
- [x] Auth unit tests: derived `Session`/`Account`/`Verification` fields carry `db.isNullable: false` (`packages/auth/tests/derive-auth-lists.test.ts`)
- [x] Auth end-to-end schema tests: the fully assembled generated schema emits required `userId`/`user` and `expiresAt` columns (`packages/auth/tests/generated-fk-shape.test.ts`)
- [x] `pnpm build`, `pnpm -w exec turbo run test` (core, cli, auth, ui) all pass
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` — no changes needed beyond what's committed
- [x] Docs link-check passes

Closes #863

---
_Generated by [Claude Code](https://claude.ai/code/session_01QD1mhmPMEZFxWWjpRToN2C)_

### DIFF
--- a/.changeset/tiny-plums-relax.md
+++ b/.changeset/tiny-plums-relax.md
@@ -1,0 +1,19 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-auth': minor
+---
+
+A relationship field's foreign key can now be declared non-nullable via `db.isNullable: false` — the generated FK column and its relation field lose their `?` together. Omitting the option leaves every existing relationship unchanged (still nullable by default).
+
+```typescript
+user: relationship({
+  ref: 'User.sessions',
+  db: { isNullable: false },
+})
+// Generates: userId String  (was String?)
+//            user   User    @relation(...)  (was User?)
+```
+
+`@opensaas/stack-auth`'s derived Auth lists now use this to match better-auth's own Prisma schema: `Session.expiresAt`, `Verification.expiresAt`, and the `Session.user`/`Account.user` foreign keys generate as required instead of nullable.
+
+**Migration note:** this changes the generated schema for existing greenfield apps. Running `opensaas generate` followed by `prisma db push`/`prisma migrate dev` will produce a migration that adds `NOT NULL` to `Session.expiresAt`, `Verification.expiresAt`, `Session.userId`, and `Account.userId`. Since better-auth's own adapter always writes these columns, no existing row should violate the new constraint — but back up production data before applying, as with any schema migration.
--- a/docs/content/reference/fields-api.md
+++ b/docs/content/reference/fields-api.md
@@ -921,6 +921,7 @@ relationship(options: {
   many?: boolean
   db?: {
     foreignKey?: boolean
+    isNullable?: boolean
   }
   ui?: {
     displayMode?: 'select' | 'cards'
@@ -1045,6 +1046,40 @@ The `db.foreignKey` option is only needed for one-to-one relationships where you
 {% callout type="warning" %}
 Setting `db.foreignKey: true` on both sides of a one-to-one relationship will cause a validation error. Only one side can store the foreign key.
 {% /callout %}
+
+##### `db.isNullable`
+
+Controls DB-level nullability of the foreign key column and its relation field, together — they can never disagree.
+
+**Type:** `boolean`
+**Default:** `true` (nullable, matching every relationship generated before this option existed)
+
+**Constraints:**
+
+- Only valid on the FK-owning (single) side of a relationship — the many side has no foreign key column of its own and rejects this option
+
+**Example:**
+
+```typescript
+Session: list({
+  fields: {
+    // Every session genuinely belongs to a user — make the FK required
+    user: relationship({
+      ref: 'User.sessions',
+      db: { isNullable: false },
+    }),
+  },
+})
+```
+
+**Generated Prisma schema:**
+
+```prisma
+model Session {
+  userId String
+  user   User   @relation(fields: [userId], references: [id])
+}
+```
 
 ##### `ui.displayMode`
 
--- a/examples/starter-auth/prisma/schema.prisma
+++ b/examples/starter-auth/prisma/schema.prisma
@@ -48,15 +48,15 @@ model User {
 }
 
 model Session {
-  id        String    @id @default(cuid())
-  token     String    @unique
-  expiresAt DateTime?
+  id        String   @id @default(cuid())
+  token     String   @unique
+  expiresAt DateTime
   ipAddress String?
   userAgent String?
-  userId    String?   @map("user")
-  user      User?     @relation(onDelete: Cascade, fields: [userId], references: [id])
-  createdAt DateTime  @default(now())
-  updatedAt DateTime  @default(now()) @updatedAt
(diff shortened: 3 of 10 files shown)


borisno2:
## Code Review

### Overview

Fixes #863: derived Auth `expiresAt` (Session/Verification) and the `user` FK (Session/Account) were nullable in the generated Prisma schema even though better-auth's own schema — and its adapter's write behavior — treats them as required. The fix adds a general-purpose `db.isNullable` option to `relationship()` fields (mirroring the existing scalar-field convention) and uses it in the auth plugin's list derivation.

### Code quality & correctness

- **Core (`relationship()` / `getPrismaRelation`)**: `db.isNullable` moves the FK column and its relation field together, defaulting to today's nullable behavior when omitted — verified byte-for-byte unchanged output for the default case via test. Good reuse of the existing `db.isNullable` naming convention already used by every scalar field builder.
- **Validation coverage**: the PR validates misuse on the `many` side at field-construction time (mirrors the existing `db.foreignKey` check). During review I found one more silent-no-op path: setting `db.isNullable: false` on the **non-FK-owning side of a one-to-one relationship** (FK ownership isn't known until generation time, since it depends on `db.foreignKey` on either side / alphabetical tie-break across the whole config) previously did nothing — no error, no effect. Fixed in a follow-up commit (`0f2483a`): `getPrismaRelation` now throws with a message naming the correct side, with a test locking in the behavior (`packages/cli/src/generator/prisma.test.ts`).
- **Auth derivation**: `userRelationshipDb` and both `expiresAt` fields now set `db.isNullable: false`, matching better-auth's own schema. Correctly scoped to just the four columns named in the issue — `Account.accessTokenExpiresAt`/`refreshTokenExpiresAt` (genuinely optional per-provider) are untouched.

### Test coverage

- Unit: `relationship()` constructor validation (both the `many` and non-FK-owner rejection paths).
- Generator: FK column/relation nullability for bidirectional and list-only refs, plus the unchanged-default case.
- Auth: derived field config assertions (`derive-auth-lists.test.ts`) and full end-to-end generated-schema assertions (`generated-fk-shape.test.ts`) covering all four columns.
- `examples/starter-auth/prisma/schema.prisma` regenerated to reflect the new shape (verified via `pnpm build` across the workspace, including the example's Next.js build).

### Risk / migration

- This is a schema-changing default for existing greenfield adopters (not opt-in) — intentional per the issue's maintainer-reviewed brief, since better-auth's adapter always writes these columns so no live row should violate the new `NOT NULL`. Documented in the changeset with a migration note.
- `examples/auth-demo/prisma/schema.prisma` was intentionally **not** regenerated in this PR — it already carried unrelated pre-existing drift from an earlier fix (#679) that predates this change, and mixing that fix in here would have widened the diff beyond this issue's scope. Filed as a separate observation, not a blocker.

### Verdict

No outstanding issues. `pnpm build`, `pnpm -w exec turbo run test` (core/cli/auth/ui), `pnpm lint`, `pnpm format`, and the docs link-check all pass on the current head (`0f2483a`).

---

PR #872, opened 2026-08-02
fix(auth): createAuth forwards previously-ignored AuthConfig options

### LINKED ISSUE #865: auth: createAuth silently ignores AuthConfig options it normalizes (minPasswordLength, sendEmail, passwordReset, …)

## Summary

`normalizeAuthConfig` resolves a full `NormalizedAuthConfig`, but `createAuth` (`packages/auth/src/server/index.ts`) reads only part of it. Six documented `AuthConfig` options are normalized and then never forwarded to `betterAuth()`, so the running auth instance silently disagrees with the config the developer wrote — no warning, no type error.

Separately, `getSessionFromAuth` calls `auth.api.getSession()` with an empty `Headers`, so it can only ever return `null`.

## Silently ignored options

| `AuthConfig` option | Normalized in `config/index.ts` | Read by `createAuth` |
| --- | --- | --- |
| `emailAndPassword.minPasswordLength` | ✅ line 47 | ❌ |
| `emailAndPassword.requireConfirmation` | ✅ | ❌ |
| `emailVerification.sendOnSignUp` | ✅ line 55 | ❌ |
| `emailVerification.tokenExpiration` | ✅ line 56 | ❌ |
| `passwordReset.*` (`enabled`, `tokenExpiration`) | ✅ lines 60–65 | ❌ |
| `sendEmail` | ✅ line 87 (with a console fallback) | ❌ |

`createAuth`'s `emailAndPassword` block is:

```ts
emailAndPassword: authConfig.emailAndPassword.enabled
  ? {
      enabled: true,
      requireEmailVerification: authConfig.emailVerification.enabled,
    }
  : undefined,
```

`minPasswordLength` never reaches better-auth, so an app that configures `minPasswordLength: 12` gets better-auth's default of 8 — a security setting silently weakened. `sendEmail` is the one that bites hardest: it is documented on `AuthConfig` as *"Custom email sending function for verification and password reset"*, it is normalized with a `console.log` fallback, and nothing ever calls it — so a greenfield app that follows the docs and sets `sendEmail` ships with **no verification or reset emails at all**, and `passwordReset: { enabled: true }` wires up nothing.

## `session.updateAge` computes the wrong value

```ts
session: {
  ...toBetterAuthModelOptions(authConfig.models.session),
  expiresIn: authConfig.session.expiresIn || 604800,
  updateAge: authConfig.session.updateAge
    ? (authConfig.session.expiresIn || 604800) / 10
    : 0,
},
```

`SessionConfig.updateAge` is typed `boolean`, and when true the refresh interval is derived as `expiresIn / 10` rather than from anything the developer set. With the default 7-day expiry that is 16.8 hours — an arbitrary number that matches neither better-auth's own default (1 day) nor anything configurable. There is no way to express "7-day sessions, refreshed daily", which is better-auth's documented default pairing.

Suggest typing it as `updateAge?: number | false` (seconds), defaulting to better-auth's 1 day, and passing it straight through.

## `getSessionFromAuth` passes empty headers

```ts
export async function getSessionFromAuth(auth, sessionFields: string[]) {
  const session = await auth.api.getSession({
    headers: new Headers(),   // <- no cookies, so never resolves a session
  })
  if (!session?.user) return null
  ...
}
```

A fresh `Headers()` carries no session cookie, so `getSession` always resolves `null` and the function always returns `null`. The doc comment says it is *"used internally by the generated context"*, which makes this look like a silent always-signed-out path for anyone relying on it. It needs a `headers` parameter threaded from the caller's request.

## Impact

We hit this migrating an app with a hand-wired better-auth install onto the stack. The reason we still cannot adopt `createAuth()` is tracked separately (missing `databaseHooks`, email senders, `baseURL`, `cookieCache`, `storeIdentifier`), but the ignored knobs above are a distinct problem: they make the config *look* adopted while behaving differently. Anyone comparing `authPlugin({ ... })` against their live `betterAuth({ ... })` to prove parity will conclude the settings carried over when they did not.

## Proposed

1. Forward `minPasswordLength`, `sendOnSignUp`, `emailVerification.tokenExpiration` (→ `expiresIn`), and the password-reset settings into the `betterAuthConfig` that `createAuth` builds.
2. Wire `sendEmail` into `emailVerification.sendVerificationEmail` and `emailAndPassword.sendResetPassword`, or delete it from `AuthConfig` if it isn't meant to be wired — today it is a documented no-op.
3. Type `session.updateAge` as a duration and pass it through.
4. Give `getSessionFromAuth` a `headers` parameter.
5. Failing that, at minimum: make `normalizeAuthConfig` warn (or `createAuth` throw) when it receives an option it will not forward, so silent divergence isn't the default outcome.

borisno2:
## Summary

`normalizeAuthConfig` resolved a complete `NormalizedAuthConfig`, but `createAuth` only read part of it when building the `betterAuthConfig` passed to `betterAuth()`. Several documented `AuthConfig` options were silently dropped, meaning verification/reset emails never actually sent even when `sendEmail` was configured.

- Forward `emailAndPassword.minPasswordLength` into `betterAuthConfig.emailAndPassword`.
- Wire `passwordReset.enabled`/`tokenExpiration` to better-auth's `sendResetPassword`/`resetPasswordTokenExpiresIn`.
- Wire `emailVerification.enabled`/`sendOnSignUp`/`tokenExpiration` to better-auth's `emailVerification` block (`sendVerificationEmail`/`sendOnSignUp`/`expiresIn`).
- `sendEmail` is now actually invoked by the above callbacks instead of being normalized and never called.
- Retype `session.updateAge` from `boolean` to `number | false` (seconds between refreshes), passed straight through to better-auth instead of the old `expiresIn / 10` arbitrary math. Default changes from `true` to `86400`, matching better-auth's own default.
- `emailAndPassword.requireConfirmation` has no better-auth server-side equivalent (it's UI-only, already covered by `<SignUpForm>`/`<ResetPasswordForm>`'s own `requirePasswordConfirmation` prop) — `createAuth()` now warns via `console.warn` when it's set, rather than silently ignoring it.
- `getSessionFromAuth` gains a required `headers: Headers` parameter and its doc comment is corrected. Previously it always called `auth.api.getSession({ headers: new Headers() })` — an empty header set that can never resolve a session cookie — so it always returned `null`. It has no callers in this repo today (confirmed by grep and by a maintainer's triage comment on the issue); apps hand-roll the same transform against real headers (see `examples/starter-auth/lib/auth.ts`).

Docs (`docs/content/reference/auth.md`, `docs/content/how-to/authentication.md`, `packages/auth/README.md`) updated to reflect the new `updateAge` semantics and clarify `requireConfirmation` is UI-only.

## Test plan

- [x] `packages/auth/tests/server.test.ts` (new) — mocks `better-auth`/`better-auth/adapters/prisma`/`better-auth/next-js` and asserts each forwarded option appears correctly in the `betterAuthConfig` passed to `betterAuth()`, including negative/disabled cases, the `updateAge` number/false mapping, the `requireConfirmation` warning, and `getSessionFromAuth`'s new `headers` behavior.
- [x] `packages/auth/tests/config.test.ts` — added assertions for `session.updateAge`'s new default (86400) and a custom numeric value.
- [x] `pnpm test` (packages/auth) passes — 162/162 (one pre-existing, unrelated failure in `generated-fk-shape.test.ts` due to a missing built dependency, confirmed present before this change via `git stash`).
- [x] `pnpm build` (core, cli, auth) passes.
- [x] `pnpm lint` passes (only pre-existing, unrelated warnings).
- [x] `pnpm manypkg fix` — no changes.
- [x] `pnpm format` — applied.
- [x] Changeset added (`.changeset/short-hats-fetch.md`, minor — calls out the `getSessionFromAuth` signature change and `updateAge` behavior change explicitly).

Closes #865


borisno2:
## Review (via `/review` skill)

Ran 5 finder agents across correctness, removed-behavior, cross-file, simplification, and conventions angles against this PR's diff. Two findings survived verification and are now fixed:

**1. `session.updateAge: false` mapped to the wrong better-auth option (severe, fixed)**

The original implementation mapped `updateAge: false` → better-auth's `updateAge: 0`. But better-auth's actual refresh-decision formula (`shouldBeUpdated = expiresAt - expiresIn*1000 + updateAge*1000 <= Date.now()`) treats `updateAge: 0` as "refresh on essentially every request" — the opposite of disabling refresh, and the opposite of what this PR's own docs/changeset promised. Fixed by setting better-auth's separate `disableSessionRefresh: true` flag instead when `updateAge` is `false`. Updated the test and changeset accordingly.

**2. `requireConfirmation`/`passwordReset` warnings not gated on `emailAndPassword.enabled` (fixed)**

The `requireConfirmation` warning could fire even when `emailAndPassword.enabled` was `false` (nothing to confirm). Also, `passwordReset.enabled` had no warning when `emailAndPassword.enabled` was `false`, even though password reset silently no-ops in that case — inconsistent with this PR's own "warn instead of silently drop" precedent. Fixed both: gated the `requireConfirmation` warning on `emailAndPassword.enabled`, and added a matching warning for `passwordReset.enabled` without `emailAndPassword.enabled`.

Both fixes are covered by new/updated tests in `packages/auth/tests/server.test.ts`; `pnpm test`, `pnpm lint`, `pnpm build` (core + auth) all pass. Changeset updated to describe the corrected `disableSessionRefresh` mechanism.

Lower-severity items considered but left out of scope: a pre-existing falsy-`expiresIn` edge case not introduced by this PR, hardcoded English email copy, and a suggestion to flatten some nested config objects.

Issue #876, opened 2026-08-03
auth: buildBetterAuthOptions returns the widened BetterAuthOptions, erasing better-auth's plugin/session type inference


borisno2:
**Package:** `@opensaas/stack-auth` 0.37.0

First — thank you for #861, #862, #863, #865 and #866. We've now moved a real app fully onto the stack seam: `auth-server.ts` is three lines, the `authPlugin` config is the single source of truth for the whole auth setup, `useBetterAuthTableNames` removed our last `extendPrismaSchema` surgery, and `betterAuthOptions` carries the `databaseHooks`/`cookieCache`/`storeIdentifier` residue. `opensaas generate` is byte-identical to the schema we had before. This is the one rough edge left.

## Summary

`buildBetterAuthOptions()` returns `Promise<BetterAuthOptions>` — the **widened, non-generic** type. better-auth infers most of its public API *from the options object's literal type*, so constructing from that return value erases it:

```ts
export const auth = betterAuth(await buildBetterAuthOptions(config, rawOpensaasContext))
```

gives an `auth` where:

- `auth.api.signInEmailOTP` / `sendVerificationOTP` / `checkVerificationOTP` don't exist (from `emailOTP()`)
- `auth.api.test.*` doesn't exist (from `testUtils()`)
- **`auth.api.getSession()` returns better-auth's default `{ user, session }`** instead of the `customSession` callback's return type

The last one is the painful one. Our `customSession` fully replaces the session shape, and the app's `Session` type is that shape. With inference erased we got ~20 errors of this form across route handlers, server actions and helpers:

```
src/lib/auth.ts(16,3): error TS2322: Type '{ session: {...}; user: {...} } | null'
  is not assignable to type 'Session | null'.
  ... is missing the following properties from type 'Session':
  allowAdminUI, subjectId, email, userId, data
```

## Workaround

Re-pass the plugin array concretely so `betterAuth` can infer from it again:

```ts
// auth-plugin.ts — the config's array plus the nextCookies() the stack appends internally
export const liveBetterAuthPlugins = [...appBetterAuthPlugins, nextCookies()]

// auth-server.ts
export const auth = betterAuth({
  ...(await buildBetterAuthOptions(config, rawOpensaasContext)),
  plugins: liveBetterAuthPlugins,
})
```

[...]

PR #889, opened 2026-08-04
fix(auth): project sessionFields from the resolved session, not just user

### LINKED ISSUE #881: auth: sessionFields describes a flattened projection of the better-auth user, not the resolved session — mismatch is silent

Split out of #876 (final aside). The reporter recorded it as "minor, no action needed"; filing it separately so the contract question doesn't get lost, and so #876's brief stays scoped to the type-erasure fix.

## Observation

`sessionFields` is documented as controlling "what data is available in access control functions", and the auth reference presents it as the session's shape:

```typescript
authPlugin({ sessionFields: ['userId', 'email', 'name', 'role'] })
// session will be: { userId: string, email: string, name: string, role: string }
```

`getSessionFromAuth` is the exported helper that realises that contract. It builds a **flat** record keyed by `sessionFields`, mapping `userId` to the better-auth user's `id` and reading every other name off the better-auth **user** object.

Two things follow that the documented contract doesn't say:

1. **It projects the user, not the session.** A field that lives on the better-auth session rather than the user is never picked up, regardless of being listed in `sessionFields`.
2. **A `customSession` callback can return any shape.** better-auth's `customSession` plugin fully replaces the resolved session; #876's reporter nests most of theirs under `data`. `sessionFields` and the actual resolved session then describe different shapes, and reconciling them is left to the app (they keep a re-nesting helper for it).

In both cases a name in `sessionFields` that isn't found is skipped — no error, no warning, just an absent key. The failure surfaces later as an access-control function reading `undefined`, which fails closed and looks like a permissions bug rather than a config one.

Also worth noting while this is open: `getSessionFromAuth` isn't called by any generated code today — its own docstring says so, and apps hand-roll the same transform against `auth.api.getSession()` (the `starter-auth` example does exactly this). So the helper is the documented contract's only implementation, and nothing exercises it in the default path. It also swallows all errors and returns `null`, which is indistinguishable from "no session".

## Open question for triage

Which of these is `sessionFields` meant to be?

- **A documentation fix** — state that it describes a *flattened view of the better-auth user*, name `customSession` as the case where the resolver's return value and this view diverge, and say the reconciliation is the app's job. Smallest change, contract unchanged.
- **A behavior fix** — have the projection read from the resolved session (whatever `customSession` returned) rather than only from `user`, and/or warn when a listed field can't be resolved instead of silently dropping it. Larger, and needs a call on whether a nested resolver shape should be flattened at all.

Worth deciding alongside it: whether `getSessionFromAuth` should be wired into the generated context (so there's one implementation rather than every app hand-rolling it), or documented as a convenience the examples deliberately don't use.

## Not in scope

The type-erasure fix — that's #876.

borisno2:
## Summary

- `getSessionFromAuth` now projects `sessionFields` off the **resolved** better-auth session (whatever `auth.api.getSession()` returns), not just its `user` sub-object. Resolution precedence for a non-`userId` field: a top-level key on the resolved session, then `user`, then `session`. `userId` stays special-cased to the authenticated user's `id`.
- A `customSession` shape with no top-level `user` key (see #876) is now correctly treated as a signed-in session instead of being misreported as anonymous — `null` means "genuinely no session," nothing else.
- Errors from the underlying `auth.api.getSession()` call now **propagate** instead of being swallowed into `null` — collapsing a lookup failure into "anonymous" is indistinguishable from a mass sign-out under fail-closed access control.
- A `sessionFields` entry that can't be resolved is omitted and logs a warning (once per field, per process) naming what was checked, instead of silently surfacing later as an access-control function reading `undefined`.
- The scaffolded `getSession()` — the CLI feature generator's `lib/auth.ts` template, and `examples/starter-auth`/`examples/auth-demo` — now call this single shared helper, reading `sessionFields` from the resolved config **at runtime** instead of baking a field list in at generation time (so editing `sessionFields` after scaffolding takes effect without regenerating).
- Fixes `examples/auth-demo`'s `getSession()`, which previously returned a truthy object of `undefined` values for an anonymous visitor instead of `null`.
- Updated the `sessionFields` JSDoc and the auth reference doc (`docs/content/reference/auth.md`) to document the flattened-projection contract, the resolution precedence, and that reconciling a `customSession`'s nested shape is the application's job.

## Test plan

- [x] New/updated tests in `packages/auth/tests/server.test.ts` cover: the no-`user`-key `customSession` case, a session-sub-object field, a deliberate top-level/user/session precedence collision, warn-once-per-field behavior, error propagation vs. genuine `null`, and the unchanged happy path — 214/214 auth tests pass
- [x] `packages/cli` test suite passes (346/346)
- [x] `packages/core` test suite passes (948/948)
- [x] `examples/starter-auth` and `examples/auth-demo` regenerate and typecheck cleanly against the updated `lib/auth.ts`
- [x] Changeset added for `@opensaas/stack-auth` and `@opensaas/stack-cli` (minor)

Closes #881

---
_Generated by [Claude Code](https://claude.ai/code/session_01W3AJScedpPQTTZbJfbNqfZ)_

### DIFF
--- a/examples/starter-auth/lib/auth.ts
+++ b/examples/starter-auth/lib/auth.ts
@@ -20,8 +20,7 @@ export async function getSession(): Promise<Session | null> {
   const resolvedConfig = await config
   const authConfig = resolvedConfig._pluginData?.auth as NormalizedAuthConfig | undefined
   const sessionFields = authConfig?.sessionFields ?? ['userId', 'email', 'name']
-  const session = await getSessionFromAuth(auth, sessionFields, await headers())
-  return session as unknown as Session | null
+  return getSessionFromAuth(auth, sessionFields, await headers())
 }
 
 /**
--- a/examples/starter-auth/lib/auth.ts
+++ b/examples/starter-auth/lib/auth.ts
@@ -1,4 +1,6 @@
-import { createAuth } from '@opensaas/stack-auth/server'
+import { createAuth, getSessionFromAuth } from '@opensaas/stack-auth/server'
+import type { NormalizedAuthConfig } from '@opensaas/stack-auth'
+import type { Session } from '@opensaas/stack-core'
 import config from '../opensaas.config'
 import { headers } from 'next/headers'
 import { rawOpensaasContext } from '@/.opensaas/context'
@@ -10,19 +12,16 @@ import { rawOpensaasContext } from '@/.opensaas/context'
 export const auth = createAuth(config, rawOpensaasContext)
 
 /**
- * Get the current session in OpenSaas format
- * Extracts configured sessionFields from Better Auth session
+ * Get the current session in OpenSaas format. Reads `sessionFields` from the
+ * resolved config at runtime, so changing it doesn't require regenerating
+ * this file.
  */
-export async function getSession() {
-  const session = await auth.api.getSession({
-    headers: await headers(),
-  })
-  if (!session || !session.user) return null
-  return {
-    userId: session.user.id,
-    email: session.user.email,
-    name: session.user.name,
-  }
+export async function getSession(): Promise<Session | null> {
+  const resolvedConfig = await config
+  const authConfig = resolvedConfig._pluginData?.auth as NormalizedAuthConfig | undefined
+  const sessionFields = authConfig?.sessionFields ?? ['userId', 'email', 'name']
+  const session = await getSessionFromAuth(auth, sessionFields, await headers())
+  return session as unknown as Session | null
 }
 
 /**
--- a/.changeset/warm-sessions-project.md
+++ b/.changeset/warm-sessions-project.md
@@ -1,0 +1,22 @@
+---
+'@opensaas/stack-auth': minor
+'@opensaas/stack-cli': minor
+---
+
+Fix `getSessionFromAuth` to project `sessionFields` from the _resolved_ better-auth session instead of only its `user` sub-object. A `customSession` plugin's replaced shape with no `user` key is now correctly treated as a signed-in session (never misreported as anonymous), and a session-only field (e.g. the admin plugin's `impersonatedBy`) is now resolvable. Errors from the underlying session lookup now propagate instead of silently becoming `null`, and a `sessionFields` entry that can't be resolved is omitted and logs a warning (once per field, per process) instead of vanishing silently.
+
+The scaffolded `getSession()` — the CLI feature generator's `lib/auth.ts` template, and `examples/starter-auth`/`examples/auth-demo` — now call this single shared helper, reading `sessionFields` from the resolved config at runtime instead of baking a field list in at generation time. `examples/auth-demo`'s `getSession()` also now correctly returns `null` for an anonymous visitor (previously returned a truthy object of `undefined` values).
+
+```typescript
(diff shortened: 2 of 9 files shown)


examples/starter-auth/lib/auth.ts
```diff
@@ -10,19 +12,16 @@ import { rawOpensaasContext } from '@/.opensaas/context'
...
-    userId: session.user.id,
-    email: session.user.email,
-    name: session.user.name,
-  }
+export async function getSession(): Promise<Session | null> {
+  const resolvedConfig = await config
+  const authConfig = resolvedConfig._pluginData?.auth as NormalizedAuthConfig | undefined
+  const sessionFields = authConfig?.sessionFields ?? ['userId', 'email', 'name']
+  const session = await getSessionFromAuth(auth, sessionFields, await headers())
+  return session as unknown as Session | null
```
borisno2: a consumer shouldn't need to cast as unknown for the types to be happen - infact a consumer shouldn't need to cast at all

borisno2:
## Self-review

Ran an independent review pass against the diff (fresh context, not the authoring session) covering the fix in `packages/auth/src/server/index.ts` and the three call-site rewrites.

**Verdict: no correctness issues found.**

- Resolution precedence (`resolveSessionField`) is top-level → `user` → `session`, with `userId` special-cased to `user.id` only (never falling back to the other sources) — verified against the dedicated collision tests.
- `null` is returned only when `auth.api.getSession()` itself is falsy, before ever touching `.user` — a `customSession` shape with no `user` key is projected, not misreported as anonymous.
- No `try/catch` remains in `getSessionFromAuth` — errors from the underlying session lookup propagate, confirmed by a `rejects.toThrow` test.
- The warn-once cache is module-level, keyed on field name, matching the existing `selectWarnings` precedent in `packages/core/src/context/index.ts`; tests correctly isolate it per-case with `vi.resetModules()`.
- All three call sites (CLI generator template, `starter-auth`, `auth-demo`) read `sessionFields` from the resolved config at runtime with no leftover hardcoded field list, and `auth-demo`'s old truthy-anonymous-session bug is fixed.
- The `session as unknown as Session | null` cast at each call site is necessary (not a smell) given `Session` is augmented with concrete field types by consuming apps.

Test coverage maps 1:1 to the issue's acceptance criteria (no-`user`-key customSession, session-only field, real precedence collision, warn-once, error-vs-null, unchanged happy path).

---

examples/starter-auth/lib/auth.ts
borisno2: Good catch, fixed in 35c75f6 — the cast wasn't actually necessary. `getSessionFromAuth` (`packages/auth/src/server/index.ts`) now declares its return type as `Session | null` (from `@opensaas/stack-core`) instead of `Record<string, unknown> | null`. Within `packages/auth`'s own build `Session` is unaugmented, so the internal `Record<string, unknown>` result satisfies it directly with no cast there either — and because the exported signature is already pinned to `Session`, every consumer (this file, `auth-demo`, and the CLI generator's template) now just does `return getSessionFromAuth(...)` with zero casts, regardless of whatever fields a consuming app's own `types/session.d.ts` augmentation adds.

---

PR #919, opened 2026-08-15
fix(core): field-level access rejects non-boolean results instead of allowing

### LINKED ISSUE #913: `checkFieldAccess` fails open when a field rule returns a filter
## What to build

A field-level access rule that returns a Prisma filter is either honoured or rejected loudly. Today it silently grants full access.

`FieldAccessControl` in `packages/core/src/access/types.ts` is typed to return `boolean | PrismaFilter<TItem> | Promise<…>`. So this is a fully type-checked field rule:

```ts
someField: text({
  access: {
    read: ({ session }) => ({ ownerId: { equals: session?.userId } }),
  },
})
```

`checkFieldAccess` handles `false`, handles `true`, and then ends:

```ts
// Default to allowing access if we can't determine
return true
```

The filter case reaches that line. The author wrote a restriction, the type system accepted it, and every caller reads the field.

This is a fail-open default in an access-control primitive, and it is the most dangerous shape a default can have: it is not reached by malformed input, it is reached by a return value the public type explicitly advertises as valid. The evidence that this was an oversight rather than a decision is in the same file — a `matchesFilter` helper sits directly below `checkFieldAccess`, written to evaluate exactly this case, and nothing calls it.

Both remedies are defensible and the choice is a design call:

- **Honour the filter** — wire `matchesFilter` (or a real filter evaluator) so the rule means what its author wrote. Note the write path calls `checkFieldAccess` with no `item` on create, so there is nothing to match against there; that path needs an answer either way.
- **Reject the filter** — narrow `FieldAccessControl` to `boolean` for field-level rules and throw on anything else. Field access is a per-field visibility decision, not a row filter; if filters are not meant to be supported, the type should say so.

What is not defensible is the current combination: advertise it in the type, ignore it at runtime, and default to allow.

## Acceptance criteria

- [ ] A field access rule returning a filter no longer results in access being granted by default
- [ ] Whichever remedy is taken, the type and the runtime agree — a rule that type-checks is a rule the runtime honours
- [ ] The `return true` fallback is removed or made unreachable; any genuinely unreachable branch fails closed
- [ ] `matchesFilter` is either wired up or deleted — dead code in an access-control module reads as coverage that does not exist
- [ ] The `create` path (no `item` available) has an explicit, documented answer
- [ ] Regression coverage asserting a filter-returning field rule does not grant blanket access
- [ ] `sudo` behaviour is unchanged

## Blocked by

None — can start immediately.

borisno2:
## Summary

`checkFieldAccess` (the single evaluator shared by every read and write path) handled a field rule's `false` and `true` results and then fell through to an unconditional `return true` for anything else. `FieldAccessControl`'s public type advertised a third legal result — a Prisma filter — so a fully type-checked field rule like:

```ts
someField: text({
  access: {
    read: ({ session }) => ({ ownerId: { equals: session?.userId } }),
  },
})
```

was silently granted **full read access** instead of being evaluated. This directly contradicted the already-recorded architecture: ADR-0001 states field-level access "returns `boolean` only (a denied field is removed, never used to exclude rows)", and the same conclusion is in `CONTEXT.md`'s "Field-level access" glossary entry. The type and the runtime had simply drifted from that decision — an unused `matchesFilter` helper sitting in the same module, written to evaluate exactly this case and called by nothing, was the clearest sign the drift was accidental.

### Chosen remedy: reject the filter (not honour it)

Recorded in `docs/adr/0030-field-level-access-fails-closed-on-a-non-boolean-result.md`, with the alternatives and why they were rejected. In short:

- **Reject the filter (chosen).** Narrows `FieldAccessControl` to `boolean | Promise<boolean>` and throws a new `InvalidFieldAccessResultError` for any other result. Matches ADR-0001's stated contract, and makes `create` (no `item` to test a filter against) a non-issue rather than a special case — neither operation ever honours a non-boolean, so there's nothing extra to design for `create`.
- **Honour the filter** was rejected: it contradicts ADR-0001 directly, the dead `matchesFilter` helper only understood `equals`/`not` and would need to grow into a partial reimplementation of Prisma's filter language to be trustworthy, and `create` has no `item` to evaluate against.
- **Deny silently (`return false`)** was rejected in favour of throwing: a filter-returning rule is a config bug, not a session being denied. Throwing surfaces it immediately (matches the existing #568 write-path precedent of throwing rather than silently dropping a denied field, and ADR-0022's read-path precedent of failing loud rather than passing through).

## Changes

- `packages/core/src/access/types.ts` — `FieldAccessControl` now returns `boolean | Promise<boolean>` only (was `boolean | PrismaFilter | Promise<...>`); corrected the doc comment that previously (incorrectly) described the fail-open fallback as intentional.
- `packages/core/src/access/field-access.ts` — `checkFieldAccess`'s fallback now throws `InvalidFieldAccessResultError` instead of `return true`; deleted the dead `matchesFilter` helper.
- `packages/core/src/access/errors.ts` — new `InvalidFieldAccessResultError`, a sibling of `AccessScopeDepthExceededError`/`ResolveOutputCycleError` (loud engine failure, distinct from `ValidationError`).
- `packages/core/src/access/index.ts`, `packages/core/src/index.ts` — export the new error alongside its siblings.
- `docs/adr/0030-field-level-access-fails-closed-on-a-non-boolean-result.md` — records the decision.
- `.changeset/plain-jars-whisper.md` — patch changeset for `@opensaas/stack-core`.

`sudo` is unchanged — `checkFieldAccess` still short-circuits to `true` before a sudo context ever calls the field rule, so a sudo-invoked filter-returning rule never reaches the throw (covered by a test).

## Test plan

- [x] `packages/core/src/access/field-access.test.ts` — added a `checkFieldAccess` describe block covering: allow on `true`, deny on `false`, allow when unconfigured, throw on a filter for `read`/`create`/`update`, throw on other non-boolean results (`undefined`), and sudo bypassing the rule entirely (regression test reproducing the issue's exact repro case).
- [x] Added a `filterWritableFields` case asserting a filter-returning rule throws rather than granting a blanket write, for both `create` and `update`.
- [x] `pnpm test` in `packages/core` — 968/968 passing.
- [x] `pnpm build` in `packages/core` and `packages/ui` (the UI package imports `checkFieldAccess`/`FieldAccess`) — both clean.
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` — no errors, no diffs.

Closes #913

---
_Generated by [Claude Code](https://claude.ai/code/session_01DkAFwoAXsp3TzsmzUP9RJ9)_


borisno2:
## Review cycle summary

Ran a full code review (correctness + reuse/simplification/efficiency) after the initial push, plus caught two real CI failures. Both are addressed as of dd4fc72.

### CI failures (real, not flaky)

The `test` and `e2e` jobs failed because `examples/starter`, `starter-auth`, `blog`, `auth-demo`, `composable-dashboard`, and `mcp-demo` all reused a single `isSignedIn`/`isAuthor` helper for **both** operation-level access (which can legitimately return a filter) **and** field-level access (which, after this PR's core fix, no longer type-checks with a filter-returning function). In several of these examples, the field-level usage of `isAuthor` was itself a live instance of the exact bug #913 fixes — a filter-returning rule silently granting the field full access. Fixed by:
- Narrowing `isSignedIn`'s type to `boolean`-only (it never actually returned a filter).
- Replacing the field-level `isAuthor` call sites with inline boolean checks against `item?.authorId`, since field access can't scope rows.

`isAuthor`/`isOwner` themselves are untouched and keep returning filters — they're used only at the operation level now, which is exactly what they're for. Verified with `pnpm generate && tsc --noEmit` (and, for `starter`/`starter-auth`, a full `pnpm build`) across all 7 example apps.

### Code review findings

1. **`InvalidFieldAccessResultError.result: unknown`** exposed `unknown` as part of the package's public API, violating CLAUDE.md's rule against exposing `unknown`/`any` externally. Fixed: dropped the field. The description of what was returned is folded into the error message instead; there's no concretely-typed shape to give it (that's the whole problem being reported).
2. **Changeset was `patch`**, but the PR narrows a publicly exported type (`FieldAccessControl` drops `PrismaFilter`) — a breaking change for TS consumers, which CLAUDE.md's versioning rules reserve for minor/major. Bumped to `minor` with a migration example in the changeset body.

No correctness issues were found in the core fix itself (`checkFieldAccess`, the type narrowing, `matchesFilter` removal) — 968/968 core tests passing, and every call site of `checkFieldAccess` either lets the new throw propagate as intended or already treated any throw as "deny"/"can't decide statically."


---

borisno2:
The `Vercel` status check failed with "Deployment rate limited — retry in 24 hours" — that's an account-level Vercel build quota, unrelated to this PR's code (it's the docs-preview deployment, not the `test`/`e2e` CI checks). Nothing to fix here; it should clear on its own once the quota resets. Not blocking the actual test suite, which is still running/passing on the latest commit.

---

borisno2:
## e2e failure diagnosed and fixed (d8cd020)

The `e2e` job failed on 3 tests, all about the Relationship-table inline-edit affordance in `starter-auth`:
- `05-item-view.spec.ts` › clicking an editable relationship-table cell edits in place instead of navigating
- `08-relationship-inline-edit.spec.ts` › commits a single-field edit... / reverts with a visible reason...

**Root cause:** in the previous commit I wrote the field-level author checks as `item?.authorId` (optional chaining). But `isFieldPotentiallyWritable` (`packages/ui/src/lib/operationAccess.ts`) — which decides whether to show the inline-edit affordance — deliberately calls field-level `update` rules **without** an `item` (it's a per-column decision, not per-row), and its own doc comment says an item-dependent rule is expected to **throw** when it dereferences the missing item, which it then catches and treats as "potentially writable — let the real per-row check at commit time decide."

`item?.authorId` never throws — it quietly evaluates to `undefined`, so the comparison is always `false` and the affordance stopped showing for the signed-in author. Switched to `item!.authorId` (dereferences unguarded, throws when actually absent) across all four examples that use this pattern (auth-demo, composable-dashboard, mcp-demo, starter-auth), with a comment explaining why the non-null assertion is deliberate.

Verified with `tsc --noEmit` on all four examples (clean). Couldn't run the Playwright suite itself locally — this sandbox's Prisma guard correctly refuses `db push` for an unattended agent — but the fix is a direct, mechanical restoration of the documented contract in `operationAccess.ts`, and matches the test's own comment ("a Post's columns are all editable for its author, so clicking one opens the inline editor").


---

Issue #915, opened 2026-08-15
A field `read` gate withholds the value but not the predicate — every gated column is a filter oracle


borisno2:
## What to build

A field whose `read` access denies the caller cannot be named in `where` or `orderBy` either. Today the gate withholds the **value** and leaves the **predicate** entirely unconstrained, so the withheld value can be recovered by probing.

Take a list with a public `query` gate and a field-level `read` gate on one column — the shape you reach for when a model needs a public identity but carries a few private columns:

```ts
Organisation: list({
  access: { operation: { query: () => true } },
  fields: {
    name: text(),
    slug: text({ isIndexed: 'unique' }),
    billingAddress: text({ access: { read: isOwnRow } }),
  },
})
```

With no session at all, through `context.db`:

```ts
await context.db.organisation.count({ where: { billingAddress: { startsWith: '12 ' } } }) // => 1
await context.db.organisation.count({ where: { billingAddress: { startsWith: '99 ' } } }) // => 0
```

`count()` alone recovers the withheld value one character at a time. `findMany({ where: { billingAddress: { contains: … } } })` and `findMany({ orderBy: { billingAddress: 'asc' } })` discriminate the same way — the second leaks ordering across the whole table without naming a value.

## Why no access rule can close this

- Field-level `read` access is **post-query only**. `filterReadableFields` strips denied keys from rows that have already come back, and it is the sole consumer of `checkFieldAccess(…, 'read', …)` on any read path.
- `findMany` and `count` call `checkAccess(queryAccess, { session, context })` — the gate is **never handed `args.where`**, so no access rule can see the caller's predicate, let alone restrict which columns it names.
- The result is combined via `mergeFilters(args.where, accessResult)`, an AND. For any constant access filter `F`, `count(P AND F)` still varies with `P`. This is not a matter of writing a cleverer gate.
- `orderBy` is passed to Prisma untouched.
- There are no hooks on the read path.

So a list's predicate surface is currently **exactly its column set**, and the only levers an application author has are which columns sit on a publicly-queryable model and whether the model is publicly queryable at all. Both are schema-shape workarounds for what is an access-layer gap.

## This is reachable in the stack's own admin UI

Established during triage, and it is why this is not merely a latent hazard for applications that choose to forward a caller-supplied `where`:

[...]

PR #920, opened 2026-08-15
fix(core): reject undeclared where/orderBy keys on findMany and count

### LINKED ISSUE #912: Read path forwards undeclared `where` / `orderBy` keys straight to Prisma
## What to build

A read's `where` and `orderBy` keys are validated against the list config and rejected when the list does not declare them — the same fail-closed rule the write path already applies to `data` keys.

The write path settled this in #564: an undeclared key in `data` throws, because the generated Prisma model has strictly more fields than the config declares, and letting an undeclared key through hands a non-sudo caller a surface the config never exposed. `sudo` is the single trusted bypass.

The read path has no equivalent. The caller's `where` and `orderBy` go to Prisma unchanged (after `mergeFilters` ANDs in the access filter). So the same asymmetry the write path closed is still open on reads.

The concrete surface this exposes is **back-relations**. Prisma emits a back-relation on a model for *every* foreign key pointing at it, whether or not the list config declares the reverse relationship. Making a relationship one-sided in the list config does not remove it from the generated client. An anonymous caller can therefore filter a publicly-queryable list by a relation the config never declared:

```ts
// `Organisation` declares no `documents` relationship. Prisma emits
// `from_Document_organisation` anyway, because `Document` holds an FK to it.
await context.db.organisation.count({
  where: { from_Document_organisation: { some: { /* … */ } } },
})
```

This ticket is the key-walking seam. It does not by itself decide what a *declared* key is allowed to do — that is the field-access question (#915) and the relation-access question (#916), both of which need this walk to exist first.

## Scope: the seam is `findMany` and `count`

Corrected during triage — the original criteria named four operations, two of which are already guarded:

- **`findUnique`** already validates its `where` (`assertUniqueWhere`: exactly one key, and it must be a unique key). Not vulnerable, and not in scope.
- **`findFirst`** delegates to `findMany`, so it is covered for free.
- **`update` / `delete`** type `where` as `{ id: string }`.
- **`updateMany` / `deleteMany`** route their `where` through `findMany`, so they are covered for free.

That leaves `findMany` and `count` as the entire surface. Both must be fixed — `count` is not a lesser case, it is the one that leaks most cleanly, since a bare count answers a predicate with no rows returned at all.

## Acceptance criteria

- [ ] An undeclared key in `where` or `orderBy` throws, on `findMany` and `count`
- [ ] `findFirst`, `updateMany` and `deleteMany` inherit the behaviour through `findMany` rather than growing their own copy of the check
- [ ] The rejection is loud — an error naming the list and the offending key, not a silent drop, so a caller never gets a result set that quietly ignored their predicate
- [ ] Nested keys are walked, not just the top level: an undeclared key inside a relation filter or a logical operator (`AND` / `OR` / `NOT`) is rejected the same way
- [ ] Prisma's own filter operators (`equals`, `contains`, `startsWith`, `in`, `some`, `every`, `none`, `is`, …) are not mistaken for field names
- [ ] `sudo` bypasses the check, matching the write path
- [ ] Foreign key scalars the config implies rather than declares (e.g. `authorId` for `author: relationship(…)`) remain filterable — this must not break ordinary reads
- [ ] Regression coverage for a back-relation Prisma generates but the list config does not declare

## Blocked by

None — can start immediately.

borisno2:
## Summary

- Adds `packages/core/src/access/query-validation.ts`: a recursive walk (`validateQueryKeys`) that validates a caller-supplied `where`/`orderBy` against the list config before the query reaches Prisma, mirroring the write path's #564 undeclared-key reject.
- Wires the validation into `createFindMany` and `createCount` in `packages/core/src/context/index.ts` (the only two operations the issue identifies as the real surface — `findUnique` already validates its unique `where`, and `findFirst`/`updateMany` inherit for free by delegating to the same validated `findMany`).
- The walk recurses into logical operators (`AND`/`OR`/`NOT`) and relation quantifiers (`some`/`every`/`none`/`is`/`isNot`), switching to the *related* list's fields when it crosses a relationship — so a nested key is checked against the right list.
- Prisma's own filter operator vocabulary (`equals`, `contains`, `startsWith`, `in`, …) and aggregate ordering (`_count`) are never mistaken for field names.
- System fields (`id`, `createdAt`, `updatedAt`), implied foreign-key scalars (e.g. `authorId` for `author: relationship(...)`), and multi-column field sub-columns (`getColumnNames`) remain filterable, matching the same allowances the write path already grants.
- `sudo` bypasses the check entirely, matching the write path's escape hatch.
- Only the caller-supplied `where`/`orderBy` is walked — the access filter produced by the list's own `query` access control is trusted config and is never walked (that's out of scope here, and belongs to #915/#916).

## Test plan

- [x] New `describe('read-path key validation (#912)')` block in `packages/core/tests/context.test.ts` covering: undeclared `where` key on `findMany`/`count`, undeclared `orderBy` key, nested `AND`/`OR`/`NOT`, nested relation filter (naming the *related* list), the back-relation regression case from the issue (`from_Document_organisation`), operator-vs-field-name disambiguation, `sudo` bypass, implied FK scalar filterability, `findFirst`/`updateMany` inheriting the check, and an ordinary read remaining unaffected.
- [x] `pnpm test` (core package): 970/970 passing, no regressions in the existing suite.
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all clean.
- [x] `tsc --noEmit` clean.
- [x] Changeset added (`@opensaas/stack-core`: patch).

Closes #912

---
_Generated by [Claude Code](https://claude.ai/code/session_01YD1qbFLq2BhYgVcqAP1j7G)_

### DIFF
--- a/packages/core/src/access/query-validation.ts
+++ b/packages/core/src/access/query-validation.ts
@@ -128,7 +128,7 @@ function walkWhere(
     // Scalar field filters use Prisma's own operator vocabulary (`equals`,
     // `contains`, `in`, …) and never nest another field name — trusted as-is,
     // no further walk needed. Only a relationship field's filter nests a
-    // WHERE clause for another list, via a relation quantifier.
+    // WHERE clause for another list.
     if (
       resolved.isRelationship &&
       value !== null &&
@@ -137,14 +137,27 @@ function walkWhere(
     ) {
       const related = getRelatedListConfig(resolved.fieldConfig.ref, config)
       if (!related) continue
-      for (const [quantifier, quantifierValue] of Object.entries(
-        value as Record<string, unknown>,
-      )) {
-        if (RELATION_QUANTIFIERS.has(quantifier)) {
-          walkWhere(quantifierValue, related.listConfig, related.listName, config, isSudo)
+
+      const relationEntries = Object.entries(value as Record<string, unknown>)
+      const hasQuantifier = relationEntries.some(([k]) => RELATION_QUANTIFIERS.has(k))
+
+      if (hasQuantifier) {
+        // Wrapped form: `{ author: { is: {...} } }` / `{ posts: { some: {...} } }`.
+        // Only the quantifier's own value is a nested WHERE clause for the
+        // related list.
+        for (const [quantifier, quantifierValue] of relationEntries) {
+          if (RELATION_QUANTIFIERS.has(quantifier)) {
+            walkWhere(quantifierValue, related.listConfig, related.listName, config, isSudo)
+          }
         }
-        // Other keys under a relation field (e.g. `equals: null` for a to-one
-        // nullability check) are not nested WHERE clauses — nothing to walk.
+      } else {
+        // Direct-nesting form: Prisma's documented default for a to-one
+        // relation filter nests the related list's own fields with no `is`
+        // wrapper at all (`{ author: { email: { contains: '...' } } }`). The
+        // whole value object IS the nested WHERE clause here — walk it
+        // directly, or an undeclared key reached exactly this way (one hop
+        // through a to-one relation) would pass through unchecked.
+        walkWhere(value, related.listConfig, related.listName, config, isSudo)
       }
     }
   }
--- a/packages/core/src/access/query-validation.ts
+++ b/packages/core/src/access/query-validation.ts
@@ -1,0 +1,206 @@
+import type { ListConfig, OpenSaasConfig } from '../config/types.js'
+import { getRelatedListConfig } from './engine.js'
+import { ValidationError } from '../hooks/index.js'
+
+/**
+ * #912 — read-path key validation.
+ *
+ * The write path settled #564: an undeclared key in `data` throws, because the
+ * generated Prisma model has strictly more fields than the config declares (most
+ * notably back-relations — Prisma emits one for every inbound foreign key, whether
+ * or not the list config declares the reverse relationship). Reads had no
+ * equivalent — a caller's `where`/`orderBy` reached Prisma unchanged, so an
+ * anonymous caller could filter or order by a relation the config never exposed.
+ *
+ * This module is the read-path counterpart: every key named in a caller's `where`
+ * or `orderBy` is resolved against the list config before the query runs. A key
+ * with no entry in the config throws, naming the list and the key. `sudo` is the
+ * single trusted bypass, mirroring the write path.
+ *
+ * Deliberately NOT walked: the access filter produced by the list's own `query`
+ * access control. That filter is trusted config authored by the same person who
+ * declares the fields — walking it would make this an access-control decision
+ * (that's #915/#916), not the key-existence seam this ticket establishes.
+ */
+
+// Prisma's logical combinators for a WHERE clause — never field names.
+const LOGICAL_OPERATORS = new Set(['AND', 'OR', 'NOT'])
+
+// Prisma's relation quantifiers. The value nested under one of these is itself a
+// WHERE clause for the RELATED list, and is walked against that list's fields.
+const RELATION_QUANTIFIERS = new Set(['some', 'every', 'none', 'is', 'isNot'])
+
+// Always present, never declared in a list's `fields` — the write path
+// (`filterWritableFields`) excludes the same three names from `fieldConfigs`.
+const SYSTEM_FIELDS = new Set(['id', 'createdAt', 'updatedAt'])
+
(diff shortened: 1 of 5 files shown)


packages/core/src/access/query-validation.ts
```diff
@@ -0,0 +1,206 @@
...
+
+function walkWhere(
+  where: unknown,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
+  listConfig: ListConfig<any>,
+  listName: string,
+  config: OpenSaasConfig,
+  isSudo: boolean,
+): void {
+  if (where === null || typeof where !== 'object') return
```
borisno2, line 109: **Undeclared-key check is bypassable for to-one relation filters that don't use `is`/`isNot` (the standard, documented Prisma syntax).**

`walkWhere`'s relation-recursion only fires when a key nested under a relationship field is one of `RELATION_QUANTIFIERS` (`some`/`every`/`none`/`is`/`isNot`):

```ts
for (const [quantifier, quantifierValue] of Object.entries(value as Record<string, unknown>)) {
  if (RELATION_QUANTIFIERS.has(quantifier)) {
    walkWhere(quantifierValue, related.listConfig, related.listName, config, isSudo)
  }
  // Other keys under a relation field ... are not nested WHERE clauses — nothing to walk.
}
```

But Prisma's canonical way to filter a to-one relation is to nest the related list's fields **directly**, with no `is`/`isNot` wrapper at all (this is literally the example in Prisma's own docs: `where: { author: { email: { contains: '...' } } }`, no `is`). For that shape, the loop's `quantifier` variable is the related list's field name (e.g. `email`, or an undeclared back-relation name), which is never in `RELATION_QUANTIFIERS`, so the `if` never fires and **nothing gets walked or validated** — the undeclared key silently passes through to Prisma unchecked.

Concrete failure: given `Post.author = relationship({ ref: 'User.posts' })` and a Prisma-generated back-relation on `User` that the config never declares (e.g. `from_SecretList_userRef`, exactly the #912 threat model this PR targets):

```ts
context.db.post.findMany({
  where: { author: { from_SecretList_userRef: { some: {} } } },
})
```

Trace: `resolveQueryField('author', ...)` resolves fine (declared relationship) → `resolved.isRelationship === true` → the code iterates `Object.entries({ from_SecretList_userRef: { some: {} } })` → `quantifier = 'from_SecretList_userRef'` → not in `RELATION_QUANTIFIERS` → skipped, no recursion, no throw. The where clause reaches Prisma unchanged and (if `from_SecretList_userRef` is a real generated column, which it is under the exact scenario this PR's own test constructs for the root case) executes successfully — the exact "regression that matters most" back-relation leak, just reached one hop through a to-one relation instead of at the root.

This also isn't only a root-level gap — it recurs at any depth reached via a to-one hop (including inside an `is`/`isNot` wrapper that itself nests a further to-one relation), so the fix needs to walk relation-filter values whether or not a quantifier wraps them: when no `RELATION_QUANTIFIERS` key is present, the object's own keys ARE the related list's where clause and must be walked directly (falling back to `walkWhere(value, related.listConfig, ...)` rather than skipping).

The existing test suite doesn't catch this — the only relation-filter test uses the wrapped form (`{ author: { is: { bogusKey: 'x' } } }`); there's no test for the unwrapped/direct form.

(Single-pass review without the multi-agent Agent tool fan-out — flagging for a second look rather than treating this as fully cross-verified.)

---

borisno2 (commented):
Single-pass review of PR #920 (no Agent tool fan-out / subagent verify pass available in this context). Went through all 8 angles myself; one finding survived cross-checking against the diff with a concrete trace — posted inline. Everything else I looked at (findUnique's separate unique-where guard, updateMany/findFirst delegating to the validated findMany, sudo bypass, FK-scalar/system-field/split-column allowances, the orderBy path which is not affected) checked out.

---

packages/core/src/access/query-validation.ts
borisno2: Good catch — confirmed and fixed in bec9592. `walkWhere` now checks whether the relation filter's value contains any `RELATION_QUANTIFIERS` key; if none is present (Prisma's direct-nesting form for to-one relations), the whole value is walked directly as a nested WHERE clause for the related list instead of being skipped. Added regression tests for both the bypass (`{ author: { bogusKey: 'x' } }` throws) and the legitimate direct-nesting form (`{ author: { name: { equals: 'John' } } }` still works).

---

Issue #916, opened 2026-08-15
Relation filters in `where` bypass the related list's `query` access, unlike `include`


borisno2:
## What to build

A relation filter in `where` is scoped by the related list's `query` access, the way `buildAccessScopedInclude` already scopes `include`.

The stack takes access control seriously on the **shape of what comes back**: `buildAccessScopedInclude` walks the caller's `include` and applies each related list's `query` access as it goes, so a caller cannot pull back rows from a list they may not query. Nothing does the equivalent for `where`. A relation filter is handed to Prisma unchanged, and Prisma evaluates it against every row in the related table.

The result is that **a list's predicate surface is not its own column set — it is the transitive closure of everything reachable by foreign key**. Given a publicly-queryable parent and a properly gated child, with no session at all:

```ts
// `Document` has a membership-scoped `query` gate. `findMany` on it returns [].
await context.db.document.findMany() // => []

// The same column, reached as a filter through the parent, discriminates fine:
await context.db.organisation.count({
  where: { documents: { some: { billingAddress: { startsWith: '12 ' } } } },
}) // => 1   ('99 ' => 0)
```

Two hops behave identically, which is where this gets sharp — a relation chain to an identity list lets an anonymous caller confirm whether a given email address is associated with a given parent row:

```ts
await context.db.organisation.count({
  where: { members: { some: { user: { is: { email: '<known address>' } } } } },
}) // => 1 for a real pairing, 0 otherwise
```

Two properties combine to make this unavoidable at the application layer:

- Prisma emits a back-relation for **every** FK pointing at a model, whether or not the list config declares the reverse relationship. Declaring the relationship one-sided does not remove the path. (#912 closes the *undeclared* half of this; this ticket is about the declared half, which #912 will correctly let through.)
- Relation filters are not access-checked the way `include` is.

So this applies to **every** list reachable from a publicly-queryable one, and it cannot be closed by schema shape — moving private columns onto a separate, properly-gated list leaves them reachable by exactly this route.

## This is the unclosed half of ADR-0022, and it has been hand-fixed twice

Established during triage. The decision this ticket implements is already recorded — it was just applied to one of the two inputs:

- **ADR-0022** ("access control fails closed when it cannot scope a read") closed exactly this fail-open shape for `include`, on the principle that *an engine that cannot compute a scope must deny, never pass through*. It says nothing about `where`, because `where` was not what was being fixed. The same argument applies unchanged.
[...]

PR #925, opened 2026-08-16
fix(core,ui): field read access gates where/orderBy, not just returned rows

### LINKED ISSUE #915: A field `read` gate withholds the value but not the predicate — every gated column is a filter oracle

borisno2:
## Summary

A field-level `read` gate withheld the **value** but left the **predicate** entirely unconstrained: a read-denied field could still be named in a `where`/`orderBy`, so its value (or relative order) could be recovered by probing — `count()` is the cleanest instrument, since it answers a predicate while returning no rows at all. This was reachable through the stack's own admin UI (ordinary list-view filtering and sorting), not only via an application forwarding a caller-supplied `where`.

- `findMany`/`count` now reject a `where`/`orderBy` key naming a field the session cannot read, including keys nested inside `AND`/`OR`/`NOT` — throwing loudly (`ValidationError`) instead of returning a silently narrowed or empty result. `sudo` still bypasses, matching #912.
- A `read` rule that depends on the fetched row (`item`) cannot be evaluated before the query runs — there is no row yet. Rather than skip the check for such a rule (which would reopen the hole for exactly the fields most likely to be sensitive), it resolves to a denial. Detected by handing the rule a poisoned `item` Proxy that throws on any property access, including through optional chaining — see `docs/adr/0031`.
- A field rule returning a non-boolean (#913's `InvalidFieldAccessResultError`) propagates unchanged rather than being folded into a plain denial, so a config bug stays visibly a config bug.
- `collectFilterSpecs`/`buildListFilterWhere`/`collectFilterSuggestions` (`@opensaas/stack-core`) now take a required `{ session, context }` and exclude read-denied fields, so the admin UI never suggests, autocompletes, or submits a filter the engine is going to reject — a `field:value` token for such a field degrades to free text instead.
- The admin list view's sort validation (`@opensaas/stack-ui`, `ListView.tsx`) excludes the same fields from what a `?sort=` URL param may activate.
- Deliberately scoped to the **current list only** — a relationship field's own `read` access is checked, but a field on a *related* list reached through a relation filter is out of scope (tracked separately as #916), matching the issue's stated boundary.

## Test plan

- [x] `pnpm test` — core: 1008 passed (was 991); UI: 563 passed (was 559)
- [x] `pnpm exec tsc --noEmit` clean on both `@opensaas/stack-core` and `@opensaas/stack-ui`
- [x] New regression coverage:
  - `count()` probe returns the identical denial for a matching and a non-matching prefix (no oracle)
  - `orderBy` on a read-denied field is denied
  - keys nested inside `AND`/`OR`/`NOT` are checked
  - a row-dependent `read` rule (including one using `item?.x` optional chaining) resolves to denial
  - `InvalidFieldAccessResultError` propagates rather than being swallowed
  - `sudo` bypasses the check entirely
  - a readable field remains filterable/sortable (no regression)
  - a read-denied field is absent from collected Filter specs, from filter suggestions, and is excluded from the sortable set
  - a relation-filter key on a *related* list is unaffected (out of scope, #916)
- [x] Changeset added (`@opensaas/stack-core`, `@opensaas/stack-ui`, minor)
- [x] ADR added: `docs/adr/0031-a-predicate-cannot-name-a-field-the-session-cannot-read.md`

Closes #915

---
_Generated by [Claude Code](https://claude.ai/code/session_01E6e4y65UkDMY7FayQ5jpQz)_

### DIFF
--- a/packages/core/src/access/field-access.ts
+++ b/packages/core/src/access/field-access.ts
@@ -7,6 +7,44 @@ import { ValidationError } from '../hooks/index.js'
 import { InvalidFieldAccessResultError } from './errors.js'
 
 /**
+ * Marks a throw caused by touching {@link createPoisonedItem}'s `item`, as
+ * opposed to some other error a field rule legitimately raises. Not exported
+ * — callers only ever see its effect (a `false` from
+ * `isFieldReadableForPredicate`), never the class itself.
+ */
+class PredicateTimeItemAccessError extends Error {}
+
+/**
+ * An `item` that throws {@link PredicateTimeItemAccessError} on ANY attempt to
+ * read a property off it — including via optional chaining (`item?.x`),
+ * since the Proxy itself is a truthy object and optional chaining still
+ * performs the property read once its base is non-nullish. Used by
+ * `isFieldReadableForPredicate` to detect a read rule that depends on the
+ * fetched row: there is no row yet at predicate-evaluation time (see that
+ * function's doc), so any rule that reaches into `item` at all cannot be
+ * answered here.
+ */
+function createPoisonedItem(): Record<string, unknown> {
+  return new Proxy(
+    {},
+    {
+      get(_target, prop) {
+        throw new PredicateTimeItemAccessError(String(prop))
+      },
+      has(_target, prop) {
+        throw new PredicateTimeItemAccessError(String(prop))
+      },
+      ownKeys() {
+        throw new PredicateTimeItemAccessError('ownKeys')
+      },
+      getOwnPropertyDescriptor(_target, prop) {
+        throw new PredicateTimeItemAccessError(String(prop))
+      },
+    },
+  ) as Record<string, unknown>
+}
+
+/**
  * Shared field-level access evaluation.
  *
  * This module is the single, canonical home for field-level access checks. Both
@@ -89,6 +127,57 @@ export async function checkFieldAccess(
   // available, so `create` (which has no `item` to evaluate a filter against)
   // needs no separate answer: neither operation ever honours a non-boolean.
   throw new InvalidFieldAccessResultError(operation, result)
+}
+
+/**
+ * Whether a field's `read` access allows it to be NAMED in a `where`/`orderBy`
+ * predicate, evaluated BEFORE the query runs (#915).
+ *
+ * Field-level `read` access has always been a post-query check (Field
+ * Visibility, `field-visibility.ts`): it strips a denied key from a row that
+ * has already been fetched. That leaves the predicate itself unconstrained —
+ * a caller can still filter or sort by a field whose value they could never
+ * read, and recover it (or its relative order) by probing. This function is
+ * the pre-query counterpart, called from `query-validation.ts`'s `where`/
+ * `orderBy` walk for every key that resolves to a declared field.
+ *
+ * It delegates to the SAME evaluator Field Visibility uses
+ * (`checkFieldAccess`) — there is no second, parallel field-access evaluator
+ * for predicates, matching this module's own rule for every other caller.
+ * The one difference is the `item` it hands the rule: there is no fetched row
+ * yet, so a rule that depends on one (the shape `FieldAccess['read']`
+ * documents as the norm, e.g. `item?.ownerId === session?.userId`) cannot be
+ * answered here. Rather than skip the check for such a rule — which would
+ * reopen exactly the hole this closes for the fields most likely to be
+ * sensitive — it is handed a poisoned `item` that throws on any property
+ * read, and that throw is caught and resolved to `false`: a row-dependent
+ * `read` rule always denies at predicate time, documented and deliberate
+ * (see docs/adr/0031). A rule that never touches `item` (checking only
+ * `session`, the common case for a field meant to be filterable/sortable at
+ * all) evaluates normally and returns its real answer.
+ *
+ * A rule that returns a non-boolean is a distinct, louder failure (#913,
+ * ADR-0030) — `InvalidFieldAccessResultError` — and is deliberately NOT
+ * folded into the `false` here; it propagates so the config bug it signals is
+ * never mistaken for an ordinary field-level denial.
+ */
+export async function isFieldReadableForPredicate(
+  fieldAccess: FieldAccess | undefined,
+  args: {
+    session: Session | null
+    context: AccessContext & { _isSudo?: boolean }
+  },
+): Promise<boolean> {
+  try {
+    return await checkFieldAccess(fieldAccess, 'read', {
+      session: args.session,
+      context: args.context,
+      item: createPoisonedItem(),
+    })
+  } catch (err) {
+    if (err instanceof InvalidFieldAccessResultError) throw err
+    return false
+  }
 }
 
 /**
--- a/packages/core/src/access/field-access.ts
+++ b/packages/core/src/access/field-access.ts
@@ -175,8 +175,12 @@ export async function isFieldReadableForPredicate(
       item: createPoisonedItem(),
     })
   } catch (err) {
-    if (err instanceof InvalidFieldAccessResultError) throw err
-    return false
+    // Only the poisoned-item signal means "row-dependent, deny". Anything
+    // else — including `InvalidFieldAccessResultError` and a genuine bug in
+    // the rule itself — propagates unchanged rather than being silently
+    // folded into an ordinary denial (found in review of #925).
+    if (err instanceof PredicateTimeItemAccessError) return false
+    throw err
   }
 }
 
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -1135,35 +1135,8 @@ function createFindMany<TPrisma extends PrismaClientLike>(
       )
     }
 
-    // #912 — reject a `where`/`orderBy` key the list config doesn't declare
-    // (e.g. a Prisma-generated back-relation) before it ever reaches the
-    // access filter merge below. `sudo` bypasses, matching the write path.
-    validateQueryKeys({
-      where: args?.where,
-      orderBy: args?.orderBy,
-      listConfig,
-      listName,
-      config,
-      isSudo: context._isSudo === true,
-    })
-
-    // #915 — reject a `where`/`orderBy` key naming a field this session
-    // cannot READ, so a field-level `read` gate can't be probed via a
-    // predicate (a `count()` that varies with the withheld value, an
-    // `orderBy` that leaks relative ordering). Runs after the #912 check
-    // above so an undeclared key is always #912's rejection, never this
-    // one's. `sudo` bypasses, matching #912.
-    await validateQueryFieldReadAccess({
-      where: args?.where,
-      orderBy: args?.orderBy,
-      listConfig,
-      listName,
-      session: context.session,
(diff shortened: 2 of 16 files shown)


packages/core/src/context/index.ts
```diff
@@ -1146,6 +1147,22 @@ function createFindMany<TPrisma extends PrismaClientLike>(
...
+      where: args?.where,
+      orderBy: args?.orderBy,
+      listConfig,
+      listName,
+      session: context.session,
+      context,
+      isSudo: context._isSudo === true,
+    })
+
     // Check query access (skip if sudo mode)
```
borisno2: **[high-effort single-pass review, no Agent-tool fan-out]** — `validateQueryKeys`/`validateQueryFieldReadAccess` (added just above, lines 1141/1156) run *before* the operation-level `query` access check on this line, in both `findMany` (here) and `count` (same pattern at line ~1428-1449).

This means a caller with **zero** operation-level access to a list (e.g. `access.operation.query: () => false`) can now enumerate the list's field names and their field-level `read`-access status via the thrown `ValidationError` messages, before the list-level denial is ever evaluated. Previously (pre-PR), `query: () => false` short-circuited to `[]`/`0` uniformly regardless of the `where` clause, with no distinguishing signal.

Concrete scenario: list `SecretDocs` has `access.operation.query: () => false` and a field `passphrase: text({ access: { read: () => false } })`. An anonymous caller runs `context.db.secretDocs.count({ where: { passphrase: { startsWith: 'a' } } })`. Instead of the expected silent `0` (per CLAUDE.md's "Silent Failures... prevent information leakage" principle), it throws `"Cannot query \"SecretDocs\" — \"passphrase\" is denied by field-level read access"` — confirming to a fully-unauthorized caller that the list exists, has that field, and that the field is read-gated. This is new information exposure introduced by this PR's own ordering, not something #915 set out to add.

Suggest moving the `validateQueryKeys`/`validateQueryFieldReadAccess` calls to run *after* the operation-level `query` access check returns something other than `false` (or gating the thrown messages so they never fire for a statically/dynamically fully-denied list).

---

packages/core/src/access/field-access.ts
```diff
@@ -91,6 +129,57 @@ export async function checkFieldAccess(
...
+): Promise<boolean> {
+  try {
+    return await checkFieldAccess(fieldAccess, 'read', {
+      session: args.session,
+      context: args.context,
+      item: createPoisonedItem(),
+    })
+  } catch (err) {
+    if (err instanceof InvalidFieldAccessResultError) throw err
+    return false
```
borisno2: **[high-effort single-pass review, no Agent-tool fan-out]** — This `catch` is broader than its own documented intent. The `PredicateTimeItemAccessError` class (line 15) and its doc comment explicitly say it exists to distinguish "a throw caused by touching `item`... as opposed to some other error a field rule legitimately raises," and ADR-0031 says the poisoned item is what "`isFieldReadableForPredicate` catches and turns into `false`." But the actual catch here has no `instanceof PredicateTimeItemAccessError` check — it swallows *any* thrown error except `InvalidFieldAccessResultError` and converts it to `false`.

Concrete scenario: `read: ({ session }) => { if (session === null) throw new Error('unexpected anonymous access'); return true }` — a rule that never touches `item` but throws for its own reasons. Called from `isFieldReadableForPredicate`/`checkKeyReadableOrThrow` (query-validation.ts) or `collectFilterSpecs` (filter/collect.ts) or `isSortableField` (ui/ListView.tsx) with `session: null`, this genuine error is silently swallowed and turned into "field not filterable/sortable" instead of propagating — masking a real bug in the access rule as an ordinary field-level denial, and behaving inconsistently with the post-query path (`filterReadableFields`), which lets the same throw propagate.

Suggest `if (err instanceof PredicateTimeItemAccessError) return false; throw err` (after the existing `InvalidFieldAccessResultError` re-throw), so only the poisoned-item signal is converted to a denial.

---

packages/core/src/filter/collect.ts
```diff
@@ -1,28 +1,54 @@
...
   listConfig: ListConfig<any>,
   listKey: string,
   config: OpenSaasConfig,
-): Record<string, FilterSpec> {
+  args: FilterAccessArgs,
+): Promise<Record<string, FilterSpec>> {
   const specs: Record<string, FilterSpec> = {}
   for (const [fieldName, field] of Object.entries(listConfig.fields)) {
     if (typeof field.getFilterSpec !== 'function') continue
+    const readable = await isFieldReadableForPredicate(field.access, args)
```
borisno2: **[high-effort single-pass review, no Agent-tool fan-out; lower-confidence/efficiency, not a correctness bug]** — `collectFilterSpecs` now `await`s `isFieldReadableForPredicate` once per filterable field, sequentially in the `for` loop. Since the result is just assembled into a plain `Record` (unlike the query-validation walk, which needs sequential awaiting to deterministically throw for the *first* denied key), field order doesn't matter here — these per-field access evaluations are independent and could run concurrently (e.g. `Promise.all` over the filterable-field entries) instead of serially. For a list with many fields, and a `read` rule that does any async work (e.g. a DB lookup), this serializes list-view page latency that could otherwise be parallel.

---

borisno2 (commented):
Automated single-pass code review (high effort, no Agent-tool fan-out — the usual multi-agent fan-out/subagent verify pass did not run for this review; treat findings as unverified-by-a-second-pass). 3 findings posted inline, ranked most-severe first: (1) the new #912/#915 predicate validation runs *before* the operation-level `query` access check on `findMany`/`count`, letting a caller with zero list access enumerate field names and field-read-access status via thrown error messages; (2) `isFieldReadableForPredicate`'s catch block is broader than its documented intent and swallows any error (not just the poisoned-item throw) as a silent denial; (3) a minor efficiency note on sequential vs. parallel field-access awaits in `collectFilterSpecs`.

---

borisno2:
All three findings from the automated review addressed in e6b0dc8:

1. **Check ordering (most severe).** The #912/#915 `where`/`orderBy` validation now runs only after the operation-level `query` access check confirms the caller has *some* access to the list. Previously it ran first, so a caller with zero list access could learn a field's name and read-gating status from the thrown validation error before the list-level denial was ever reached — a new information-disclosure vector this PR would otherwise have introduced. Added a regression test asserting a fully-denied caller gets the ordinary silent `[]`/`0` for both an undeclared key and a read-denied key.
2. **Overly broad catch.** `isFieldReadableForPredicate`'s catch now only treats the poisoned-item signal (`PredicateTimeItemAccessError`) as a denial; any other thrown error (a genuine bug in a `read` rule, or `InvalidFieldAccessResultError`) propagates instead of being silently folded into "not readable." Added a regression test.
3. **Sequential awaits.** `collectFilterSpecs` now runs its per-field read-access checks concurrently via `Promise.all` instead of one at a time in the loop.

All tests pass (core 1010, UI 563), lint/format/typecheck clean.

---

PR #926, opened 2026-08-16
feat(core): add bigInt() field type for 64-bit integers

### LINKED ISSUE #907: No `bigInt` field type — nothing in the stack can hold a millisecond epoch
## Current behavior

There is no field builder that produces a Prisma `BigInt` column. Neither existing candidate works for a 64-bit integer such as a millisecond epoch:

- `integer()` hardcodes `type: 'Int'` — 32-bit, overflows on `Date.now()`.
- `decimal()` reads back as an object whose `valueOf()` is a string, so arithmetic and comparison against a number don't work without conversion at every call site.

This blocks Better Auth's database-backed rate limiter, which stores `lastRequest` as a millisecond epoch.

## Investigation notes

- `packages/core/src/fields/index.ts:190` — `integer()`'s `getPrismaType` returns a hardcoded `{ type: 'Int' }`, and `getTypeScriptType` returns `number`.
- Line 231 of the same function comments `// Native type modifier (e.g., @db.SmallInt, @db.BigInt)`. That is misleading: `db.nativeType` only appends a modifier and the scalar stays `Int`, and `@db.BigInt` is not valid on a Prisma `Int` — so the documented escape hatch is a dead end.
- `decimal()` (line 318) maps to `@db.Decimal(p, s)`; the TypeScript type and the runtime `Decimal` object are the problem, not the storage.
- No `bigInt` export anywhere in `packages/core/src/fields/`.

## Suggested fix

Add a first-class `bigInt()` field builder: Prisma scalar `BigInt`, TypeScript `bigint`, with `validation` (`isRequired`, `min`, `max`) and `db` (`isNullable`, `map`, `defaultValue`) options matching `integer()`.

Decide and document the serialisation story at the UI and server-action boundaries, since `bigint` is not JSON-serialisable.

Separately, correct or remove the `@db.BigInt` example in `integer()`'s comment — it currently points at a workaround that cannot work.

## Downstream workaround

`apps/enrolm8/src/platform/lists/fields/big-integer.ts` implements this against the documented field self-containment contract (`getZodSchema` / `getPrismaType` / `getTypeScriptType`) — a useful reference for the shape, and to be deleted once a first-class `bigInt` lands.

---

_Reported from downstream use in `borisno2/enrolm8` (PR #601)._

borisno2:
## Summary

- Adds a first-class `bigInt()` field builder to `@opensaas/stack-core/fields` — Prisma `BigInt` scalar, TypeScript `bigint` — for values that overflow `integer()`'s 32-bit `Int` (e.g. a millisecond epoch, per the issue's motivating case in Better Auth's database-backed rate limiter).
- Write coercion accepts `bigint`, an integer `number`, or a numeric `string`, always resolving to `bigint`. A non-integer value or a `number` above `Number.MAX_SAFE_INTEGER` is rejected (not silently coerced) since it has already lost precision before the field sees it.
- `validation.min`/`max` (as `bigint`), `db.isNullable`/`map`/`nativeType`, `defaultValue`, and `isIndexed` all mirror `integer()`'s options; a `getFilterSpec` supports the comparison operators and degrades to free text on a non-integer token.
- **MCP**: `JSON.stringify` throws on a bare `bigint`. The MCP handler's response serializer now uses a replacer rendering `bigint` as a decimal string (the field's TS type stays `bigint` in application code — see ADR-0029), with a regression test covering the previously-uncaught `TypeError`.
- **Admin UI**: registers a `BigIntField` form component (text input, holds the raw string in local state like `JsonField` rather than parsing every keystroke like `IntegerField`, since `BigInt('-')` throws on an in-progress negative number) and reuses `IntegerCell` for the list-table Cell. Also fixes a real bug this surfaced: `ListView.tsx`, `RelationshipTable.tsx`, and `prepareItemForm.ts` each do an explicit `JSON.parse(JSON.stringify(...))` round-trip to cross the server→client boundary, which throws on `bigint` — a new shared `jsonSafeClone` helper round-trips a `bigint` back to a `bigint` instead.
- **RSC boundary investigation (per ADR-0029, "verify by observation")**: confirmed empirically against this repo's pinned React 19.2.8 / Next 16.2.10 that React's own Flight serializer supports `bigint` natively (tagged `"$n"` string, round-trips losslessly) — so no conversion was needed there. The `JSON.parse(JSON.stringify(...))` calls above run *before* Flight ever sees the value, which is the actual throw site; fixed as described.
- **Migration introspector**: Prisma `BigInt` columns now map to `bigInt()` instead of the previous lossy `text()` fallback (`prisma-introspector.ts`, `migration-generator.ts` warnings updated accordingly).
- Also fixes a pre-existing dead entry in `NUMERIC_FIELD_TYPES` (`'bigint'` lowercase, which never matched any real field-type discriminator) by adding `'bigInt'`.
- Docs added alongside `integer()`/`decimal()` in `docs/content/reference/fields-api.md` and `docs/content/concepts/field-types.md`, including the MCP wire-representation note; the Keystone migration field-mapping table updated.
- Changeset added (`@opensaas/stack-core`, `@opensaas/stack-ui`, `@opensaas/stack-cli` — minor).

## Test plan

- [x] Unit tests for `bigInt()`'s `getZodSchema`/`getPrismaType`/`getTypeScriptType`/`getFilterSpec` (`packages/core/tests/field-types.test.ts`, `packages/core/src/filter/filter.test.ts`)
- [x] MCP regression test asserting a `bigint`-carrying row serializes as a decimal string instead of throwing (`packages/core/tests/mcp-handler.test.ts`)
- [x] `BigIntField` component tests, including precision-preserving input beyond `Number.MAX_SAFE_INTEGER` (`packages/ui/tests/components/BigIntField.test.tsx`)
- [x] `jsonSafeClone` unit tests + a `prepareItemForm` integration test carrying a `bigInt` field value through unchanged (`packages/ui/tests/lib/`)
- [x] Prisma schema generation test for the `BigInt` column (`packages/cli/src/generator/prisma.test.ts`)
- [x] Introspector mapping/warning tests updated for `BigInt → bigInt()` (`packages/cli/tests/introspectors/prisma-introspector.test.ts`)
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format`, `pnpm build` (full monorepo, including the docs site's link-checker) all pass
- [x] Full test suites green: core (1014 tests), cli (348 tests), ui (576 tests)

Closes #907

---
_Generated by [Claude Code](https://claude.ai/code/session_01MfJ1v7cpvoaDUFza1hciKT)_


borisno2:
## Review summary

Ran a `/code-review` pass (medium effort) against this diff. One real finding surfaced and has been fixed:

**`packages/cli/src/generator/types.ts` — `context.db.<list>.create()/update()`'s write-`data` type override was narrowing a `bigInt()` field to bare `bigint`, rejecting the `number` writes both Prisma's own generated input type and `bigInt()`'s own `getZodSchema` accept.** Same failure mode `decimal` is already exempted from (`shouldNarrowScalarWrite`), just missing for `bigInt`. Fixed in 688dbef by adding `bigInt` to the exemption list, with compile-time coverage added to the existing `types-write-narrowing.test.ts` tsc fixture.

While fixing it I verified empirically (generated a real Prisma client against a scratch `BigInt` schema) that Prisma's own input type for `BigInt` is `bigint | number` — no `string`, unlike `Decimal`'s `Decimal | number | string`. So the exemption gives full write-path parity for `number`/`bigint`; a numeric-`string` write via the strictly-typed `context.db.<list>.create()` call site specifically still needs a `bigint` literal or a less-strictly-typed boundary (server action, MCP) — `bigInt()`'s own Zod validation still accepts and coerces a string there. Documented this nuance in the code comment.

One suspected issue was investigated and ruled out: whether `bigint` values crossing the admin UI's React Server→Client props boundary would break. Confirmed empirically against this repo's pinned React 19.2.8 / Next 16.2.10 that Flight's serializer supports `bigint` natively — the actual throw site was this package's own pre-existing `JSON.parse(JSON.stringify(...))` round-trips (`ListView.tsx`, `RelationshipTable.tsx`, `prepareItemForm.ts`), which is what the new `jsonSafeClone` helper in this PR fixes.

## CI

The `test` job's browser-test step failed with a wall of `EMFILE: too many open files` errors while vite pre-transformed `packages/ui/dist/index.js` — a runner file-descriptor exhaustion before any test body ran, unrelated to this diff. Re-ran the failed job.


---

Issue #749, opened 2026-07-19
Admin UI: nested relationship label filter is not access-scoped on the related list


borisno2:
## Parent

Follow-up from #728 (Generic reusable admin UI). Surfaced during review of #746 (filter engine, #730) — filed as non-blocking so #746 could merge. Low priority / by-design-for-now.

## Observation

The relationship Filter spec maps a token like `author:Ada` to a nested Prisma condition, e.g. `{ author: { is: { name: { contains: 'Ada' } } } }`, which runs as part of the parent list's query. Top-level access on the parent list still applies (so the session only ever sees rows it's already allowed to see), but the **nested condition on the related list is not itself filtered by the session's access to that related list**.

This is inherent to Prisma nested relation filtering and matches Keystone's behaviour. Because the filter can only narrow the already-access-scoped parent result set, the exposure is marginal — a session could, in principle, distinguish parent rows by a related-record attribute it might not be allowed to read directly. Note the asymmetry: relationship label *lookups* in the suggestions metadata **are** routed through the access-controlled path, so this nested-filter path is the one place that isn't.

## Possible directions (to evaluate, not prescribe)

- Resolve the related-list access filter for the session and AND it into the nested condition (`{ author: { is: { AND: [<relatedAccessFilter>, { name: { contains } }] } } }`).
- Or document the boundary explicitly as accepted behaviour consistent with Keystone.

## Acceptance criteria

- [ ] Decide: access-scope the nested relationship filter, or document it as accepted behaviour with rationale
- [ ] If scoping: nested relationship filter conditions are ANDed with the related list's access filter for the session; test covers a session that can see a parent row but not the related record's filtered field
- [ ] If documenting: the boundary is recorded (CONTEXT.md / ADR note) so it's an explicit, reviewed decision

## Blocked by

None. Independent of the rest of #728's remaining tickets.

PR #929, opened 2026-08-16
fix(core,ui): scope relation filters in `where` by the related list's query access

### LINKED ISSUE #916: Relation filters in `where` bypass the related list's `query` access, unlike `include`

borisno2:
## Summary

A relation filter in a caller's `where` (`some`/`every`/`none`/`is`/`isNot`) was handed to Prisma unscoped, evaluated against every row of the related table regardless of the caller's access to that list. `include` was already scoped this way (`buildAccessScopedInclude`, ADR-0022) — this was the unclosed `where` half, and it composes across multi-hop chains, letting an anonymous caller confirm arbitrary facts (down to identity-confirmation via a chain reaching a `User` list) by varying a `count()`/`findMany()` predicate.

- Added `buildAccessScopedWhere` (`packages/core/src/access/access-filter.ts`), the `where` counterpart to `buildAccessScopedInclude`. It recurses through `AND`/`OR`/`NOT` and every hop of a relation chain, and for each relationship key:
  - Evaluates the related list's `query` access.
  - **Denied (`false`) → throws** `RelationFilterAccessDeniedError` — a loud failure, not a silently-empty match (ADR-0022; a `where` predicate has no neutral "not requested" outcome the way a missing `include` key does).
  - **Filter** → AND-folds it into the relation quantifier's nested clause (never replacing the caller's own condition).
  - **`true`** → passes the caller's filter through unperturbed.
  - Applies the related list's field-level `read` access to keys named inside the filter (closing the same gap #915 closed for the current list's own fields).
- Reuses existing primitives rather than reimplementing the where-shape walk: `resolveQueryField`/`LOGICAL_OPERATORS`/`RELATION_QUANTIFIERS`/`walkWhereReadAccess` (exported from `query-validation.ts`), and `checkAccess`/`andWhere` (the same ones `buildAccessScopedInclude` uses).
- Wired into `createFindMany` and `createCount` in `context/index.ts`, after the existing #912/#915 checks, inside the `!context._isSudo` branch (sudo unaffected).
- `every`'s AND-fold is documented as deliberately conservative rather than exactly precise (it can reveal that an inaccessible related row exists, never that row's field values) — the safe direction, since a precise fix needs a `NOT`/`OR` transform this doesn't attempt.
- `resolveRelationshipLabelFilters` (admin list view, #749) folded the same kind of access filter into a to-one relationship's `{ is: {...} } }` label filter by hand; that's now exactly the shape the engine scopes automatically, so its fold is redundant. It's now a pass-through (kept exported, unchanged signature, for API compatibility) and `ListView.tsx` no longer calls it. `resolveRelationshipCountFilters` (#732) is untouched — it folds access into a `_count.select`, a distinct Prisma feature the new `where`-relation-filter scoping doesn't touch.

## Test plan

- [x] New `context.test.ts` suite `relation filter access scoping (#916)`: single-hop denial (findMany + count), no-oracle count-probe pairs (matching vs. non-matching, and the exact two-hop identity-confirmation case from the issue), access-filter folding into a relation quantifier, pass-through when fully readable, sudo bypass, and the "zero list access → silent denial, not the new error" ordering invariant.
- [x] Flipped the pre-existing `context.test.ts` test that pinned the old (buggy) behavior — a read-denied field on a related list nested in a relation filter is now rejected.
- [x] Rewrote `relationship-label-filter.test.ts` to pin the new pass-through behavior.
- [x] Updated `ListView.test.tsx`'s label-filter suite to match (ListView now hands the unscoped filter to `context.db`; the real scoping is exercised in `context.test.ts`).
- [x] `pnpm build` (core, ui, cli)
- [x] `pnpm test` — all packages green (core 1043, ui 579, auth 217, cli 353, rag 178, storage 167, storage-s3 43, storage-vercel 56, create-opensaas-app 41)
- [x] Changeset added (`@opensaas/stack-core` + `@opensaas/stack-ui`, patch)

Closes #916

---
_Generated by [Claude Code](https://claude.ai/code/session_01VZ8TTn9E8VKvGFRxApm7zW)_

### DIFF
--- a/.changeset/tiny-hounds-relate.md
+++ b/.changeset/tiny-hounds-relate.md
@@ -1,0 +1,6 @@
+---
+'@opensaas/stack-core': patch
+'@opensaas/stack-ui': patch
+---
+
+Fix: a relation filter in `where` (`some`/`every`/`none`/`is`/`isNot`) no longer bypasses the related list's `query` access — it is now scoped exactly like `include` already is, recursing through every hop of a chain, on both `findMany` and `count`. A filter through a related list that denies query access now throws `RelationFilterAccessDeniedError` instead of silently running unscoped; field-level `read` access on the related list also now applies to keys named inside the filter. `@opensaas/stack-ui`'s admin list view no longer needs its own relationship label-filter access fold, since the engine now covers it.
--- a/packages/core/src/access/access-filter.ts
+++ b/packages/core/src/access/access-filter.ts
@@ -1,8 +1,14 @@
 import type { Session, AccessContext, PrismaFilter } from './types.js'
-import type { OpenSaasConfig, FieldConfig } from '../config/types.js'
+import type { OpenSaasConfig, FieldConfig, ListConfig } from '../config/types.js'
 import { checkAccess, getRelatedListConfig } from './engine.js'
 import { READ_INCLUDE_MAX_DEPTH } from './depth-limits.js'
-import { AccessScopeDepthExceededError } from './errors.js'
+import { AccessScopeDepthExceededError, RelationFilterAccessDeniedError } from './errors.js'
+import {
+  LOGICAL_OPERATORS,
+  RELATION_QUANTIFIERS,
+  resolveQueryField,
+  walkWhereReadAccess,
+} from './query-validation.js'
 
 /**
  * Access Filter — phase 1 of the two-phase read (pre-query).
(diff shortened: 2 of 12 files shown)


borisno2:
CI status: `test`, `e2e`, `coverage`, and GitGuardian are all green.

The only red status is the `Vercel` (stack-docs) deployment, which failed with `Resource is limited - try again in 24 hours (more than 100, code: api-deployments-free-per-day)` — an account-wide daily deployment quota on Vercel's free tier, unrelated to this PR's diff. Not something a code push can fix; it should clear on its own once the quota resets.

---

PR #931, opened 2026-08-16
fix(core,cli): db.indexes supports single-field entries; new empty-array and isIndexed-collision errors

### LINKED ISSUE #918: core: single-field `db.indexes` entries work but are documented as forbidden — so a single-column unique can't be named

## Current behavior

*(Body corrected during triage — this issue was filed on the premise that naming a single-column unique constraint required a new API. It does not.)*

There appear to be two ways to declare an index, neither able to name a single-column unique:

- Field-level `isIndexed` is typed `boolean | 'unique'`. It takes no name override.
- List-level `db.indexes` supports an explicit `name` (Prisma's `map:`), but its type doc comment and `CLAUDE.md` both say each entry must name **two or more** fields.

**The two-field minimum is not enforced anywhere.** The index emitter maps over an entry's `fields` with no length check, and nothing in core or the CLI validates `db.indexes` arity. A single-field entry generates correct Prisma today, `map:` included — it is simply undocumented, untested, and forbidden by prose.

Also worth recording, since it is not obvious: field-level `isIndexed` does not emit a field-level `@unique`. It emits model-level `@@unique([field])` / `@@index([field])`. Both surfaces converge on the same generated line.

## Why it matters

ADR-0007's promise is that a project with pre-existing tables can model them and reach schema parity **without a destructive migration** — the generated schema diffs clean against the live database. A live table's unique constraint carries whatever name its own migration produced, which need not match Prisma's derived name. Every column can line up and the constraint still diffs.

Surfaced by #909, where better-auth's limiter table declares `key` as `unique: true` and the derived list must reproduce it. The constraint itself is load-bearing there — the limiter races concurrent requests into a create and relies on the violation to serialise them — so it cannot be omitted to sidestep the naming question.

The practical cost of the incorrect prose is already visible: it shaped #909's out-of-scope section around a limitation that was not real.

## Suggested fix

Correct the documentation to describe one-or-more fields, pin the single-field behaviour with tests, and add the two errors the surface is genuinely missing — an empty `fields` array, and a `db.indexes` entry colliding with a field-level `isIndexed` on the same column.

No new API. In particular, no object form on `isIndexed` — that would be a second spelling for something that already works.

See the agent brief below for the full specification.

## Scope note

Split out of #909's triage, and out of scope there. Threading a constraint-name override through `authPlugin`'s derived lists — so #909's derived limiter list can actually use this — is a separate follow-up against #909's model config, to be raised once this lands.

borisno2:
## Summary

- Corrects the `ListIndex`/`db.indexes` doc comments (core `types.ts` and root `CLAUDE.md`), which wrongly claimed an entry must span **two or more** fields. A single-field entry already generated correct Prisma (including `map:`) on `main` — it was simply undocumented, untested, and forbidden by prose. Docs now describe the real rule: `isIndexed` is sugar for the unnamed single-column case, `db.indexes` is the full form for a name, a `sort` direction, or a composite constraint, and arity is incidental.
- Adds the two generation-time errors the surface was genuinely missing, each naming the list and the entry:
  - An entry whose `fields` array is empty (previously produced invalid Prisma, `@@unique([])`/`@@index([])`, surfaced only downstream by `prisma validate`).
  - A single-field entry that indexes the exact column a field-level `isIndexed` on the same list already indexes — the error also names the colliding field and its `isIndexed` value, since one of the two should be removed.
- Adds a `db.indexes` reference-docs section (`docs/content/reference/config-api.md`) and cross-links from every `isIndexed` reference entry (`docs/content/reference/fields-api.md`).
- No new API — no object form on `isIndexed`, no signature changes to `ListIndex`/`ListIndexFieldRef`/`isIndexed`.

## Test plan

- [x] New tests in `packages/cli/src/generator/prisma.test.ts`: single-field entry with `unique`+`name`, `unique` with no name, plain `@@index`, a `sort` direction, and a relationship field's FK column resolved through a single-field entry.
- [x] New tests for the empty-`fields` error, and for the `isIndexed` collision error covering both the `'unique'` and plain-`true` forms (scalar and relationship-FK-default cases), plus a non-collision control case.
- [x] Existing composite-index and error-path tests unchanged and passing.
- [x] `pnpm test` green in `packages/cli` (363 tests) and `packages/core` (1043 tests).
- [x] `pnpm build` green in `packages/core`, `packages/cli`, `packages/auth`.
- [x] `pnpm lint` clean (pre-existing unrelated warnings only) and `pnpm format` / `pnpm manypkg fix` applied.
- [x] Changesets added via the `pr-changeset` skill (patch on `@opensaas/stack-core`, patch on `@opensaas/stack-cli`).

Closes #918

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vag2xhpy1CViebAasA27Fx)_

### DIFF
--- a/.changeset/tame-plums-heal.md
+++ b/.changeset/tame-plums-heal.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-cli': patch
+---
+
+`db.indexes` generation now fails with a descriptive error for an empty `fields` array, and for a single-field entry that duplicates a column already indexed by that field's own `isIndexed` — previously these silently produced invalid or duplicate Prisma.
--- a/.changeset/tiny-plums-listen.md
+++ b/.changeset/tiny-plums-listen.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Correct `ListIndex`/`db.indexes` doc comments, which wrongly claimed an entry must span two or more fields — a single-field entry is fully supported and now documented as such.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -435,9 +435,9 @@ The function receives:
 
 Field-level `extendPrismaSchema` is applied before the global `db.extendPrismaSchema`, allowing both granular and broad modifications.
 
-**Model-Level Composite Indexes (`db.indexes`):**
+**Model-Level Indexes (`db.indexes`):**
 
-Field-level `isIndexed` (on a scalar or relationship field) can only ever produce a single-column `@@unique`/`@@index`. A list's `db.indexes` is the multi-column case: each entry names two or more of the list's own OpenSaaS field names (not raw database column names) and the generator resolves each to its Prisma column — a scalar field's own name, or a relationship field's foreign key column (`<field>Id`) when this side owns it.
+Field-level `isIndexed` is the sugar for the unnamed single-column case. `db.indexes` is the full form — reach for it when a constraint needs a `name` (Prisma's `map:`, for adopting an existing live constraint), a `sort` direction, or spans more than one column. Arity is incidental: each entry names one or more of the list's own OpenSaaS field names (not raw database column names) and the generator resolves each to its Prisma column — a scalar field's own name, or a relationship field's foreign key column (`<field>Id`) when this side owns it. See the [reference docs](https://stack.opensaas.au/docs/reference/config-api#dbindexes) for the full rules.
 
 ```typescript
 Audition: list({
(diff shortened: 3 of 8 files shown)


borisno2:
The "Vercel" status failure on this PR (`stack-docs` deployment) is a Vercel free-plan daily rate limit (`api-deployments-free-per-day`, "try again in 24 hours") — unrelated to this diff and not fixable from the code side. Waiting on the real `test`/`e2e` GitHub Actions checks, which are still in progress.


---

PR #946, opened 2026-08-16
fix(core): restore two comments trimmed too far in the context/ cleanup

### LINKED ISSUE #936: Comment cleanup: worked examples doc, calibrated on core/context
## What to build

The Comments rule landed in `CLAUDE.md` in #933 but nothing shows what applying it actually looks like. This ticket produces that reference **by doing the hardest pass first**, so the examples are real diffs rather than invented ones.

Two outputs, in this order:

1. Clean the comments in `packages/core/src/context/` — 6 files, ~601 line comments, the densest concentration in the repo. It has all four shapes in one place: restatement (`// Skip virtual fields`, `// Check if it's a Prisma error`), multi-paragraph rationale blocks citing issues and ADRs (see the type-member comments near the top of `context/index.ts`), keeper `eslint-disable ... --` directives, and genuine footgun warnings.
2. Write `docs/agents/comments.md` from what that pass taught you — 6–8 before/after pairs drawn from your own diff, linked from the Comments section of `CLAUDE.md`.

The doc is the point. Every other ticket in this set reads it before starting, and it is what stops seven agents making seven different calls on the same rationale block.

**The judgement calls it must settle**, because they recur in every later batch:

- A rationale block that cites an issue or ADR: when does it become a one-line link, and when does it go entirely because the ADR already says it?
- Public-API TSDoc on exported config options, field builders and plugin surfaces: **kept** (`CLAUDE.md` names this exception). Show where the boundary sits between contract documentation and narration of the implementation beneath it.
- A warning where the obvious edit is wrong: **kept**. Show one.
- File headers: 63 files repo-wide open with a `/** */` block. Show a trim and a deletion.

## Scope

- `packages/core/src/context/` only.
- **Test files are out of scope** across this whole ticket set — leave `*.test.ts` / `*.test.tsx` alone.
- Comments only. No behaviour changes, no renames, no refactors folded in.

## Acceptance criteria

- [ ] `packages/core/src/context/` comment cleanup complete, following the `CLAUDE.md` Comments rule
- [ ] `docs/agents/comments.md` exists with 6–8 real before/after pairs from this diff
- [ ] The Comments section of `CLAUDE.md` links to it
- [ ] Public-API TSDoc, `eslint-disable` reasons and genuine footgun warnings are demonstrably preserved
- [ ] No test files touched
- [ ] `pnpm lint`, `pnpm format` and the core test suite pass unchanged — a comment-only change must not move a single test
- [ ] Patch changeset added (repo rule: any `packages/*` change needs one)

## Blocked by

None — can start immediately.

borisno2:
## Summary

Follow-up to #945 (already merged), from that PR's own `/code-review` pass, which finished after auto-merge had already landed the PR:

- `packages/core/src/context/nested-operations.ts`: `capturePreExistingIds`'s docblock dropped the constraint that its `prisma` parameter must be the transaction client — without it, a future caller passing the outer non-transactional client would silently break the pre-persist read's consistent-snapshot guarantee (the id-diff used to identify newly-created nested rows would race the enclosing transaction). Restored as a one-line note.
- `CONTEXT.md`: the "Write Pipeline" glossary entry stated the phase order as `hooks → validation → operation-level access → ...`, but the pipeline's own pre-transaction gate (`#590`, see the comment in `write-pipeline.ts`) resolves operation-level access *first*, outside the transaction, before any hook fires — a denied write short-circuits to `null` before `resolveInput`/`validate` ever run. `write-pipeline.ts`'s docblock was condensed in #945 to point at this glossary entry instead of restating the order inline, so the stale order went from redundant-but-harmless to actively misleading. Fixed the order and added the `#590` short-circuit note.

No behavior changes — comment/doc text only.

## Test plan

- [x] `pnpm lint` — clean (2 pre-existing warnings elsewhere, unrelated)
- [x] `pnpm format` — clean
- [x] `cd packages/core && pnpm test` — 1054 tests passed, 0 changed

No changeset — comment/doc-only fix on top of #945's own patch changeset; nothing package-behavior-relevant changed.

Closes #936

---
_Generated by [Claude Code](https://claude.ai/code/session_01JS2L5jj1uEkNgcCUuWB4wy)_

### DIFF
--- a/.changeset/tame-otters-listen.md
+++ b/.changeset/tame-otters-listen.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Restore two comments in `src/context/` that were trimmed too far in a prior comment cleanup (#945). No behavior changes.
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -59,7 +59,7 @@ The convention that an access-denied operation returns `null` (single) or `[]` (
 _Avoid_: access error, permission error
 
 **Write Pipeline**:
-The single module that runs the canonical, secured write sequence (hooks → validation → operation-level access → writable-field filtering → nested operations → persistence → after-hooks → Field Visibility) for one create/update/delete. Owns the phase order in one place; per-operation differences (target resolution, which input phases run, the database verb and returned row) are supplied by a per-operation strategy.
+The single module that runs the canonical, secured write sequence (operation-level access → hooks → validation → writable-field filtering → nested operations → persistence → after-hooks → Field Visibility) for one create/update/delete. Operation-level access is resolved first, outside the transaction (#590) — a denied write short-circuits to `null` before any hook fires. Owns the phase order in one place; per-operation differences (target resolution, which input phases run, the database verb and returned row) are supplied by a per-operation strategy.
 _Avoid_: operation handler, mutation service
 
 **Hook Pipeline**:
--- a/packages/core/src/context/nested-operations.ts
+++ b/packages/core/src/context/nested-operations.ts
@@ -150,6 +150,9 @@ function createCreatedRowRecovery(
  * Capture the ids of the rows currently linked to the parent via `fieldName`,
  * BEFORE the parent persists, so created rows can later be identified by id-diff.
  * Empty for a parent CREATE (no parent row exists yet to have related rows).
+ *
+ * `prisma` must be the transaction client — the read has to participate in the
+ * transaction to see a consistent snapshot.
  */
 async function capturePreExistingIds(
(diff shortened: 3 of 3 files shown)


borisno2:
The only failing status is `Vercel` (`Deployment rate limited — retry in 24 hours` — the docs-site project has hit its free-tier daily deployment quota). This is an external platform quota, not caused by this diff — the same failure appeared on #945 and didn't block that PR's merge. GitGuardian's check passed; `pnpm lint`, `pnpm format`, and the core test suite (1054 tests) all pass locally. No action needed on this PR for it.

---

PR #947, opened 2026-08-16
chore(core): comment cleanup across access, config, fields, filter, hooks, mcp, query, lib, utils, validation

### LINKED ISSUE #938: Comment cleanup: packages/core (access, fields, config, hooks, mcp, query, filter)
## What to build

Apply the Comments rule from the `## Comments` section of `CLAUDE.md` to the area below.

**Read `docs/agents/comments.md` first** (produced by #936). It carries the worked before/after pairs and settles the recurring judgement calls — when a rationale block citing an issue or ADR becomes a one-line link versus goes entirely, and where the boundary sits between keeper public-API TSDoc and narration of the implementation beneath it. Do not re-derive those calls; follow them.

**Delete:** comments restating the line beneath them, docblocks on self-describing constants, notes explaining why something is *absent*, commented-out code, and file headers that run longer than a line or two.

**Keep:** TSDoc on exported config options, field builders and plugin surfaces — that is the consumer's editor documentation and a public contract. Also `eslint-disable ... --` justifications, external-constraint notes (Prisma behaviour, Better-auth contracts, Next.js boundaries, upstream bugs), and warnings where the obvious edit is the wrong one.

**Rationale does not move into the code — it moves out of it.** If a comment is carrying reasoning worth keeping, its home is an ADR, the issue, or the PR body, not a docblock.

**Area:** all of `packages/core/src` except `context/` (done in #936) — roughly 1,036 line comments across ~40 source files.

The two densest spots are `access/` (~339, including the field-visibility and field-access engines) and `fields/index.ts` (~231). `fields/index.ts` is where the public-API exception bites hardest: the field builders are the surface users see in their editors, so their TSDoc stays while the narration inside the builder bodies goes. `config/types.ts` is almost entirely public contract — expect to delete very little there.

This is the largest ticket in the set. If it does not fit one context window, split it at the `access/` boundary and say so on the issue rather than half-finishing.

## Scope

- Comments only. No behaviour changes, no renames, no adjacent refactors folded in.
- **Test files are out of scope** — leave `*.test.ts` / `*.test.tsx` alone.

## Acceptance criteria

- [ ] Comment cleanup complete for the area named above
- [ ] Public-API TSDoc, `eslint-disable` reasons and genuine footgun warnings preserved
- [ ] No test files touched
- [ ] `pnpm lint`, `pnpm format` and the affected package's test suite pass unchanged — a comment-only change must not move a single test
- [ ] Patch changeset added (repo rule: any `packages/*` change needs one)

## Blocked by

- #936 — the worked-examples doc calibrates every call in this ticket.

borisno2:
## Summary

Applies the `CLAUDE.md` Comments rule to all of `packages/core/src` except `context/` (cleaned up in #945/#946), following the judgement calls calibrated in `docs/agents/comments.md` (produced by #936):

- Deleted comments that fail the **staleness** or **restatement** tests: restated function/variable names, per-field boilerplate repeated across sibling field builders (`text`, `integer`, `decimal`, etc.), duplicate rationale blocks re-explained at every call site instead of stated once and pointed to.
- Kept every `eslint-disable ... --` justification verbatim.
- Kept public-API TSDoc on exported field builders (`fields/index.ts`), config builders (`config()`/`list()`), the `AccessControl`/`FieldAccessControl` contract types (`access/types.ts`), and plugin surfaces — trimmed only the narration inside their implementation bodies.
- Kept genuine outside-constraint notes and footgun warnings (e.g. the fail-closed default-deny comment in `access/engine.ts`, the bulk-action error-leak warning pattern, ADR/issue-citing correctness rationale in the access-control engine files).
- `config/types.ts` (~3,200 lines, almost entirely public config-authoring contract) was pruned conservatively per the issue's explicit guidance — only ~14 items removed/condensed against several hundred docblocks left untouched.
- Along the way, fixed three comments that had gone stale relative to the code they sit beside (corrected the text, not deleted): `BaseFieldConfig.virtual`'s doc contradicted ADR-0027 on when virtual fields compute; `PluginContext.addList`/`extendList` docs described a merge strategy that doesn't exist (the engine actually throws per ADR-0013); and an `OpenSaasConfig` docblock had been orphaned above the wrong interface by an earlier edit.

`access/` and `fields/index.ts` were the two densest areas named in the issue; `config/types.ts` was handled with the "delete very little" guidance the issue calls out explicitly.

## Test plan

- [x] `pnpm build` (core) — clean, no type errors
- [x] `pnpm test` (core) — 1054/1054 tests pass, unchanged
- [x] `pnpm lint` — clean (2 pre-existing warnings unrelated to this change)
- [x] `pnpm format` — no changes needed
- [x] `pnpm manypkg fix` — clean
- [x] No test files touched (`git diff --stat -- '*.test.ts' '*.test.tsx'` is empty)
- [x] Patch changeset added for `@opensaas/stack-core`

Closes #938

---
_Generated by [Claude Code](https://claude.ai/code/session_01LfZLYh8qhK3cEYZAVRFHoH)_

### DIFF
--- a/packages/core/src/config/label.ts
+++ b/packages/core/src/config/label.ts
@@ -1,8 +1,11 @@
 import type { ListConfig } from './types.js'
 
 /**
- * Resolve the field name used as a list's Label field — see the "Label
- * field" glossary entry in `CONTEXT.md`.
+ * Resolve the field name used as a list's Label field: `ui.labelField` if
+ * set, else `name`, else `title`, else `id`.
+ *
+ * @throws if `ui.labelField` names a field not declared on the list, or a
+ * relationship field (the Label field must be a scalar).
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
 export function getLabelFieldName(listConfig: ListConfig<any>): string {
@@ -29,8 +32,8 @@ export function getLabelFieldName(listConfig: ListConfig<any>): string {
 }
 
 /**
- * Render a row's Label field as text — see the "Item label" glossary entry
- * in `CONTEXT.md`.
+ * Render a row's Label field ({@link getLabelFieldName}) as text, falling
+ * back to `item.id` when the value is `null`/`undefined`.
  */
 export function getItemLabel(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
--- a/packages/core/src/config/label.ts
+++ b/packages/core/src/config/label.ts
@@ -1,16 +1,8 @@
 import type { ListConfig } from './types.js'
 
 /**
- * Resolve the field name that represents a list's rows as a single label —
- * the projection half of the label seam. `getItemLabel` reads exactly the
- * field this returns, so the field chosen for projection can never drift
- * from the field used for rendering.
- *
- * Fallback order: configured `ui.labelField` → `name` → `title` → `id`
- * (first field that exists on the list).
- *
- * @throws if `ui.labelField` is set but does not reference a declared,
- * non-relationship field on the list.
+ * Resolve the field name used as a list's Label field — see the "Label
+ * field" glossary entry in `CONTEXT.md`.
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
 export function getLabelFieldName(listConfig: ListConfig<any>): string {
@@ -37,10 +29,8 @@ export function getLabelFieldName(listConfig: ListConfig<any>): string {
 }
 
 /**
- * Resolve the display label for a single row — the render half of the label
- * seam. Reads the field `getLabelFieldName` resolves, falling back to `id`
- * when that field is missing from the row (e.g. stripped by field-level
- * access).
+ * Render a row's Label field as text — see the "Item label" glossary entry
+ * in `CONTEXT.md`.
  */
 export function getItemLabel(
   // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
--- a/.changeset/quiet-otters-listen.md
+++ b/.changeset/quiet-otters-listen.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Clean up comments across `access/`, `config/`, `fields/`, `filter/`, `hooks/`, `lib/`, `mcp/`, `query/`, `utils/` and `validation/` per the CLAUDE.md Comments rule. No behavior changes.
--- a/packages/core/src/access/access-filter.ts
+++ b/packages/core/src/access/access-filter.ts
@@ -69,8 +69,7 @@ function asEntryObject(value: unknown): IncludeEntryObject | null {
  *
  * The access filter is authoritative: the caller's filter may only NARROW the
  * result further, never widen past what access permits. We therefore wrap both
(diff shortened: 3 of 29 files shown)


packages/core/src/config/label.ts
```diff
@@ -1,16 +1,8 @@
...
- * from the field used for rendering.
- *
- * Fallback order: configured `ui.labelField` → `name` → `title` → `id`
- * (first field that exists on the list).
- *
- * @throws if `ui.labelField` is set but does not reference a declared,
- * non-relationship field on the list.
+ * Resolve the field name used as a list's Label field — see the "Label
+ * field" glossary entry in `CONTEXT.md`.
  */
```
borisno2, line 9: **[conventions] Public API docblock now points to a repo-only doc that ships-consumers never see**

`getLabelFieldName`/`getItemLabel` are exported from the package root (`packages/core/src/index.ts`), so per this repo's root `CLAUDE.md`: *"Public API docblocks are the deliberate exception... it is what the consumer's editor shows — that is a contract."*

The old docblocks were self-contained: fallback order (`ui.labelField` → `name` → `title` → `id`) and the `@throws` contract for `getLabelFieldName`; the `id`-fallback behavior for `getItemLabel`. The new docblocks replace all of that with "see the ... glossary entry in `CONTEXT.md`" — but `CONTEXT.md` lives at the repo root, not inside `packages/core`, so it does not ship with the published `@opensaas/stack-core` npm package. A consumer hovering `getLabelFieldName` in their editor now sees no fallback order and no indication the function throws on a bad `ui.labelField`, with no way to reach the doc that used to explain it.

Failure scenario: an app author sets `ui.labelField: 'author'` on a list where `author` is a relationship field. `getLabelFieldName` throws `"ui.labelField \"author\" must reference a scalar field, not a relationship."` — but the published type's docblock gives no hint this call can throw at all, so the author has no reason to wrap it or notice the constraint until it fails at runtime.

_Single-pass review without the Agent tool (no multi-agent fan-out / subagent verify pass) — this PR is otherwise a clean, verified-safe comment-only cleanup with no behavior changes found._

---

borisno2 (commented):
Single-pass code review (high effort, Agent tool unavailable — no multi-agent fan-out/verify pass ran). PR #947 is a comment-only cleanup enforcing the root CLAUDE.md "Comments: default to none" rule across `access/`, `config/`, `fields/`, `filter/`, `hooks/`, `lib/`, `mcp/`, `query/`, `utils/`, `validation/`. I read every hunk plus the enclosing functions; no logic, control flow, or behavior changed anywhere — the changeset's "No behavior changes" claim checks out. Most trims that condense a "why" comment (declared-dependencies.ts cycle guard, relationship-count.ts sibling-preservation note) still carry the reasoning in a referenced module-doc comment or sibling function docblock, so nothing load-bearing was lost there.

One finding survived verification, posted inline: `packages/core/src/config/label.ts`'s two exported docblocks (`getLabelFieldName`, `getItemLabel`) were replaced with pointers to `CONTEXT.md`, a repo-root doc that doesn't ship with the published npm package — losing the fallback order and `@throws` contract from what a consumer's editor actually shows for this public API.

---

packages/core/src/config/label.ts
borisno2: Good catch — fixed in 25cd93d. Restored self-contained docblocks with the fallback order (`ui.labelField` → `name` → `title` → `id`) and the `@throws` contract inline, rather than pointing at `CONTEXT.md`. Core build/tests still pass (1054/1054).

---

PR #960, opened 2026-08-17
fix(cli): remove dead prisma-extensions.ts and its unreachable $extends branch

### LINKED ISSUE #956: TS2589 still occurs post-0.39.1 fix — second cause: createExtendedPrisma's inferred return type in generated context.ts
Follow-up to #952. The 0.39.1 fix (converting `CustomDB`/`BaseContext`/`Context` from `type` to `interface` to stop `Context.sudo()`'s self-reference forcing eager expansion) is correct and I can confirm it's present in generated output. But `TS2589: Type instantiation is excessively deep and possibly infinite` still reproduces once a schema's model count reaches ~13, independent of that fix.

**Repro:** any config with 13 registered lists (independent of field shape — reproduced with a list that has just 2 `relationship()`s, `text()`, `timestamp()`, `select()`, no `json()` fields). `rm tsconfig.tsbuildinfo && pnpm opensaas generate && tsc --noEmit` fails:
```
.opensaas/context.ts(45,73): error TS2589: Type instantiation is excessively deep and possibly infinite.
```
Line 45 is `if (process.env.NODE_ENV !== 'production') globalForPrisma.prisma = prisma`.

**Cause:** `dist/generator/context.js`'s static template for `createExtendedPrisma` has no return-type annotation:
```ts
function createExtendedPrisma(basePrisma: PrismaClient) {
  if (Object.keys(prismaExtensions).length === 0) {
    return basePrisma
  }
  return basePrisma.$extends(prismaExtensions)
}
```
TypeScript infers the return type as the union of `PrismaClient` and `ReturnType<typeof basePrisma.$extends>`. `$extends`'s generic signature is expensive to instantiate against a `PrismaClient` this wide, and doing so during inference is what exhausts the instantiation budget a few lines later — this is unrelated to `CustomDB`/`Context`/`GetPayload`, it's plain Prisma Client extension typing cost, triggered purely by model count.

**Confirmed fix:** annotating the function `: PrismaClient` (with `as unknown as PrismaClient` on the `$extends` branch, since `prismaExtensions` is `{}` whenever no list configures a `resultExtension` and that branch is dead at runtime) makes `tsc --noEmit` clean at 13, and stays clean when synthetically pushed to 15 models — so it's not just moving the ceiling.

**Suggested upstream fix:** add an explicit return-type annotation to the generated `createExtendedPrisma`, computed from whether `resultExtension`s are actually configured (e.g. `PrismaClient` when there are none, or the extended type only when there's something to extend), rather than relying on inference across both branches unconditionally.

We've applied a local `afterGenerate` patch in our app to unblock development in the meantime (post-processes the generated `context.ts` to add the return-type annotation) — happy to share the exact patch or a minimal repro schema if useful.

### LINKED ISSUE #958: Generated Prisma result extensions are never applied — the guard in the generated context is always true

Found while investigating #956 / #957. The generated context decides whether to apply result extensions with:

```ts
if (Object.keys(prismaExtensions).length === 0) {
  return basePrisma
}
return basePrisma.$extends(prismaExtensions)
```

`Prisma.defineExtension(...)` returns a **function**, and `Object.keys(someFunction)` is `[]`. So the guard is `true` in every generated app — including apps that do configure virtual fields or `resultExtension`s — and `$extends` is never called. Verified on `examples/blog`:

```
typeof prismaExtensions: function
Object.keys(...).length: 0
=> generated guard `length === 0` evaluates to: true
```

The no-extensions case emits `export const prismaExtensions = {}`, which also has zero keys — so that branch is right by accident, and the two cases are indistinguishable at runtime.

## Runtime confirmation

`examples/blog` configures both a `password` field (`resultExtension`) and a `displayName` virtual field. I instrumented every branch of every generated `compute` with counters, seeded a user, and read it back:

```
after ctx.prisma read: {"preResolution":0,"resolved":0,"nullPass":0}
password value: PLAINTEXT-MARKER | ctor: String
displayName: undefined
via ctx.db password ctor: HashedPassword
```

No `compute` ran, on any path. `context.prisma` returns the raw column and no virtual field. `context.db` is correct — because core's read pipeline (`field-visibility.ts`) runs `resolveOutput` itself, with a real context and the resolve-chain cycle protection from ADR-0023.

The only consumer of `prismaExtensions` anywhere in the repo is that dead branch.

## Why it matters

- The generated `prisma-extensions.ts` module is dead weight in every app: generated, imported, type-checked, bundled, never executed.
- It is a second, parallel implementation of `resolveOutput` dispatch that has drifted from the real one in core — it passes `context: null as any` (#XXX — see companion issue), has no resolve-chain or cycle protection, and carries an unreachable pre-resolution fallback (#957).
- It is the direct cause of the type-instantiation ceiling in #956: the `$extends` call whose inferred type exhausts TypeScript's budget is a call that never runs.

## The decision this needs

Two coherent directions, and they lead to very different work:

1. **The extension is redundant — delete it.** Core already applies `resolveOutput` correctly for `context.db.*`. Removing the generation of `prisma-extensions.ts` and the `$extends` path would resolve #956 outright (no `$extends`, no expensive inference, no annotation or assertion debate), and would close #957 and the `context: null` companion issue as moot. The behavioral question to settle first: should raw `context.prisma` reads ever apply `resolveOutput`, or is "raw means raw" the intended contract? Today raw means raw, and has for as long as the guard has been broken — so this option is also the one that changes no runtime behavior.
2. **The extension is intended — fix the guard.** Then every latent defect in that module becomes live simultaneously: the `context: null` hand-off, the unreachable-until-now fallback paths, the untyped `compute` returns, and the interaction with core's pipeline running `resolveOutput` a second time on the same field (double transformation is a real risk — `HashedPassword` wrapping a value that is already a `HashedPassword`). This option needs its own design pass before any code moves.

Recommend settling this before #956 is picked up, since option 1 would change that fix substantially.

## Notes

- Not a regression I can date — the guard appears to have always had this shape. Worth checking whether any app has ever depended on the extension path working.
- `examples/blog` was restored after testing; no changes committed.

borisno2:
## Summary

- The generated context factory decided whether to apply Prisma result extensions with `Object.keys(prismaExtensions).length === 0`. `Prisma.defineExtension(...)` returns a **function**, and `Object.keys()` on a function is always `[]` — so the guard was always `true` and `$extends` never actually ran in any generated app.
- Field-level `resolveOutput` transformations (password hashing, virtual fields, etc.) already reach `context.db.*` correctly through core's read pipeline, which is completely unaffected by this change. Raw `context.prisma` never applied `resolveOutput` and still doesn't — that was already the contract, just an accidental one.
- Removes the dead `packages/cli/src/generator/prisma-extensions.ts` generator module (and its test), the `$extends` call and `createExtendedPrisma` function from the generated `context.ts`, and every reference across `output-paths.ts`, `node-build.ts`, `generator/index.ts`, `commands/generate.ts`, and docs.
- The untyped `$extends` call's inferred return type (`PrismaClient | ReturnType<typeof basePrisma.$extends>`) was what exhausted TypeScript's instantiation budget on schemas with more lists, causing `TS2589: Type instantiation is excessively deep and possibly infinite`. Deleting the branch resolves that too — the Prisma client singleton is now typed directly as `PrismaClient`, no inference, no assertion.
- `opensaas generate` now removes a stale `prisma-extensions.ts` left on disk by a project generated before this fix, instead of leaving it orphaned.
- Removed two stray debug scripts in `examples/blog/` (`test-prisma-direct.ts`, `test-virtual-direct.ts`) that imported the now-deleted module directly.

No runtime behavior changes.

## Test plan

- [x] Added `packages/cli/src/generator/context-large-schema.test.ts`: type-checks the generated `context.ts` + `types.ts` for a 26-list schema with relationships under the repo's `typescript` dependency (aliased to the TS 6 compatibility shim, which reproduces the original ceiling — TS 7's native binary does not), asserting no `excessively deep` / `Excessive stack depth` diagnostics.
- [x] Updated existing generator tests/snapshots (`context.test.ts`, `output-paths.test.ts`, `node-build.test.ts`, `commands/generate.test.ts`) to drop all references to the removed module.
- [x] `examples/blog`: ran `opensaas generate` — no `prisma-extensions.ts` emitted, no "Prisma extensions generated" log line; confirmed a manually-planted stale `prisma-extensions.ts` is removed on the next `generate`; `npx tsc --noEmit` passes with zero errors.
- [x] Full `packages/cli` test suite: 376/376 passing.
- [x] `pnpm build` (core, cli) and `pnpm lint` pass on touched files.
- [x] Changesets added for `@opensaas/stack-cli` (patch) and `@opensaas/stack-core` (patch, doc-comment fix).

Closes #958
Closes #956

---
_Generated by [Claude Code](https://claude.ai/code/session_01BEueUfjVXkDL2B2tEQSPgx)_


borisno2:
## Self-review (code-review skill, high effort)

Reviewed the diff against the maintainer brief in #958's comment. No findings.

**Verified:**
- `Prisma.defineExtension(...)` returns a function, so `Object.keys(prismaExtensions).length === 0` was genuinely always `true` — the `$extends` branch never ran in any generated app. Confirmed against `@prisma/client`'s actual runtime.
- Every reference to `prisma-extensions.ts` / `prismaExtensions` / `createExtendedPrisma` / `writePrismaExtensions` / `generatePrismaExtensions` is removed consistently across `context.ts`, `output-paths.ts`, `node-build.ts`, `generate.ts`, `generator/index.ts`, snapshots, tests, and docs.
- `context.db.*` behavior is unaffected — core's read pipeline (`field-visibility.ts`) runs `resolveOutput` independently of the deleted module.
- Stale-file cleanup in `generate.ts` correctly removes a leftover `prisma-extensions.ts` from pre-fix projects on regenerate (verified against `examples/blog`).
- `examples/blog` regenerates cleanly, `tsc --noEmit` is clean, and the admin UI's password/virtual field types are unchanged.
- New regression test (`context-large-schema.test.ts`) type-checks the generated `context.ts` for a 26-list schema under the TS6 shim with no `excessively deep` diagnostics.
- Full `packages/cli` suite: 376/376 passing. `pnpm build` (core, cli), lint, and format all clean.
- Changesets present for `@opensaas/stack-cli` and `@opensaas/stack-core`.

No changes requested.


---

PR #983, opened 2026-08-22
core: redact HashedPassword.toJSON() so JSON.stringify never leaks the hash

### LINKED ISSUE #965: core: HashedPassword.toJSON() returns the raw hash, so JSON.stringify of a row leaks it — inconsistent with the field's own valueForClientSerialization redaction
> Reported from a downstream app review. No app-specific data included.

## Summary

The `password()` field deliberately redacts on the admin-UI serialization path — its builder sets:

```ts
ui: {
  valueForClientSerialization: ({ value }) => ({ isSet: !!value }),
}
```

But the value returned on read is a `HashedPassword` instance whose `toJSON()` returns the **raw bcrypt hash**:

```ts
// packages/core/src/utils/password.ts
// JSON.stringify calls this when present, so the hash serializes as a plain string.
toJSON(): string {
  return this.hash
}
```

So any serialization path that does **not** go through `valueForClientSerialization` — e.g. a consumer calling `JSON.stringify(...)` (or a Next.js server→client prop boundary, `Response.json(row)`, structured logging) on a `context.db.<list>.find*` row that includes the password field — serializes the raw hash instead of the `{ isSet }` redaction.

## Why it matters

The two redaction paths disagree, and the *unsafe* one is the language default (`JSON.stringify` calls `toJSON` automatically). A consumer who reasonably assumes "the password field is redacted" (because the admin UI shows `{ isSet }`) can leak the stored hash the moment they serialize a row through any path other than the UI helper. Field-level `read` access is the only thing standing between a `context.db` read and this — and field access defaults to allow when a list doesn't set a `read` gate on the password field.

For contrast, Keystone's `password` GraphQL type structurally exposed only `{ isSet }`, so no serialization path could surface the hash.

## Suggested direction

Make the safe behaviour the default rather than opt-in:

- Have `HashedPassword.toJSON()` return a redacted marker (e.g. `{ isSet: true }`) or throw, so `JSON.stringify` of a row never emits the hash. (`toString()` / `valueOf()` / `[Symbol.toPrimitive]` can keep returning the hash for the internal string/compare uses — `toJSON` is specifically the *serialize-for-output* hook, which is where redaction belongs.)
- And/or default the `password()` field's `read` access to deny so the field value never leaves the server unless a config explicitly opts in.

Either makes "don't serialize the hash" the default instead of relying on every consumer routing through `valueForClientSerialization`.

## Repro sketch

```ts
const user = await context.db.user.findUnique({ where: { id } })
JSON.stringify(user) // includes password: "$2b$..." (the raw hash), not { isSet: true }
```

(Not a data-loss issue; a credential-exposure footgun in the default serialization path.)

borisno2:
## Summary

- `HashedPassword.toJSON()` now returns `{ isSet: boolean }` instead of the raw bcrypt hash, so `JSON.stringify` of any row carrying a `password()` field (server→client props, `Response.json()`, the MCP `query` tool's response, structured logging) no longer leaks the stored hash.
- This matches the redaction the `password()` field's admin-UI path already applies via `valueForClientSerialization`, closing the gap where that was the *only* safe serialization path.
- String-coercion surfaces (`toString()`, `valueOf()`, `[Symbol.toPrimitive]`, template-literal interpolation, `==` comparison) are unchanged and still return the raw hash, for internal string/compare use. `compare()` is unaffected.
- Corrected the `password()` field builder's docblock, which incorrectly claimed password fields are excluded from reads by default in access control — there is no such special-casing, and field-level read access defaults to allow. The docblock now describes the actual redaction-on-serialization behavior.
- Rewrote the pre-existing test that explicitly pinned the leaky behavior (`should serialize to JSON correctly`) to assert the redacted shape instead.

Out of scope (per issue discussion): changing the `password()` field's default `read` access, and the derived auth lists' plain-text credential fields (`Account.password`/`accessToken`/etc.) — tracked separately.

## Test plan

- [x] `JSON.stringify(new HashedPassword(hash))` yields `{ isSet: true }` and contains no substring of the hash
- [x] `JSON.stringify(row)` for a row read through `context.db` with a `password()` field emits the redacted marker, no bcrypt hash anywhere in the output
- [x] The MCP `query` tool's serialized response for a list carrying a `password()` field contains no bcrypt hash, asserted against the handler's actual response body
- [x] `toString()`, `valueOf()`, `String(wrapped)`, `` `${wrapped}` ``, and `wrapped == hash` all still yield the raw hash
- [x] `compare()` unaffected
- [x] `pnpm build`, `pnpm lint`, `pnpm format`, `pnpm manypkg fix`, and the core package's full test suite (1057 tests) pass

Closes #965

---
_Generated by [Claude Code](https://claude.ai/code/session_012jDvGRH1p9YEscgU9h33nS)_

### DIFF
--- a/.changeset/silent-otters-redact.md
+++ b/.changeset/silent-otters-redact.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix `HashedPassword.toJSON()` returning the raw bcrypt hash, so `JSON.stringify` of a row (e.g. a server→client prop, `Response.json()`, an MCP tool response) no longer leaks the stored hash for a `password()` field.
+
+`toJSON()` now returns `{ isSet: boolean }`, matching the redaction the admin UI already applies via `valueForClientSerialization`. `toString()`, `valueOf()`, `[Symbol.toPrimitive]`, and `==` comparison against the hash are unchanged. If you parse `JSON.stringify`'d rows and read the password field as a string, update that code to read `.isSet` instead — this is a visible output/type change on `HashedPassword.toJSON()`, though the field's read access remains the application's to configure (unchanged).
--- a/packages/core/src/fields/index.ts
+++ b/packages/core/src/fields/index.ts
@@ -930,7 +930,10 @@ function formatCalendarDay(value: unknown): string | null | undefined {
  * ```
  *
  * **Important Notes:**
- * - Password fields are excluded from read operations by default in access control
+ * - Password field values are redacted to `{ isSet: boolean }` on serialization
+ *   (`JSON.stringify`, the admin UI). Field-level `read` access is not denied by
+ *   default — configure `access.field.read` if the raw value should never reach
+ *   `context.db` callers at all.
  * - Always use the `compare()` method to verify passwords - never compare strings directly
  * - The password field value has type `HashedPassword` which extends string with compare()
  * - Empty strings and undefined values are skipped (not hashed) to allow partial updates
--- a/packages/core/src/utils/password.ts
+++ b/packages/core/src/utils/password.ts
@@ -67,9 +67,11 @@ export class HashedPassword {
     return this.hash
   }
 
-  // JSON.stringify calls this when present, so the hash serializes as a plain string.
-  toJSON(): string {
-    return this.hash
+  // JSON.stringify calls this when present, so this is the serialize-for-output
+  // redaction point — never return the hash here. toString()/valueOf()/
+  // Symbol.toPrimitive stay unredacted for internal string/compare use.
+  toJSON(): { isSet: boolean } {
+    return { isSet: !!this.hash }
   }
 
   valueOf(): string {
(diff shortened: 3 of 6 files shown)


borisno2:
## Self-review

Ran `/review` (medium effort, single-pass) against this diff, tracing `HashedPassword.toJSON()` through every caller (field builders, MCP handler, admin UI serialization, write pipeline, auth package, examples, tests). One finding surfaced:

- **Changeset bump type (`patch` vs `major`)** — `toJSON()`'s return type changes from `string` to `{ isSet: boolean }`, which is source-breaking for any consumer typed against the old return and behavior-breaking for code that reads the serialized password field as a string. Under the repo's general versioning rules that would normally warrant `major`.

This was a deliberate call, not an oversight: issue #965's triage comment explicitly directs *"Do not select a major bump unless a maintainer explicitly asks for one"* for this exact change, while still requiring the changeset body to call out the observable-output change (which it does). Kept as `patch` per that instruction — happy to bump to `major` if a maintainer prefers.

No other findings — the fix is well-isolated; no other code path assumed `toJSON()` returned a raw string.

---

PR #984, opened 2026-08-22
Extend isIndexed to integer, timestamp, and select fields

### LINKED ISSUE #860: isIndexed is available on an arbitrary subset of field types
Found while fixing #859 (`isIndexed: true` emitting an invalid inline `@index`). Not a bug — filing so it doesn't evaporate.

## The gap

`isIndexed` is declared on an arbitrary subset of field types:

| Accepts `isIndexed` | Does not |
| --- | --- |
| `text`, `decimal`, `calendarDay`, `relationship` | `integer`, `timestamp`, `select`, `checkbox`, `json`, `password` |

The split looks like historical accretion rather than a design decision. `decimal` and `calendarDay` have it while `integer` and `timestamp` don't, which is hard to justify on any axis — indexing a timestamp you filter or sort by is considerably more common than indexing a decimal.

Today `timestamp({ isIndexed: true })` is a compile error. That's an honest failure and nowhere near as bad as what #859 fixed, so this isn't urgent — but it is a papercut for anyone coming from Keystone, where the option is uniform across field types.

## Why this wasn't folded into #859

That PR is a bug fix (patch). Adding `isIndexed` to more field types is additive (minor), and it carries design questions of its own that deserve deciding rather than assuming.

## Open questions

1. **Which types?** `integer` and `timestamp` seem clearly worth having. `select` is less obvious — it maps to a Prisma enum on Postgres and a `String` on SQLite, so it's worth confirming the index is well-defined on both before offering it.
2. **`json`?** Indexability is provider-specific (Postgres wants GIN and an expression, which `@@index([field])` doesn't express). Probably out.
3. **`password`?** Indexing a bcrypt hash has no sensible use case. Probably out.
4. **Should `timestamp` default to indexed?** `relationship` defaults FKs to `isIndexed: true` for Keystone parity. Timestamps are a common sort key, but silently adding indexes to every existing table would be a surprising and non-trivial migration for existing users, so I'd lean no.
5. **`virtual` must be excluded** regardless — it has no column, so `isIndexed` there could only ever no-op. This is the main argument against simply hoisting `isIndexed` onto `BaseFieldConfig` to make it uniform: it would document an option that lies for at least one field type.

## Implementation note

The mechanism already exists as of #859. `getPrismaType()` returns an optional `index?: boolean | 'unique'` and the generator emits `@@index([field])` / `@@unique([field])` from it. Extending a field type is therefore: add `isIndexed?: boolean | 'unique'` to its config type, and return `index` from its `getPrismaType()`. No generator changes needed.

borisno2:
## Summary

- `integer`, `timestamp`, and `select` field types now accept `isIndexed?: boolean | 'unique'`, matching `text`, `decimal`, `bigInt`, `calendarDay`, and `relationship`.
- `isIndexed: true` requests a block-level `@@index([field])`; `isIndexed: 'unique'` requests an inline `@unique` on the column; `false`/omitted generates neither.
- `select` supports both under its default plain-string column and under an opt-in native-enum column (`db: { type: 'enum' }`).
- No default indexing behavior changes for any field type — `timestamp` in particular does **not** become indexed by default, so an existing config generates a byte-for-byte identical schema unless it opts in.
- No generator changes were needed: the out-of-line `@@index`/`@@unique` mechanism from #859 already reads `getPrismaType().index` generically for every field type, and the context layer's unique-field detection (`context/index.ts`) already tests `'isIndexed' in fieldConfig` generically.
- `json`, `password`, `checkbox`, and `virtual` remain excluded (per the issue's decisions: provider-specific indexability, no sensible use case, and no column at all, respectively).

## Test plan

- [x] `packages/core/tests/field-types.test.ts` — added `isIndexed: true` / `'unique'` / omitted cases for `integer`, `timestamp`, and `select` (both plain-string and native-enum `select` columns)
- [x] `packages/cli/src/generator/prisma.test.ts` — extended the existing "should generate @@index for every scalar type that accepts isIndexed" test (rather than adding a parallel test) to cover `integer`, `timestamp`, and `select`; added a dedicated test for the native-enum `select` column
- [x] `pnpm test` passes in `packages/core` (1064 tests) and `packages/cli` (377 tests)
- [x] `pnpm build` passes in both packages
- [x] `pnpm format` — no changes needed
- [x] Minor changeset added (`@opensaas/stack-core`)
- [x] Field reference docs (`docs/content/reference/fields-api.md`, `docs/content/concepts/field-types.md`) updated to list the new types

Closes #860

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gkgv5rBZSxTuCjiWdsqXof)_

### DIFF
--- a/.changeset/silly-plums-arrive.md
+++ b/.changeset/silly-plums-arrive.md
@@ -1,0 +1,18 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Extend `isIndexed` to `integer`, `timestamp`, and `select`, matching `text`, `decimal`, `bigInt`, `calendarDay`, and `relationship`.
+
+```typescript
+fields: {
+  rank: integer({ isIndexed: true }),
+  publishedAt: timestamp({ isIndexed: true }),
+  status: select({
+    options: [{ label: 'Draft', value: 'draft' }],
+    isIndexed: 'unique',
+  }),
+}
+```
+
+`isIndexed: true` generates a block-level `@@index([field])`; `isIndexed: 'unique'` generates an inline `@unique`. `select` supports both under the default string column and a native-enum column (`db: { type: 'enum' }`). No field type's default indexing behavior changes — an existing config generates the same schema as before.
--- a/docs/content/concepts/field-types.md
+++ b/docs/content/concepts/field-types.md
@@ -58,6 +58,9 @@ fields: {
     },
     defaultValue: 0,
   }),
+  rank: integer({
+    isIndexed: true,
+  }),
 }
 ```
 
@@ -67,6 +70,7 @@ fields: {
 - `validation.min`: Minimum value
 - `validation.max`: Maximum value
 - `defaultValue`: Default integer value
+- `isIndexed`: Boolean or `'unique'` for indexing
 
 ### BigInt Field
 
@@ -221,6 +225,9 @@ fields: {
   updatedAt: timestamp({
     db: { updatedAt: true }, // Auto-update on changes
   }),
+  scheduledAt: timestamp({
+    isIndexed: true,
+  }),
 }
 ```
 
@@ -228,6 +235,7 @@ fields: {
 
 - `defaultValue.kind`: `'now'` for current timestamp
 - `db.updatedAt`: Boolean - auto-update on record changes
+- `isIndexed`: Boolean or `'unique'` for indexing (not indexed by default)
 
 ### Calendar Day Field
 
@@ -377,6 +385,10 @@ fields: {
       displayMode: 'select', // or 'radio', 'segmented-control'
     },
   }),
+  category: select({
+    options: [{ label: 'News', value: 'news' }],
+    isIndexed: true,
+  }),
 }
 ```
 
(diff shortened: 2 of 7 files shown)


borisno2:
## Self-review

Ran the diff through `/review`: manual trace of `packages/cli/src/generator/prisma.ts`'s `getFieldIndex`/`collectFieldLevelIndexColumns` and `packages/core/src/context/index.ts`'s `getUniqueWhereKeys`, plus the affected test suites (`packages/core`: 134/134 relevant tests, `packages/cli`: 138/138 relevant tests — full suites 1064/1064 and 377/377). No correctness bugs or cross-file breakage found; every generic downstream consumer of `isIndexed` reads it structurally rather than switching on field type, so `integer`, `timestamp`, and `select` compose with the existing `@@index`/`@@unique`/unique-`where`-key machinery with zero code changes needed there.

One finding surfaced, not fixed:

> The `isIndexed → { @unique modifier, index flag }` logic (the `if (isIndexed === 'unique') modifiers += ' @unique'` block paired with `index: isIndexed === true ? true : undefined`) is now duplicated across 8 call sites in `fields/index.ts` (text, integer, decimal, bigInt, timestamp, calendarDay, and both branches of select). This PR's 3 new field types push the count from 4 to 8.

I left this as-is rather than extracting a shared helper: the issue brief explicitly asked to "follow the shape the three existing implementations already share," and consolidating would mean touching the 4 pre-existing field types (text/decimal/bigInt/calendarDay) that are out of this issue's scope, for a refactor rather than a fix. Flagging it here as a reasonable follow-up if the pattern grows further (e.g. a future field type or a change to index semantics).

---

PR #989, opened 2026-08-22
fix(auth): Session.user/Account.user FK column now maps to userId

### LINKED ISSUE #935: auth: derived Session/Account FK column generates as `user`, not `userId` — every adoption diffs dirty

Found while verifying #921. Unrelated to that issue's subject, but the same ADR-0007 clean-diff promise.

## Current behavior

The generator maps a relationship's foreign-key column to the **relationship field name**, not to the Prisma FK field name. This is deliberate Keystone-parity behaviour, pinned by a test:

```
packages/cli/src/generator/prisma.test.ts:693
  it('should default foreign key @map to field name for bidirectional relationships', ...)
  // expects: profileId String? @unique @map("profile")
```

Applied to the derived auth lists, `Session.user` / `Account.user` therefore generate:

```prisma
model Session {
  userId String @map("user")
  user   User   @relation(onDelete: Cascade, fields: [userId], references: [id])
}
```

The Prisma field is `userId`, but the **physical column is `user`**.

better-auth's column is `userId`:

```js
// @better-auth/core@1.6.25 — dist/db/get-tables.mjs:114
userId: {
  type: "string",
  fieldName: options.session?.fields?.userId || "userId",
  references: { model: "user", field: "id", onDelete: "cascade" },
  index: true
}
```

Same for `account.userId` (line 199).

## Why it matters

ADR-0007's promise is that a project with a pre-existing better-auth install can model its tables and reach parity **without a destructive migration**. A column-name mismatch is the most destructive diff there is — `prisma migrate diff` reads it as a drop-and-add on the foreign key of the two busiest auth tables.

`userRelationshipDb` (`packages/auth/src/config/derive-auth-lists.ts:121`) exists precisely to mirror better-auth's FK shape, and its doc comment enumerates three dimensions it aligns — no separate FK index, `onDelete: Cascade`, non-nullable. The column name is a fourth dimension it does not cover: it only sets `foreignKey: { map: column }` when `fields.userId` is explicitly configured, so with no override the generator default (`user`) stands.

Reaching parity today requires the developer to hand-write the better-auth **default** as if it were a rename:

```ts
authPlugin({
  session: { fields: { userId: 'userId' } },
  account: { fields: { userId: 'userId' } },
})
```

`adoptBetterAuthTables()` — the recipe that exists specifically to remove this kind of hand-repetition — does not set it, so the recommended adoption path produces the wrong column.

## Reproduction

Generate a schema from any `authPlugin` config, with or without adoption knobs, and read the `Session` / `Account` blocks:

```ts
const schema = await generatePrismaSchema(
  await config({
    db: { provider: 'sqlite' },
    plugins: [authPlugin({ emailAndPassword: { enabled: true } })],
    lists: {},
  }),
)
// Session: userId String @map("user")
// Account: userId String @map("user")
```

## Note on existing test coverage

`packages/auth/tests/generated-fk-shape.test.ts` asserts the FK shape for #863 with `expect(block).toMatch(/userId\s+String\s/)`. That regex matches `userId String @map("user")`, so the current coverage does not catch this — a fix should tighten it to assert the absence of the `@map`, or the presence of `@map("userId")`, whichever form is chosen.

## Suggested fix

Two candidate shapes, worth a decision during triage:

1. Have `userRelationshipDb` always set the FK column explicitly (`foreignKey: { map: fields.userId ?? 'userId' }`), so the auth lists pin their own column and never inherit the Keystone-parity default.
2. Have `adoptBetterAuthTables()` seed `fields: { userId: 'userId' }` — narrower, but leaves a bare `authPlugin()` (no adoption recipe) still generating `user`, which is the wrong column for anyone who later points it at a real better-auth database.

Option 1 looks right: the derived lists mirror better-auth, and the Keystone FK-naming convention is not something better-auth shares. Both change the generated schema for existing greenfield projects, so this needs a migration note either way.

## Related

- ADR-0007 — the clean-diff promise this breaks
- #863 / #679 — the earlier FK-shape alignments, which covered index/cascade/nullability but not the column name
- #921 — the constraint-name gap, found in the same pass

borisno2:
## Summary

- `userRelationshipDb()` (`packages/auth/src/config/derive-auth-lists.ts`) only set the FK column `@map` when a `userId` field override was explicitly configured in `authPlugin`. With no override, the generator's own Keystone-parity default stood: the FK column mapped to the **relationship field name** (`user`) rather than the Prisma field name (`userId`).
- better-auth's own column is `userId` on both `Session` and `Account`, so a bare `authPlugin()` config (and `adoptBetterAuthTables()`, which doesn't set the override either) generated a schema that diffs as a drop-and-add on the FK of the two busiest auth tables — breaking the ADR-0007 clean-diff promise.
- The helper now always sets `foreignKey: { map: ... }`, defaulting to `'userId'` when `fields.userId` isn't configured. An explicit override still wins, unchanged.
- Tightened `generated-fk-shape.test.ts` and `derive-auth-lists.test.ts`, both of which previously asserted patterns that matched the wrong (`user`) column and would not have caught this regression.
- Updated `packages/auth/CLAUDE.md`'s derivation summary to describe the new default.

## Migration note

This changes generated schema output for existing greenfield projects that ran with the `user` column. They'll need to rename the column on `Session` and `Account` to match (e.g. `ALTER TABLE "Session" RENAME COLUMN "user" TO "userId";` and the same for `Account`), called out in the changeset.

## Test plan

- [x] `pnpm --filter @opensaas/stack-auth test` — 261 passed, 2 skipped
- [x] `pnpm --filter @opensaas/stack-core test` — 1067 passed
- [x] `pnpm --filter @opensaas/stack-cli test` — 377 passed
- [x] `pnpm lint` — clean (pre-existing unrelated warnings only)
- [x] Changeset added (`@opensaas/stack-auth`: patch)

Closes #935

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gnnd1nT6Y9Yq2HXeizBYZ8)_

### DIFF
--- a/.changeset/tidy-otters-clap.md
+++ b/.changeset/tidy-otters-clap.md
@@ -1,0 +1,7 @@
+---
+'@opensaas/stack-auth': patch
+---
+
+Fix `Session.user`/`Account.user` foreign key generating a `user` physical column instead of `userId`, mismatching better-auth's own schema and breaking clean-diff adoption (ADR-0007). An explicit `fields: { userId: ... }` override is unaffected.
+
+**Migration note:** existing greenfield projects need to rename the column on `Session` and `Account` (e.g. `ALTER TABLE "Session" RENAME COLUMN "user" TO "userId";` and the same for `Account`) to match the new generated schema.
--- a/packages/auth/CLAUDE.md
+++ b/packages/auth/CLAUDE.md
@@ -77,7 +77,10 @@ and the plugin's add-vs-extend logic consume:
   (defaults to `modelName` when it differs from the better-auth default,
   otherwise unset — i.e. unchanged output when `tableName` isn't set)
 - per-model `fields` (better-auth field → column) → field-level `@map`
-- the `userId` column override → the `user` relationship foreign-key `@map`
+- the `user` relationship foreign key always maps to a `userId` column
+  (mirroring better-auth's own column, not the generator's Keystone-parity
+  default of the relationship field name) — a `userId` column override
+  in `fields` replaces that default, exactly like any other field `@map`
 - relationship refs between the Auth lists follow the derived keys
   (e.g. `Session.user → AuthUser.sessions`)
 
--- a/packages/auth/src/config/derive-auth-lists.ts
+++ b/packages/auth/src/config/derive-auth-lists.ts
@@ -78,18 +78,23 @@ function fieldDb(fieldName: string, fields: Record<string, string>): { map: stri
  * `Account.user`), honouring a `userId` column override from the better-auth
  * `fields` map and mirroring better-auth's own FK shape: no separate FK index
  * — the index is applied at the field level via `isIndexed: false` —
- * `onDelete: Cascade`, and a required (non-nullable) foreign key, since
- * better-auth's adapter always writes a `userId` on every session/account row
- * it creates. This means a generated Auth schema diffs clean against a live
- * better-auth database on all three dimensions instead of showing a spurious
- * index drop, a referential-action change, and a `DROP NOT NULL` (issues #679,
- * #863).
+ * `onDelete: Cascade`, a required (non-nullable) foreign key, and a `userId`
+ * column, since better-auth's adapter always writes a `userId` column on
+ * every session/account row it creates. The column map is set unconditionally
+ * (defaulting to `userId`) rather than only when `fields.userId` is
+ * configured, since the generator's own Keystone-parity default would
+ * otherwise map the column to the relationship field name (`user`) — the
+ * wrong column for a database better-auth itself wrote to. This means a
+ * generated Auth schema diffs clean against a live better-auth database on
+ * all four dimensions instead of showing a spurious index drop, a
+ * referential-action change, a `DROP NOT NULL`, and a column rename (issues
+ * #679, #863, #935).
  */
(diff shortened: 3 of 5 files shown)


borisno2:
## Self-review

Reviewed the diff (`packages/auth/src/config/derive-auth-lists.ts`, tests, `CLAUDE.md`, changeset) at medium effort. No correctness, reuse, simplification, or efficiency issues found.

**Findings:** none.

**Notes:**
- The fix is a single-line logic change (`fields.userId ?? 'userId'`, `foreignKey` now set unconditionally) confirmed against the generator's FK `@map` logic (`packages/core/src/fields/index.ts`) and the existing "default FK `@map` to field name" test it deliberately does not touch (non-auth relationships are unaffected).
- `generated-fk-shape.test.ts` and `derive-auth-lists.test.ts` were tightened — both previously asserted patterns that matched the buggy `user` column and would not have caught this regression.
- `adoptBetterAuthTables()` needed no change — it inherits the corrected default from the helper, as called out in the issue.
- Verified locally: `@opensaas/stack-auth` (261 passed), `@opensaas/stack-core` (1067 passed), `@opensaas/stack-cli` (377 passed), `pnpm lint` clean (pre-existing unrelated warnings only).


---

PR #990, opened 2026-08-22
fix(auth): emit the three indexes better-auth declares (Session.user, Account.user, Verification.identifier)

### LINKED ISSUE #935: auth: derived Session/Account FK column generates as `user`, not `userId` — every adoption diffs dirty

### LINKED ISSUE #937: auth: three indexes better-auth creates are never emitted — and ADR-0007's stated rationale for omitting the FK index is wrong

Found while verifying #921. Unrelated to that issue's subject, but the same ADR-0007 clean-diff promise.

## Current behavior

`@better-auth/core@1.6.25` declares `index: true` on exactly three columns (`dist/db/get-tables.mjs`):

| Model | Column | Line |
| --- | --- | --- |
| `session` | `userId` | 123 |
| `account` | `userId` | 200 |
| `verification` | `identifier` | 48 |

better-auth's own migrator turns each into a named index (`better-auth/dist/db/get-migration.mjs`) — `<table>_<field>_idx` on the create-table path (line 321) and on the alter-table path (line 282). So a live better-auth database carries `session_userId_idx`, `account_userId_idx` and `verification_identifier_idx`.

The derived auth lists emit **none of the three**:

- `Session.user` and `Account.user` explicitly opt out with `isIndexed: false` (`packages/auth/src/config/derive-auth-lists.ts:201,229`).
- `Verification.identifier` (`:259`) carries no `isIndexed` at all, so nothing is emitted.

Confirmed against generated output — no `@@index` line appears on any of the three models.

## Why it matters

Two separate problems.

**1. The generated schema drops indexes that exist in the live database.** `prisma migrate diff` against a real better-auth install reads three absent indexes as three `DROP INDEX` statements. That is the destructive-migration failure mode ADR-0007 exists to prevent, in the same shape as #679 (a spurious index *add*) but in the opposite direction.

**2. ADR-0007's recorded rationale is factually wrong against the pinned better-auth version.** The ADR says:

> The `Session.user` / `Account.user` foreign keys deliberately **omit** the generator's default `@@index([userId])` (via `isIndexed: false`) […] This is not a generator regression — it is what makes an adopted auth schema diff **clean** against a live better-auth database, which emits no separate FK index.
>
> The tradeoff is that these FK columns are unindexed by default (matching better-auth, which relies on the `token`/`identifier` lookup indexes rather than a `userId` index)

Against better-auth@1.6.25 both claims are inverted: better-auth *does* declare a `userId` index on session and account, and it does **not** index `token` via this channel (`token` is `unique: true` with no `index`, so it gets a unique constraint, not a separate index). The parenthetical about relying on `identifier` is also awkward — `verification.identifier` is one of the three indexes the stack omits.

So the ADR both misstates the tradeoff and justifies the current shape with a premise that no longer holds. The performance cost is real too: `Session` and `Account` are looked up by `userId` on ordinary session flows, and both are currently unindexed by deliberate choice.

## Open question for triage

**Did `index: true` postdate #679/ADR-0007?** I did not check better-auth's history. If those flags were added after the ADR was written, then #679 was correct at the time and this is version drift rather than an error of reasoning — which changes how the ADR should be amended (a "this changed in better-auth X.Y" note versus a correction). Worth settling before rewriting the ADR text.

**Not verified:** what `@better-auth/cli generate --prisma` emits. The CLI package isn't installed in this workspace, so I could only confirm the Kysely migrator's behaviour. If the Prisma generator ignores `index: true`, a Prisma-origin live database would *not* have these indexes, and the fix would need to be configurable rather than unconditional.

## Suggested fix

Emit all three indexes so the derived lists mirror better-auth:

- `Session.user` / `Account.user` — drop the `isIndexed: false`, or set it to `true`, so the FK index is generated. Prisma's derived name for `@@index([userId])` on a model mapped to `session` is `session_userId_idx`, which matches better-auth's migrator name exactly — no naming knob needed (contrast #921).
- `Verification.identifier` — add `isIndexed: true`; likewise derives `verification_identifier_idx`.

Note this interacts with #935: the FK column currently generates as `user`, so the index would be built on the wrong column until that is fixed. These two should probably land together, or #935 first.

Then amend ADR-0007's fifth decision bullet to describe what better-auth actually does, and extend `packages/auth/tests/generated-fk-shape.test.ts` — which currently asserts the *absence* of `@@index([userId])` (lines 56-57) and would need inverting — to pin all three indexes.

## Related

- ADR-0007 — the decision this contradicts
- #679 / #698 — where the `isIndexed: false` was introduced
- #935 — the FK column name, which this index sits on
- #921 — the constraint-name gap, found in the same pass

borisno2:
## Summary

Fixes #937: better-auth declares `index: true` on `session.userId`, `account.userId`, and `verification.identifier` (`@better-auth/core@1.6.25`), and both its Kysely migrator and Prisma generator turn each into a named index (`session_userId_idx`, `account_userId_idx`, `verification_identifier_idx`). The derived auth lists emitted none of the three (`Session.user`/`Account.user` opted out via `isIndexed: false`; `Verification.identifier` had no `isIndexed` at all). All three now carry `isIndexed: true`.

A generated Auth schema no longer diffs dirty against a live better-auth database (`prisma migrate diff` previously read this as three dropped indexes), and `Session`/`Account` are no longer unindexed on their hottest lookup path (`userId`).

Also corrects ADR-0007's FK-index decision bullet, which asserted the opposite of better-auth's actual behavior (that it "emits no separate FK index" and relies on `token`/`identifier` lookups instead) — a premise that was already false against the better-auth version pinned when the bullet was written. See the issue's triage comments for the version timeline.

**Note on #935:** this PR originally also fixed #935 (the `Session.user`/`Account.user` FK column defaulting to the wrong physical name), since #937's index-naming fix depends on it — an index built on the wrong column doesn't match better-auth's naming convention. #935 has since landed separately on `main` via #989, so this branch has been merged up to pick that up; the remaining diff here is #937-only (the three `isIndexed` additions and the ADR-0007 correction).

**Migration note:** existing projects (greenfield or adopted) will see a migration on their next `prisma migrate dev`/`db push` that adds the three indexes — called out in the changeset.

## Test plan

- [x] `derive-auth-lists.test.ts` — inverted the FK-index assertions, added coverage for `Verification.identifier`
- [x] `generated-fk-shape.test.ts` — inverted the `@@index` assertions on generated schema text; added coverage for `Verification`, and for `adoptBetterAuthTables()` both with and without `useBetterAuthTableNames` (verifying `@@index([userId])`/`@@index([identifier])` are present, and that the `useBetterAuthTableNames` case's index derives the exact `session_userId_idx`-style name via a non-destructive `prisma migrate diff --script` check performed manually during development)
- [x] Full `packages/auth`, `packages/cli`, `packages/core` test suites pass (265/265, 377/377, 1067/1067) after merging up to pick up #989
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` clean
- [x] Changeset added (`@opensaas/stack-auth`, patch) — trimmed to describe only the index addition, since #989 already shipped its own changeset for the FK column fix

Closes #937


borisno2:
## Automated code review (medium effort)

Reviewed the full diff (single-pass, 8-angle: correctness, removed-behavior, cross-file tracing, reuse, simplification, efficiency, altitude, CLAUDE.md conventions).

**Verified:**
- Traced `packages/core/src/fields/index.ts:1443-1476` — confirmed the old default FK `@map` fallback was the bare field name (`"user"`), not `"userId"`, so the PR's premise for #935 is accurate.
- Confirmed a relationship's default `isIndexed` is already `true` (`field.isIndexed ?? true`) — explicitly setting `isIndexed: true` on `Session.user`/`Account.user` correctly restores that default after it had been force-disabled.
- Confirmed `@@index([...])` is emitted against the Prisma field name, not the `@map`'d column, so the column-rename and index changes compose correctly.
- Checked the one-to-one `foreignKey === true` branch in `shouldHaveForeignKey` — not reachable here (many-to-one relationship), so always passing an object for `db.foreignKey` is safe.
- Grepped for other `deriveAuthLists`/`userRelationshipDb` consumers and for stale assertions referencing the old `isIndexed: false`/no-`foreignKey` behavior — none found outside the files already updated.
- Ran `packages/auth`, `packages/cli`, `packages/core` suites: 265/265, 377/377, 1067/1067 passing.

**Not a diff bug, noted for awareness:** `examples/auth-demo/prisma/schema.prisma` and `examples/starter-auth/prisma/schema.prisma` are checked-in generated output that will drift further from a fresh `pnpm generate` after this change — pre-existing staleness, not introduced here.

**Result: no findings.**


---

PR #996, opened 2026-08-22
test(auth): assert derived auth lists match better-auth's own table definitions

### LINKED ISSUE #988: auth: add a drift test asserting the derived auth lists match better-auth's own table definitions
> *This was generated by AI during triage.* Split out of #987 so it can land without waiting on the decision about restructuring derivation.

## Agent Brief

**Category:** enhancement
**Summary:** Add a test that reads better-auth's own table definitions and asserts the derived auth lists match them, converting silent upstream drift into a failing build.

**Current behavior:**

`deriveAuthLists` hand-writes the auth lists as a manual mirror of better-auth's schema — every field, column map, type, nullability, and index transcribed by hand. Nothing checks the mirror against its source. Every divergence found so far (#921, #935, #937, #981, #986) was found by a human reading better-auth's compiled output, and each was fixed by hand-editing the mirror. The mechanism that produced them stays intact.

#987 asks whether derivation should read better-auth's definitions directly. That is a large, undecided restructure. **This issue is the cheap version of the same protection**: keep the hand-written lists, and add a test that fails when they stop matching upstream.

**Desired behavior:**

A test in the auth package reads better-auth's resolved table definitions and compares them against the derived lists, failing when they disagree on any dimension that affects the generated schema.

**This approach is verified, not speculative.** A working prototype was built during triage and correctly detected both known drifts from upstream metadata alone, with no hardcoded expectations:

```
session.userId: better-auth INDEX -> stack isIndexed=false
   FK column: better-auth="userId" stack map=undefined (=> generator default "user")
account.userId: better-auth INDEX -> stack isIndexed=false
   FK column: better-auth="userId" stack map=undefined (=> generator default "user")
verification.identifier: better-auth INDEX -> stack isIndexed=undefined
```

**No new dependency is required.** `getAuthTables` is re-exported from `better-auth/db`, and `better-auth` is already a dev dependency of the auth package — confirmed resolvable from its test suite. Do **not** add `@better-auth/core` as a direct dependency; #987's body mentions that as an obstacle, and it turned out not to be one via this path.

**What the upstream definitions provide.** For each model, per field: the column name (`fieldName`), type, requiredness, `unique`, `index`, `returned`, and `references` (with target model, field, and `onDelete`). Confirmed shape against the pinned version:

```
MODEL session (table=session)
    token->token string req uniq
    userId->userId string req INDEX ref:user.id/cascade
MODEL account (table=account)
    userId->userId string req INDEX ref:user.id/cascade
    accessToken->accessToken string notReturned
    password->password string notReturned
MODEL verification (table=verification)
    identifier->identifier string INDEX
```

**Dimensions to compare** (each maps to a real drift instance):

- **Column name** — `fieldName` vs the derived field's FK/column map. Catches #935.
- **Index presence** — `index: true` vs the derived field's `isIndexed`. Catches #937.
- **Uniqueness** — `unique: true` vs `isIndexed: 'unique'`. Related to #921.
- **Field presence** — a model field upstream with no counterpart in the derived list. Catches #986's new `account.issuer` column.
- **Requiredness** — `required` vs the derived field's nullability.
- **References** — target model/field and `onDelete`, against the derived relationship.

`returned: false` is worth surfacing in the report for #981's benefit, but is a serialization contract rather than a schema fact — do not fail the build on it.

**Handling the drifts that exist today.** The comparison fails right now, because #935 and #937 are unfixed. The test must still land green, so encode the current known divergences as an **explicit, documented exception list** — each entry naming the field, the dimension, and the issue tracking it. New drift fails immediately; known drift is visible and attributed. As #935 and #937 land, their entries are removed, and an entry that no longer corresponds to real drift should itself fail the test so the list cannot rot into a permanent silencer.

**Key interfaces:**

- `getAuthTables` from `better-auth/db` — takes a better-auth options object and returns resolved table definitions keyed by model name.
- `deriveAuthLists` — takes normalized models and returns `{ keys, lists }`; the existing derive tests show the minimal fixture shape for a default-options call.
- The mapping from better-auth model names (`user`, `session`, `account`, `verification`, `rateLimit`) to derived list keys — available from `deriveAuthLists`'s returned `keys`, so read it from there rather than hardcoding.

**Two specifics worth knowing before starting:**

- **The `rateLimit` model is absent** from the upstream tables under default options; it appears only when database-backed rate-limit storage is configured. The derived `RateLimit` list is conditional in the same way. Cover it with a second call using that option rather than letting it silently fall out of the comparison.
- **The FK is shaped differently on each side.** Upstream models it as a scalar field carrying `references`; the derived list models it as a `relationship()` whose column comes from the FK map. The comparison must bridge that — matching the upstream `userId` field against the derived `user` relationship — rather than looking for a field of the same name.

**Typing:** per the repo's conventions, type the upstream definitions properly. Do not use `any`, and do not use type casting to paper over the options-object type — if better-auth's exported types don't fit directly, declare a narrow local interface describing only the members the comparison reads.

**Acceptance criteria:**

- [ ] A test in the auth package compares derived lists against `getAuthTables` output across column name, index presence, uniqueness, field presence, requiredness, and references
- [ ] The suite passes on the current codebase, with today's known divergences captured in a documented exception list that names the tracking issue for each
- [ ] An exception-list entry that no longer corresponds to real drift causes a failure, so entries cannot outlive their fix
- [ ] Removing the `isIndexed: false` from a user relationship, or changing an FK column map, changes the test's outcome — verified by trying it, not assumed
- [ ] Introducing a field upstream that the derived lists lack is detected — simulate by removing a field from a derived list and confirming failure
- [ ] The database-backed rate-limit model is covered by its own comparison
- [ ] Failure output names the model, field, and dimension that diverged — a bare boolean assertion is not sufficient
- [ ] No new package dependency is added
- [ ] No `any` and no type casting in the new code
- [ ] `pnpm lint`, `pnpm format`, and the auth package test suite pass

**Changeset:** required, via the `pr-changeset` skill. Test-only change with no runtime effect — patch.

**Out of scope:**

- **Fixing any of the drift it detects.** #935 and #937 have their own briefs; this issue records their divergences in the exception list and changes no production code.
- **Restructuring derivation to read `getAuthTables` at runtime** (#987). This is deliberately the test-only version.
- The better-auth version bump (#986) and app-supplied indexes (#985).
- Comparing anything the stack deliberately adds beyond better-auth's schema — timestamps, access blocks, `extendUserList` merging. Only fields better-auth declares are in scope; extra fields on the stack side are not drift.

**Related:** #987 (the larger restructure this is the cheap alternative to) · #935, #937 (drift this detects today; their fixes shrink the exception list) · #921, #981, #986 (other instances of the same drift class) · ADR-0007

borisno2:
## Summary

- Adds `packages/auth/tests/auth-lists-drift.test.ts`, which reads better-auth's own resolved table definitions (`getAuthTables` from `better-auth/db`) and compares them against `deriveAuthLists()`'s output, across six dimensions: column name, index presence, uniqueness, field presence, requiredness, and FK references.
- Bridges better-auth's scalar `userId` FK field (with `references`) against the derived `user` relationship field on `Session`/`Account`, since the two model the same column differently.
- Encodes today's known divergences as a documented `KNOWN_EXCEPTIONS` list — currently **empty**, since #935 and #937 (the two divergences the issue was filed against) are already fixed on `main`. A companion test fails if an exception entry no longer corresponds to real drift, so the list can't rot into a permanent silencer.
- Adds regression tests proving the comparison actually catches drift: removing `isIndexed` from `Session.user`, changing the `userId` FK column map, and deleting a field (`Account.scope`) from the derived list.
- Covers the database-backed `RateLimit` list via a second `getAuthTables({ rateLimit: { storage: 'database' } })` call.
- No production code changes — test-only, per the issue's scope.

## Test plan

- [x] `pnpm --filter @opensaas/stack-auth test` — 272 passed, 2 skipped (including the 7 new tests)
- [x] `pnpm lint`
- [x] `pnpm format`
- [x] `pnpm manypkg fix`

Closes #988

---
_Generated by [Claude Code](https://claude.ai/code/session_01Hk65YCCUgwGEwmh7wcfjSB)_

### DIFF
--- a/packages/auth/tests/auth-lists-drift.test.ts
+++ b/packages/auth/tests/auth-lists-drift.test.ts
@@ -10,9 +10,11 @@ import type { NormalizedAuthModels } from '../src/config/types.js'
 /**
  * Drift detector for issue #988: reads better-auth's own resolved table
  * definitions (`getAuthTables`) and compares them against `deriveAuthLists`'
- * output, so an upstream schema change that the hand-written derivation
- * doesn't mirror fails this suite instead of waiting for a human to notice
- * (as #921/#935/#937/#981/#986 all were).
+ * output — column name, index presence, uniqueness, field presence,
+ * requiredness, references, scalar type, and static default values — so an
+ * upstream schema change that the hand-written derivation doesn't mirror
+ * fails this suite instead of waiting for a human to notice (as
+ * #921/#935/#937/#981/#986 all were).
  *
  * Deliberately out of scope (see the issue body): fields the stack adds that
  * better-auth doesn't declare (timestamps, access blocks, extendUserList
@@ -42,6 +44,28 @@ const TIMESTAMP_FIELDS = new Set(['createdAt', 'updatedAt'])
 
 function normalizeOnDelete(action: string): string {
   return action.replace(/\s+/g, '').toLowerCase()
+}
+
+/**
+ * The Prisma scalar type a correctly-derived field should carry for a given
+ * better-auth field type. `undefined` means "not modeled by any field in
+ * these lists" (json/array types don't appear on user/session/account/
+ * verification/rateLimit) — the caller skips the comparison in that case
+ * rather than treating it as a mismatch.
+ */
+function expectedPrismaType(upstream: DBFieldAttribute): string | undefined {
+  switch (upstream.type) {
+    case 'string':
+      return 'String'
+    case 'boolean':
+      return 'Boolean'
+    case 'date':
+      return 'DateTime'
+    case 'number':
+      return upstream.bigint ? 'BigInt' : 'Int'
+    default:
+      return undefined
+  }
 }
 
 /** Narrows a derived list's field to a `RelationshipField` — the shape the FK-bridged comparison reads. */
@@ -109,6 +133,30 @@ function compareScalarField(
       dimension: 'uniqueness',
       detail: `better-auth unique=${upstreamUnique} vs derived isIndexed='unique' is ${isUnique}`,
     })
+  }
+
+  const expectedType = expectedPrismaType(upstream)
+  if (expectedType !== undefined && prismaType.type !== expectedType) {
+    divergences.push({
+      model,
+      field: fieldKey,
+      dimension: 'scalar type',
+      detail: `better-auth type "${upstream.type}"${upstream.bigint ? ' (bigint)' : ''} expects Prisma "${expectedType}" vs derived "${prismaType.type}"`,
+    })
+  }
+
+  // Only a static, non-function default is comparable — better-auth's
+  // function-valued defaults (e.g. `createdAt: () => new Date()`) are applied
+  // at write time, not modeled as a derived field's own `defaultValue`.
+  if (typeof upstream.defaultValue !== 'function' && upstream.defaultValue !== undefined) {
+    if (derived.defaultValue !== upstream.defaultValue) {
+      divergences.push({
+        model,
+        field: fieldKey,
+        dimension: 'default value',
+        detail: `better-auth defaultValue=${JSON.stringify(upstream.defaultValue)} vs derived defaultValue=${JSON.stringify(derived.defaultValue)}`,
+      })
+    }
   }
 }
 
--- a/packages/auth/tests/auth-lists-drift.test.ts
+++ b/packages/auth/tests/auth-lists-drift.test.ts
@@ -1,0 +1,362 @@
+import { describe, it, expect } from 'vitest'
+import { getAuthTables } from 'better-auth/db'
+import type { BetterAuthOptions } from 'better-auth'
+import type { DBFieldAttribute } from 'better-auth/db'
+import type { FieldConfig } from '@opensaas/stack-core'
+import type { RelationshipField } from '@opensaas/stack-core/fields'
+import { deriveAuthLists } from '../src/config/derive-auth-lists.js'
+import type { NormalizedAuthModels } from '../src/config/types.js'
+
+/**
+ * Drift detector for issue #988: reads better-auth's own resolved table
+ * definitions (`getAuthTables`) and compares them against `deriveAuthLists`'
+ * output, so an upstream schema change that the hand-written derivation
+ * doesn't mirror fails this suite instead of waiting for a human to notice
+ * (as #921/#935/#937/#981/#986 all were).
+ *
+ * Deliberately out of scope (see the issue body): fields the stack adds that
+ * better-auth doesn't declare (timestamps, access blocks, extendUserList
+ * merging), constraint *names* (#921), and `returned: false` (a serialization
+ * contract, not a schema fact).
+ */
+
+type Divergence = {
+  model: string
+  field: string
+  dimension: string
+  detail: string
+}
+
+/**
+ * better-auth models its foreign keys as a plain scalar field carrying
+ * `references`; the derived lists model the same column as a `relationship()`
+ * field under a different key. This is the map between the two per model.
+ */
+const FK_BRIDGE: Record<string, Record<string, string>> = {
+  session: { userId: 'user' },
+  account: { userId: 'user' },
+}
+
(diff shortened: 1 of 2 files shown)


packages/auth/tests/auth-lists-drift.test.ts
```diff
@@ -0,0 +1,362 @@
...
+    divergences.push({
+      model,
+      field: fieldKey,
+      dimension: 'field presence',
+      detail: `better-auth declares '${fieldKey}' but the derived list has no matching field`,
+    })
+    return
+  }
+
+  const prismaType = derived.getPrismaType(fieldKey)
```
borisno2, line 93: `compareScalarField` never reads `prismaType.type`, so it can't detect a scalar *type* regression — only column name, requiredness, index, and uniqueness are compared. Concretely: `createRateLimitList`'s own docstring warns that `lastRequest` must be `bigInt()` rather than `integer()` "which would overflow" — but if someone reverted that to `integer()`, `getPrismaType('lastRequest')` would return `{ type: 'Int', ... }` instead of `{ type: 'BigInt', ... }`, and since `type` is never compared here, this suite would report zero divergences and pass clean, silently missing exactly the class of regression issue #988 was filed to catch. Worth adding a `type`-dimension check (mapping upstream's `type`/`bigint` to the derived Prisma type) alongside the existing four dimensions.

---

packages/auth/tests/auth-lists-drift.test.ts
```diff
@@ -0,0 +1,362 @@
...
+
+  const upstreamUnique = upstream.unique === true
+  if (isUnique !== upstreamUnique) {
+    divergences.push({
+      model,
+      field: fieldKey,
+      dimension: 'uniqueness',
+      detail: `better-auth unique=${upstreamUnique} vs derived isIndexed='unique' is ${isUnique}`,
+    })
+  }
```
borisno2, line 136: `compareScalarField` also never compares `defaultValue` (better-auth's `getAuthTables()` reports it, e.g. `user.emailVerified` upstream carries `defaultValue: false`). If a future better-auth version changed that default (or the derived `checkbox({ defaultValue: false })` in `createUserList` drifted from it), this comparator would report no divergence — the default-value dimension is silently out of scope, and it isn't listed in the file's own "Deliberately out of scope" docblock alongside timestamps/access/extendUserList/constraint-names/`returned: false`, so this reads as an oversight rather than an intentional exclusion.

---

borisno2 (commented):
Single-pass review at high effort (Agent tool unavailable in this context, so no multi-agent fan-out or subagent verify pass ran — one context worked through all 8 angles serially).

This PR is test-only (`packages/auth/tests/auth-lists-drift.test.ts` + changeset), adds no production code changes, and the new suite runs green (`vitest run tests/auth-lists-drift.test.ts` — 7/7 passing) with the actual `better-auth@1.6.25` `getAuthTables()` output verified by hand against every field the comparator reads. The three built-in regression tests (removed `isIndexed`, wrong FK column map, missing field) do correctly fail when I mutated the derived output, so the detector isn't a tautology.

The two findings below are both coverage gaps in the comparator itself — cases where a real upstream/derived divergence would still report a clean pass, undermining the drift-detector's own stated purpose (catching schema drift issue #988 was filed for). No other correctness bugs, cross-file breakage, unsafe removed behavior, or CLAUDE.md violations found.

---

packages/auth/tests/auth-lists-drift.test.ts
borisno2: Fixed in 63f9053 — added a `scalar type` dimension mapping better-auth's `type`/`bigint` to the expected Prisma type and comparing it against `getPrismaType().type`.

---

packages/auth/tests/auth-lists-drift.test.ts
borisno2: Fixed in 63f9053 — added a `default value` dimension for static (non-function) upstream defaults, and updated the file's top comment to list scalar type + default value as compared dimensions rather than leaving them ambiguously out of scope.

---

Issue #986, opened 2026-08-22
auth: upgrade better-auth to 1.7.1 — new required `account.issuer` column, new `@@unique([issuer, accountId])`, and a stale `^1.3.29` peer floor


borisno2:
> *This was generated by AI during triage.* Split out of #937, where the version bump was originally going to ride along.

## Summary

`packages/auth` currently declares `better-auth` as a `^1.3.29` peer with a `^1.6.25` dev dependency. Latest is `1.7.1`. The upgrade is not a routine bump — the 1.7 line changes the auth table schema, so it needs its own migration story rather than riding along with the index-parity fix in #937.

## What changes between the pinned 1.6.25 and 1.7.1

Verified by unpacking both `@better-auth/core` tarballs and diffing the compiled table definitions.

**1. A new required column on `account`.** `issuer` is new in the 1.7 line:

```js
issuer: { type: "string", required: true, fieldName: options.account?.fields?.issuer || "issuer" }
```

It is the only field added or removed across all models between the two versions. The derived `Account` list has no such field, so after the bump the generated schema is missing a required better-auth column — and existing projects gain a NOT NULL column that needs a backfill decision for existing rows. better-auth ships `createLocalAccountIssuer` / `createOAuthAccountIssuer` helpers (exported from `@better-auth/core/db`) which are presumably how the value is derived; worth checking whether they imply a backfill rule the migration can use.

**2. A new default composite unique index on `account`:** `@@unique([issuer, accountId])`, merged ahead of any app-supplied entries.

**3. A new table-level `indexes` API.** Each model accepts an `indexes` array (`{ name?, fields, unique? }`) merged with the defaults via a `mergeTableIndexes` helper that dedupes on `(name, fields, unique)`. This does **not** exist in 1.6.25 — zero occurrences. Relevant to #985, which is about giving apps the same capability through the stack's config; noted there.

**4. The three `index: true` flags #937 is about are unchanged** — still on `session.userId`, `account.userId`, `verification.identifier`. The bump does not invalidate #937's premise.

## The peer floor is stale regardless

[...]

Issue #921, opened 2026-08-15
auth: derived auth lists can't adopt a live constraint name, so `User.email` / `Session.token` can diff dirty


borisno2:
> *This was generated by AI during triage.*

## Current behavior

The derived auth lists emit unique constraints at the field level: the user model's `email` and the session model's `token` both carry `isIndexed: 'unique'`. That generates a constraint whose name Prisma derives.

A project adopting a pre-existing better-auth database has those same constraints already, named by whatever migration created them. If the live names differ from Prisma's derived names, the generated schema diffs on the constraint even when every column, table name and schema lines up.

`authPlugin`'s adoption surface — per-model `modelName`, `fields`, `tableName`, `schema` — has no member for a constraint name, and the application cannot supply one from outside, because **it never writes the derived list config**. `deriveAuthLists` does.

## Why this needs an auth-side change and not just the core one

#918 covers the core half: a single-field `db.indexes` entry can carry an explicit `name`, which is how any list names a single-column constraint. That capability is unreachable here. An application can only use `db.indexes` on a list it declares itself, and the whole point of the derived auth lists is that the plugin declares them.

So closing this means threading a name from the auth config through to the derived list — a new adoption knob alongside `modelName` / `fields`, and derivation emitting the named form instead of the field-level one when it is set.

**Blocked on #918**, which decides the core form the derived lists would emit.

## Scope: this is not limiter-specific

Raised out of #909's triage, where the derived limiter list reproduces better-auth's `unique: true` on `key`. But the same gap already applies to the base four:

- the user model's `email` — unique
- the session model's `token` — unique
- the verification model's `identifier`, and two further indexed columns — indexed, though these are emitted with `isIndexed: false` today under ADR-0007's clean-diff rules, so check what actually generates before assuming they need this

Whatever shape this takes should cover the base models and the limiter uniformly, not add a fifth-model special case.

## Open questions for triage

- **Where the name lives.** A sibling to `fields` keyed by field name (`indexNames: { email: '...' }`)? Or an expansion of `fields`, which currently maps a field name to a column name, into something that can carry both a column and a constraint name? The latter is a breaking change to a public type; the former adds a second per-field map.
[...]

PR #997, opened 2026-08-22
fix(auth): derive Auth lists from better-auth's own getAuthTables output

### LINKED ISSUE #987: auth: derive the auth lists from better-auth's own table definitions instead of hand-mirroring them
> *This was generated by AI during triage.* Raised by the maintainer while triaging #937: if a minor better-auth release can change the schema this much, is there a better way to drive list generation?

## The pattern

`deriveAuthLists` hand-writes the auth lists — every field, its column map, its type, its nullability, its index — as a manual mirror of better-auth's schema. Five separate issues found in recent triage are all the same defect class: **the mirror drifts, silently, and only a human reading better-auth's compiled output notices.**

| Issue | Drift |
| --- | --- |
| #935 | FK column generates as `user`, not `userId` |
| #937 | three `index: true` columns never emitted; ADR-0007's rationale was already false when written |
| #921 | unique constraint names can't be adopted |
| #981 | `returned: false` credential fields get no redaction |
| #986 | a *minor* upstream release adds a required column and a unique constraint |

Each has been fixed, or will be, by hand-editing the mirror to match what someone observed upstream. That closes the instance and leaves the mechanism intact.

## The proposal

better-auth exposes its resolved table definitions as a public API. `getAuthTables` (and `getAuthTablesWithResolvedIndexes`) are exported from `@better-auth/core/db` — better-auth's own Kysely migrator, schema generator, and adapter base all consume exactly this to build their schemas.

For a given options object it returns, per model: the model name, the table map, and for every field its column name (`fieldName`), type, requiredness, `references` (with `onDelete`), `unique`, `index`, `returned`, plus (in 1.7.x) a table-level `indexes` array.

That is a superset of what the hand-written mirror encodes. If derivation read it instead:

- **#935** — the column name comes from `fieldName`, no default to get wrong.
- **#937** — `index: true` is read directly; the ADR's premise can't silently go stale.
- **#981** — `returned: false` is available to drive redaction or field access.
- **#985** — the 1.7.x `indexes` array maps onto `db.indexes`, including composites.
- **#986** — a new upstream column appears in the generated schema automatically; the work becomes writing the migration, not discovering the drift.

The mirror stops being a thing that can be wrong.

## What this does not solve, and where it gets hard

Not a small change, and the obstacles are real:

- **Relationship modelling.** better-auth expresses the user FK as `references: { model, field, onDelete }` on a scalar field. The stack expresses it as a `relationship()` on both sides, with a synthetic reverse field. Translating one to the other — and keeping the reverse relations on `User` correct — is the substantive work.
- **Type mapping.** better-auth's `type` vocabulary (`string`, `boolean`, `date`, `number`, and the `bigInt`-ish rate-limit case) has to map onto the stack's field builders, including where the stack deliberately diverges (e.g. `password()` vs a plain string).
- **Dependency shape.** `@better-auth/core` is currently a transitive dependency, not a direct one of `packages/auth`, so it isn't resolvable from that package today. Either add it directly or reach the same export through `better-auth`. Needs deciding, and pins the version compatibility question: the returned shape itself evolves (the `indexes` array is new in 1.7.x), so the reader needs to tolerate a range.
- **The stack's own additions.** Where derivation currently adds things better-auth doesn't have — timestamps, access blocks, `extendUserList` merging — those stay hand-authored and must compose with the derived part.
- **Plugin-contributed schemas.** better-auth plugins contribute their own tables and fields, and `getAuthTables` already merges them. That is arguably an *upside* (plugin tables would derive for free, which they currently don't) but it widens the blast radius.
- **Determinism and review.** A hand-written mirror is greppable and diffable; a derived one changes when the dependency changes. That is the point, but it means schema changes can arrive with a `pnpm update`, so the generated output needs to be pinned by snapshot tests that fail loudly on upstream drift rather than absorbing it silently.

## Questions for triage

- **Full derivation, or derive-and-verify?** A cheaper intermediate: keep the hand-written lists, and add a test that calls `getAuthTables()` and asserts the derived lists match it on every dimension (column names, indexes, requiredness, references). That catches all five drift instances at CI time without restructuring derivation. It does not fix them automatically, but it converts silent drift into a failing build — most of the value for a fraction of the change.
- **If full derivation: incremental or wholesale?** Column names and indexes could be read from `getAuthTables()` while fields stay hand-declared, as a first step.
- **What happens to ADR-0007?** Its clean-diff promise is what all of this serves, and its fifth decision bullet is already being corrected in #937. If derivation lands, the ADR's framing ("we mirror better-auth's shape, here is each dimension we align") is superseded by "we read better-auth's shape". That likely wants a new ADR rather than an amendment.

**Recommendation for the triage discussion:** the derive-and-verify test is worth doing regardless of whether full derivation ever happens, and it is small. It would have caught every issue in the table above. Consider decoupling it from the larger restructure.

## Related

- #935, #937, #921, #981, #985, #986 — the drift instances motivating this
- ADR-0007 — the clean-diff promise
- ADR-0013 — application-authored access on plugin-injected lists; interacts with how much of a derived list stays configurable

borisno2:
## Summary

- `deriveAuthLists` (`packages/auth/src/config/derive-auth-lists.ts`) used to hand-transcribe the four core Auth lists (User/Session/Account/Verification, plus the conditional `RateLimit` fifth) from better-auth's schema — every field, its column map, type, nullability, and index typed out by hand. That mirror drifted repeatedly and silently: the FK column name (#935), three missing indexes (#937), unadoptable constraint names (#921), and a minor upstream release adding a required column (#986) — each caught by a human reading better-auth's compiled output, not a test.
- `deriveAuthLists` now reads better-auth's own resolved table definitions via `getAuthTables` (re-exported from `better-auth/db` — the same function better-auth's own Kysely migrator, schema generator, and adapter base consume) and translates them into stack list configs, so the derived lists cannot disagree with what better-auth itself declares.
- Scalar type/nullability/uniqueness/index/column-map/static-default and the user foreign key's column/`onDelete`/index all come from better-auth's field metadata through a small, fixed set of mapping rules. The one thing genuinely not present in that metadata — the reverse relation name (`User.sessions`/`accounts`) — is derived via pluralization, with a documented override map for a collision or bad pluralization (none needed yet).
- The app's existing `modelName`/`fields` overrides (`adoptBetterAuthTables`, a renamed model, a remapped column) pass straight through to `getAuthTables` as its own options object, so that override path is inherited for free.
- `deriveAuthLists`'s `{ keys, lists }` return contract is unchanged, so nothing downstream (plugin add-vs-extend logic, runtime user-key resolution, the `createUserList`/`createSessionList`/etc. wrappers in `src/lists/index.ts`) needed to change.
- Adds ADR-0033 recording the shift from hand-mirroring to derivation, and updates ADR-0007's FK-shape bullet to point at it (the shape itself is unchanged, only the mechanism producing it).
- Updates `packages/auth/CLAUDE.md`'s derivation section to describe the new mechanism.

## Test plan

- [x] `pnpm build` (core, cli, auth) — no type errors
- [x] `pnpm test` in `packages/auth` — all 272 tests pass unmodified, including the `generated-fk-shape.test.ts` (byte-level generated-schema assertions) and `auth-lists-drift.test.ts` (issue #988's drift detector against `getAuthTables`) suites — generated schema output is unchanged for existing projects
- [x] `pnpm test` in `packages/core` (1067 tests) and `packages/cli` (377 tests) — unaffected, all pass
- [x] `pnpm lint` — clean (2 pre-existing unrelated warnings)
- [x] `pnpm format` / `pnpm manypkg fix` — no changes needed beyond formatting already applied
- [x] Changeset added (`@opensaas/stack-auth` patch — internal refactor, no config/API surface change, no migration required)

Closes #987

---
_Generated by [Claude Code](https://claude.ai/code/session_01NCTp6XiTfh4JzusMGPJXrV)_


borisno2:
## Code review (self-review before merge)

Ran a focused review (effort: high) against `f6c78af`, the derivation rewrite. All 15 auth test files (272 tests) and `tsc` passed unmodified, confirming the "generated schema unchanged" claim for the paths already under test. The review surfaced three gaps outside that existing coverage:

1. **`assembleFields` silently dropped any better-auth field not named in the hardcoded `FIELD_ORDER` map** — reintroducing the exact "a minor upstream release adds a required column no one notices" failure class (#986) this derivation exists to close, just moved one layer down (from a missing hand-written field to a missing `FIELD_ORDER` entry).
2. **`buildScalarField` computed `isIndexed` for every scalar field but only wired it into the returned config for the `string` branch** — the `boolean`/`date`/`number` branches silently dropped an index/unique flag better-auth declared.
3. Missing an in-file note on the above (the root `CLAUDE.md` "Known limits" convention for generators).

Fixed in `6be00e8`:
- `assembleFields` now emits every derived field (scalar or FK) regardless of `FIELD_ORDER`, appending any field not named there after the pinned ones instead of dropping it — a future upstream field still reaches the generated schema, just at the end of the model until `FIELD_ORDER` is updated to place it.
- `isIndexed` is now passed through on all four scalar branches (`text`/`checkbox`/`timestamp`/`integer`/`bigInt`), not just `string`.

Re-ran the full auth suite after the fix — still 272/272 passing, `tsc` clean, `pnpm lint`/`pnpm format` clean.


---

PR #1000, opened 2026-08-22
fix(core): recover P2002 constraint info under Prisma 7 driver adapters

### LINKED ISSUE #979: P2002 unique-constraint violations do not expose which constraint fired in a stable place under Prisma 7 driver adapters
Raised from `borisno2/enrolm8`, where the shared guidance says to report a stack seam here rather than work around it silently.

## What we hit

An application enforcing several invariants through distinct unique indexes on one model needs to know **which** constraint a `P2002` violated, because different constraints mean materially different things — "this identity already exists", "you lost a race to succeed an active row", "this request was redelivered". Those map to different outcomes, and misclassifying one as another is a correctness bug that only manifests under concurrency.

The documented place to look is `error.meta.target`. Under Prisma 7 driver adapters **that field is empty**. The constraint name is instead nested under `meta.driverAdapterError.cause`.

It reproduces identically under `@prisma/adapter-pg` and under the PGlite adapter, so it is not an artefact of our test harness.

## Why it is worth surfacing

The failure mode is the bad kind: **correct-looking code that reads the documented field is silently wrong.** A handler that checks `meta.target` finds nothing, falls through to a catch-all, and classifies every violation as whichever case the author wrote first. Every single-threaded test passes, because reaching a second constraint requires concurrency or a deliberately crafted row.

The workaround available today — serialising `meta` and substring-matching a constraint name out of an adapter-specific error shape — is precisely the sort of thing our guidance says to escalate rather than absorb. It couples application code to an internal error structure that is free to change.

## What we would suggest, without assuming which is right

1. **Populate `error.meta.target`** from the adapter's constraint information (name, and columns where available), so the documented field means what it says regardless of adapter. This is the least surprising outcome for anyone following the Prisma docs.
2. **Or, if that field is Prisma's to own rather than the stack's**, expose a small helper — something like `uniqueConstraintOf(error)` — so callers have one stable, adapter-independent way to ask the question and stop reaching into `meta.driverAdapterError.cause`.
3. **Or, if neither is appropriate**, document the actual location. An accurate note is a legitimate resolution; what we would like to avoid is the current state, where the documented field exists, is empty, and gives no signal that it is the wrong place to look.

## Two related things we learned, in case they inform the fix

- **A derived constraint name truncates at 63 characters** (Postgres' identifier limit), so any index an application intends to branch on should be named explicitly. Worth a line wherever the above is documented.
- **One row can violate two unique indexes at once**, and Postgres reports whichever it checked first. So even with the constraint name available, the name alone is not always sufficient — a caller may still need to re-read the colliding row to decide what happened. That is inherent to Postgres rather than anything the stack can fix, but it affects what a helper should promise.

## Environment

- `@opensaas/stack-core` as consumed by `borisno2/enrolm8`
- Prisma 7 with driver adapters; reproduced under `@prisma/adapter-pg` and the PGlite adapter
- Multiple named unique indexes on a single tenant-owned model

Happy to supply a minimal reproduction against one of the examples — say which you would prefer.

borisno2:
## Summary

- Under Prisma 7 driver adapters (`@prisma/adapter-pg`, PGlite), a `P2002` unique-constraint error leaves `error.meta.target` `undefined` instead of populating it — the equivalent data instead sits at `error.meta.driverAdapterError.cause`. Core's `parsePrismaError` read only `meta.target`, so every unique violation under a driver adapter fell through to the generic `"A record with this value already exists"` message instead of the intended per-field errors.
- `parsePrismaError` now falls back to `meta.driverAdapterError.cause` when `meta.target` is absent, resolving both the violated column names (stripping Postgres' quoting of camelCase identifiers, e.g. `"tenantId"` → `tenantId`) and the constraint name (parsed from `cause.originalMessage`). An already-populated `meta.target` is left untouched — the documented path still wins.
- `DatabaseError` gains a `constraintName` field so the recovered constraint identity is reachable on the error the stack already raises.
- Added and exported `uniqueConstraintOf(error)` from `@opensaas/stack-core`, so any caller of `context.db.*` can resolve `{ fields, constraintName }` from a caught `P2002` without reaching into adapter internals themselves. `parsePrismaError` is implemented on top of the same helper.
- Never throws: every level of the adapter's error shape (`driverAdapterError`, `cause`, `constraint`, `fields`) is treated as optional and untrusted; missing or malformed data degrades to the pre-existing generic message.

Fixtures for the driver-adapter shape are taken from a verified repro captured during triage (PGlite behind a Postgres wire-protocol socket + `@prisma/adapter-pg@7.8.0`), not hand-invented.

## Test plan

- [x] `pnpm test` in `packages/core` (new tests: `tests/prisma-errors.test.ts` for the `uniqueConstraintOf` helper directly, and a new `describe` block in `tests/context.test.ts` exercising `parsePrismaError` via `serverAction`'s `createRelated` path for composite/single-column driver-adapter violations, camelCase quote-stripping, `meta.target`-already-populated precedence, no-recoverable-info fallback, and non-`P2002` errors)
- [x] `pnpm lint`
- [x] `pnpm format`
- [x] `pnpm manypkg fix`

Closes #979

---
_Generated by [Claude Code](https://claude.ai/code/session_01QZe1w8KNxYEjCGUYCdicbZ)_

### DIFF
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -203,6 +203,14 @@ function getDefaultData(listConfig: ListConfig<any>): Record<string, unknown> {
   return data
 }
 
+// A camelCase field name (e.g. a relationship's `tenantId` foreign key) needs
+// its word boundary split before title-casing, or it reads as one run-together
+// word ("Tenantid") in a user-facing unique-constraint message.
+function humanizeFieldName(fieldName: string): string {
+  const spaced = fieldName.replace(/([a-z0-9])([A-Z])/g, '$1 $2')
+  return spaced.charAt(0).toUpperCase() + spaced.slice(1)
+}
+
 // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
 function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
   if (
@@ -216,38 +224,30 @@ function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
 
     // P2002 is Prisma's unique constraint violation code.
     if (prismaError.code === 'P2002') {
-      const constraintInfo = uniqueConstraintOf(prismaError)
-      const target = constraintInfo?.fields
-      const constraintName = constraintInfo?.constraintName
+      const target = uniqueConstraintOf(prismaError)?.fields
       const fieldErrors: Record<string, string> = {}
 
       if (target && target.length > 0) {
         for (const fieldName of target) {
           const fieldConfig = listConfig.fields[fieldName]
-          const label = fieldName.charAt(0).toUpperCase() + fieldName.slice(1)
 
           if (fieldConfig) {
-            fieldErrors[fieldName] = `This ${label.toLowerCase()} is already in use`
+            fieldErrors[fieldName] =
+              `This ${humanizeFieldName(fieldName).toLowerCase()} is already in use`
           } else {
             fieldErrors[fieldName] = `This value is already in use`
           }
         }
 
-        const fieldLabels = target.map((f) => f.charAt(0).toUpperCase() + f.slice(1)).join(', ')
+        const fieldLabels = target.map(humanizeFieldName).join(', ')
         return new DatabaseError(
           `${fieldLabels} must be unique. The value you entered is already in use.`,
           fieldErrors,
           prismaError.code,
-          constraintName,
         )
       }
 
-      return new DatabaseError(
-        'A record with this value already exists',
-        {},
-        prismaError.code,
-        constraintName,
-      )
+      return new DatabaseError('A record with this value already exists', {}, prismaError.code)
     }
 
     return new DatabaseError(
--- a/packages/core/src/context/index.ts
+++ b/packages/core/src/context/index.ts
@@ -14,6 +14,7 @@ import {
 import type { DeclaredOnlyTree } from '../access/index.js'
 import { ValidationError, DatabaseError } from '../hooks/index.js'
 import { getDbKey } from '../lib/case-utils.js'
+import { uniqueConstraintOf } from '../lib/prisma-errors.js'
 import type { PrismaClientLike } from '../access/types.js'
 import { buildInclude, pickFields, isFragment, buildFieldSelectionScope } from '../query/index.js'
 import type { FieldSelection, FieldSelectionScope } from '../query/index.js'
@@ -215,10 +216,12 @@ function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
 
     // P2002 is Prisma's unique constraint violation code.
     if (prismaError.code === 'P2002') {
-      const target = prismaError.meta?.target
+      const constraintInfo = uniqueConstraintOf(prismaError)
+      const target = constraintInfo?.fields
+      const constraintName = constraintInfo?.constraintName
       const fieldErrors: Record<string, string> = {}
 
-      if (target && Array.isArray(target)) {
+      if (target && target.length > 0) {
         for (const fieldName of target) {
           const fieldConfig = listConfig.fields[fieldName]
           const label = fieldName.charAt(0).toUpperCase() + fieldName.slice(1)
@@ -235,10 +238,16 @@ function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
           `${fieldLabels} must be unique. The value you entered is already in use.`,
           fieldErrors,
           prismaError.code,
+          constraintName,
         )
       }
 
-      return new DatabaseError('A record with this value already exists', {}, prismaError.code)
+      return new DatabaseError(
+        'A record with this value already exists',
+        {},
+        prismaError.code,
+        constraintName,
+      )
     }
 
     return new DatabaseError(
--- a/packages/core/src/lib/prisma-errors.ts
+++ b/packages/core/src/lib/prisma-errors.ts
@@ -1,0 +1,78 @@
+type PrismaP2002Meta = {
+  target?: string[]
+  driverAdapterError?: {
+    cause?: {
+      originalMessage?: string
+      constraint?: { fields?: unknown[] }
+    }
+  }
+}
+
+// Postgres quotes an identifier in `constraint.fields` only when it needed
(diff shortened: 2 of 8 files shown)


packages/core/src/context/index.ts
```diff
@@ -235,10 +238,16 @@ function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
...
+          constraintName,
         )
       }
 
-      return new DatabaseError('A record with this value already exists', {}, prismaError.code)
+      return new DatabaseError(
+        'A record with this value already exists',
+        {},
+        prismaError.code,
+        constraintName,
```
borisno2: **Correctness/completeness: `constraintName` never reaches any caller of `serverAction`.**

`parsePrismaError` computes `constraintName` and threads it into every `DatabaseError` it constructs (lines 241 and 249), but every catch site that consumes the result only reads `.message`/`.fieldErrors` off it:

- `bulkAction` (≈538): `{ bulkAction: false, error: dbError.message }`
- `removeRelated` (≈585): `{ removed: false, error: dbError.message }`
- `createRelated` (≈645-649): `{ created: false, error: dbError.message, fieldErrors: … }`
- `updateRelated` (≈673-677): same shape
- the general create/update/delete path (≈746-758): same shape

`DatabaseError` itself also isn't exported from the package's public surface (only `ValidationError` is, in `src/index.ts`), so no external consumer can `instanceof`/read `.constraintName` off a caught error either.

**Failure scenario:** a Postgres unique violation under a driver adapter hits P2002 via `ItemFormClient`'s `create`/`update` serverAction call. The fix correctly derives `constraintName` (e.g. `'thing_tenant_slug_key'`) and attaches it to the `DatabaseError`, but the Admin UI (and any other `serverAction` caller) can never see it — the value is computed and immediately discarded. If the intent (per the PR description, "so the recovered constraint identity is reachable on the error the stack already raises") was for this to be usable from the primary write path, it isn't wired up; if it's meant only for the separately-exported `uniqueConstraintOf()` helper on raw `context.db.*` errors, then this `DatabaseError` plumbing is dead weight that should either be removed or actually surfaced in the returned result objects.

---

packages/core/src/lib/prisma-errors.ts
```diff
@@ -0,0 +1,78 @@
...
+    return { fields: meta.target }
+  }
+
+  const cause = meta?.driverAdapterError?.cause
+  const rawFields = cause?.constraint?.fields
+  const fields = Array.isArray(rawFields)
+    ? rawFields.filter((f): f is string => typeof f === 'string').map(stripIdentifierQuotes)
+    : []
+
+  const constraintName =
```
borisno2, line 78: **Robustness: `constraintName` extraction depends on Postgres running with English (`lc_messages`) error text.**

`cause.originalMessage.match(/unique constraint "([^"]+)"/)` only matches Postgres's default English wording (`duplicate key value violates unique constraint "…"`). A server configured with a non-English `lc_messages` locale (a legitimate, non-exotic production setting) emits a fully localized message, so this regex silently fails to match and `constraintName` comes back `undefined` — with no warning anywhere that this happened. `fields` still resolves correctly in that case (it comes from the structured `cause.constraint.fields`, not text), so this is scoped to the constraint-name half of the fix, but it's a real gap the "Known limits"-style doc comment above (truncation, ambiguous multi-index hits) doesn't mention. Worth calling out explicitly as a known limitation, the same way the truncation/ambiguity caveats already are.

---

packages/core/src/context/index.ts
```diff
@@ -215,10 +216,12 @@ function parsePrismaError(error: unknown, listConfig: ListConfig<any>): Error {
...
+      const constraintInfo = uniqueConstraintOf(prismaError)
+      const target = constraintInfo?.fields
+      const constraintName = constraintInfo?.constraintName
       const fieldErrors: Record<string, string> = {}
 
-      if (target && Array.isArray(target)) {
+      if (target && target.length > 0) {
         for (const fieldName of target) {
           const fieldConfig = listConfig.fields[fieldName]
           const label = fieldName.charAt(0).toUpperCase() + fieldName.slice(1)
```
borisno2: **Pre-existing latent bug newly reachable via this fix: camelCase field labels get mangled.**

A few lines down (line 230, not touched by this diff, so not directly commentable): `fieldErrors[fieldName] = \`This ${label.toLowerCase()} is already in use\`` collapses a camelCase field name to a run-together lowercase string. Before this PR, `target` was always empty under a driver adapter (that's the whole bug being fixed), so this branch was effectively dead for Postgres driver-adapter users. Now that `uniqueConstraintOf` recovers real column names (e.g. `tenantId` from the composite-constraint fixture in the tests), a unique violation on a real camelCase scalar field will produce a message like "This tenantid is already in use" instead of something readable. Worth fixing alongside this change since this PR is precisely what makes that code path live again for the users it targets.

---

borisno2 (commented):
Single-pass review (Agent tool unavailable in this context — no multi-agent fan-out or subagent verify pass; findings below are from one careful read-through). 3 findings posted inline.

---

packages/core/src/context/index.ts
borisno2: Fair — the `constraintName` plumbing through `DatabaseError` was dead weight since `DatabaseError` isn't exported and no catch site read it. Removed it from `DatabaseError`/`parsePrismaError` entirely in 0c335db and rely solely on the exported `uniqueConstraintOf(error)` helper, which is the actually-reachable path for a caller to get constraint identity (from a raw error caught off `context.db.*`).

---

packages/core/src/lib/prisma-errors.ts
borisno2: Good catch — added this to the `UniqueConstraintInfo.constraintName` doc comment alongside the truncation/ambiguity caveats in 0c335db. There's no structured alternative available from the adapter (the issue's own triage confirmed the name is only ever free text in `originalMessage`), so this documents it as a known limit rather than attempting a fix — `fields` still resolves correctly since that part is structured data.

---

packages/core/src/context/index.ts
borisno2: Agreed, fixed in 0c335db — added a `humanizeFieldName` helper that splits camelCase word boundaries before title-casing, so `tenantId` reads as "tenant id" instead of "tenantid". Added a test (`context.test.ts`) covering a camelCase field recovered from a driver-adapter error.

---

PR #1001, opened 2026-08-22
Fix include on a to-one relationship throwing PrismaClientValidationError (#974)

### LINKED ISSUE #974: include on a to-one relationship generates a nested where Prisma rejects
## What happens

`context.db.<list>.findMany({ include: { <toOneRelation>: true } })` throws `PrismaClientValidationError`. The context injects the related list's access filter as a `where` **inside** the `include`, and Prisma only accepts `where` on a to-**many** include.

Observed on `@opensaas/stack-core@0.39.2` with `@prisma/client@7.9.1`.

## Reproduction

Two lists, where `Child.owner` is a to-one relationship to a list whose `query` access resolves to a filter (rather than a plain boolean):

```ts
await context.db.child.findMany({
  where: { studioId: { equals: studioId } },
  include: { owner: true },
})
```

The generated call is:

```
model.findMany({
  where: { AND: [ ... ] },
  include: {
    owner: {
      where: {          // <- Prisma rejects this
        OR: [ { id: { equals: '<sessionUserId>' } }, { memberships: { some: { ... } } } ]
      }
    }
  }
})
```

In our case the relation is to the auth user list, whose `query` gate is membership-derived ("self, or a user in a Studio I administer"), so the filter is non-trivial and always present.

## Expected

Either:

1. the scoping is applied in a shape Prisma accepts for a to-one include — the stack already knows the relation's arity from the config; or
2. the relation is fetched and then post-filtered to `null` when the access filter excludes it, matching the stack's own "access denial is silent" contract (a denied to-one read returning `null` rather than throwing).

Option 2 seems more consistent with the rest of the context: everywhere else a denied read is silent, so this is the one place a caller has to know about access control to avoid an exception.

## Workaround

Drop the `include` and read the related rows with a second `findMany({ where: { id: { in: [...] } } })`, which is scoped correctly and stays silent on denial. Works, but it is one extra round trip and the failure mode gives no hint that this is the fix — the error points into generated stack code, not the caller's query.

## Context

Found in enrolm8 (borisno2/enrolm8#718) rendering "reversed by <name>" beside a Payment reversal.

borisno2:
## Summary

- `context.db.<list>.findMany({ include: { <toOneRelation>: true } })` threw `PrismaClientValidationError` whenever the related list's `query` access resolved to a filter (rather than a plain boolean) — the access-scoped include builder attached that filter as a nested `where`, which Prisma only accepts on a **to-many** include.
- `buildAccessScopedInclude` (`access-filter.ts`) no longer attaches a `where` to a to-one relation's include entry. Instead it records the access filter (or an outright denial) in a new `toOneAccessFilters` tree.
- `resolveToOneAccessVisibility` (new) resolves that tree **after** the Prisma query runs: one batched `id IN (...)` existence check per relation across every row in the read (never once per row), evaluated by handing the exact `PrismaFilter` `checkAccess` produced straight to Prisma — no hand-rolled filter interpretation, so relation-quantifier filters (e.g. `{ memberships: { some: {...} } }`, the reporter's real case) are evaluated correctly.
- `filterReadableFields` (`field-visibility.ts`) applies the resolved visibility: an excluded related row (or an outright-denied relation) becomes `null`, matching the "denied read is silent" contract everywhere else in the context — the parent row is still returned. This holds at every include depth, including a to-one relation nested inside another include.
- To-many relations are untouched — they still receive their access filter as a nested `where`, exactly as before.

Chosen approach follows the maintainer decision recorded on the issue (option 2: fetch + post-filter to `null`), using existence-check batching (issue's option 1 shape) rather than a per-row query.

## Test plan

- [x] New unit coverage in `access-filter.test.ts` for `buildAccessScopedInclude`'s to-one behavior (scoped/open/denied/nested) and `resolveToOneAccessVisibility` (denial passthrough, empty-id skip, batching across rows, nested-tree flattening through a to-many hop).
- [x] New end-to-end coverage in `tests/to-one-include-access-filter.test.ts` through `context.db` (`findMany`/`findUnique`/`findFirst`), including a membership-derived (`OR`/relation-quantifier) filter matching the issue's real-world case, and a to-one relation nested inside a to-many include.
- [x] Updated existing tests that asserted the old (buggy) `where`-on-to-one shape to reflect the new behavior.
- [x] `pnpm lint`, `pnpm format`, `pnpm manypkg fix`, and the full core test suite (1080 tests) pass.
- [x] `tsc --noEmit` and `pnpm build` pass for `packages/core`.
- [x] Changeset added (patch, notes the behavior change: a query that previously threw now succeeds and may return `null` for a related row).

Closes #974

---
_Generated by [Claude Code](https://claude.ai/code/session_012JTajjY3WdXYv95Y2UXBsS)_

### DIFF
--- a/packages/core/src/access/access-filter.ts
+++ b/packages/core/src/access/access-filter.ts
@@ -9,6 +9,7 @@ import {
   resolveQueryField,
   walkWhereReadAccess,
 } from './query-validation.js'
+import { getDbKey } from '../lib/case-utils.js'
 
 /**
  * Access Filter — phase 1 of the two-phase read (pre-query).
@@ -34,6 +35,26 @@ import {
  * named a nested `include` there too. A relation nobody named never has its
  * list's `query` access evaluated at all — there is no separate "build the
  * whole tree, then reconcile against what was asked for" pass to walk it.
+ *
+ * **To-one relations are scoped after the query, not inside it (issue #974).**
+ * Prisma accepts a nested `where` on an `include` entry only for a to-**many**
+ * relation; the same shape on a to-**one** relation raises
+ * `PrismaClientValidationError`. So for a to-one relation whose related
+ * list's `query` access resolves to a filter, `buildAccessScopedInclude`
+ * does NOT attach that filter as `where` — it fetches the row unscoped and
+ * records the filter in the returned `toOneAccessFilters` tree instead.
+ * `resolveToOneAccessVisibility` (below) turns that tree into the set of
+ * related ids the session may actually see, via ONE batched `id IN (...)`
+ * existence check per (relation, nesting level) across every row in the
+ * read — never a per-row query, and never a hand-rolled evaluation of the
+ * access filter (it is handed to Prisma exactly as `checkAccess` produced
+ * it). `field-visibility.ts`'s `filterReadableFields` is where the result
+ * actually becomes `null` for a row the check excludes — see its module doc.
+ * A to-one relation whose related list's `query` access is `true` (no
+ * filter) or `false` (denied outright) needs no existence check at all: the
+ * former is left as `true` in `include`, unchanged from before; the latter
+ * is recorded as `{ kind: 'denied' }`, and `field-visibility.ts` forces the
+ * key to `null` without ever asking Prisma for it.
  */
 
 /** The structured (object) form of a relation include entry — caller/fold-supplied or produced by this module. */
@@ -81,6 +102,38 @@ function andWhere(
   return accessWhere ?? callerWhere
 }
 
+/** One to-one relation's recorded access filter, or an outright denial — see the module doc's "To-one relations" section. */
+export type ToOneAccessFilterEntry =
+  { kind: 'scoped'; relatedListName: string; accessWhere: PrismaFilter } | { kind: 'denied' }
+
+/**
+ * Which to-one relations, at which nesting level of an `include`, need a
+ * post-query existence check rather than a Prisma-side `where` — because
+ * their related list's `query` access resolved to a filter (`kind: 'scoped'`)
+ * or a denial (`kind: 'denied'`) and Prisma cannot express either as a nested
+ * `where` on a to-one include. `resolveToOneAccessVisibility` consumes this
+ * tree; `filterReadableFields` (`field-visibility.ts`) applies its result.
+ */
+export type ToOneAccessFilterTree = {
+  /** To-one relation keys at THIS level needing a post-query check. */
+  filters: Record<string, ToOneAccessFilterEntry>
+  /** Per-key trees for relations present in the include for other reasons, whose own nested include may contain further to-one filters. */
+  nested: Record<string, ToOneAccessFilterTree>
+}
+
+export function emptyToOneAccessFilterTree(): ToOneAccessFilterTree {
+  return { filters: {}, nested: {} }
+}
+
+function isToOneAccessFilterTreeEmpty(tree: ToOneAccessFilterTree): boolean {
+  return Object.keys(tree.filters).length === 0 && Object.keys(tree.nested).length === 0
+}
+
+/** Whether a relationship field is to-one (at most one related row) rather than to-many. */
+function isToOneRelationship(fieldConfig: FieldConfig): boolean {
+  return !('many' in fieldConfig && fieldConfig.many === true)
+}
+
 /**
  * Build the access-scoped `include` for exactly the relations a read
  * requested, recursing only into branches `requestedInclude` itself names.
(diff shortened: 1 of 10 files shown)


packages/core/src/access/access-filter.ts
```diff
@@ -141,41 +202,154 @@ export async function buildAccessScopedInclude(
...
+
+    if (ids.size === 0) {
+      resolved.filters[key] = { kind: 'visible', ids: new Set() }
+      continue
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- dynamic model access by list name, mirroring the rest of the read pipeline
+    const model = (args.context.prisma as any)[getDbKey(entry.relatedListName)]
+    const visibleRows = await model.findMany({
+      where: { AND: [entry.accessWhere, { id: { in: [...ids] } }] },
```
borisno2, line 332: **Bug: the batched existence check breaks for a to-one relation whose target is a Singleton list.**

`ids` is built by `String()`-coercing every related id (line 320), and that stringified set is fed straight into `id: { in: [...ids] }` here. But per `generator/types.ts` (ADR-0004), a Singleton list gets a bare `id Int @id`, not the usual `String @id @default(cuid())` — its id is a `number`, not a `string`.

A to-one relationship can legitimately target a Singleton list (e.g. `Post.settings -> Settings` where `Settings` is a singleton config record) — this is exactly the natural direction for referencing a singleton, and nothing in the config/generator forbids it. If that singleton's `query` access resolves to a filter (not a plain boolean), `buildAccessScopedInclude` records it as `kind: 'scoped'` and this function runs `model.findMany({ where: { AND: [accessWhere, { id: { in: ["1"] } }] } } )` against an `Int` id column with a `string` value — Prisma's client validates the argument shape and throws `PrismaClientValidationError` (expected `Int`, got `String`).

This reintroduces, for singleton to-one targets, the exact class of error (`PrismaClientValidationError` on a to-one include) that this PR sets out to fix.

Fix: don't coerce with `String(...)` before querying — collect the raw `id` values (whatever their native type) for the `in` filter, and use `String()` only when building the `visibleIds` set used for the `.has()` comparison in `field-visibility.ts` (which is fine, since that's a JS-side membership check, not a Prisma argument).

---

packages/core/src/access/access-filter.ts
```diff
@@ -141,41 +202,154 @@ export async function buildAccessScopedInclude(
...
+  nested: Record<string, ToOneAccessVisibilityTree>
+}
+
+export function emptyToOneAccessVisibilityTree(): ToOneAccessVisibilityTree {
+  return { filters: {}, nested: {} }
+}
+
+/**
+ * Resolve a `ToOneAccessFilterTree` against the RAW rows Prisma already
+ * fetched (unscoped for the flagged to-one relations — see
```
borisno2, line 278: **Efficiency: independent existence checks/subtrees are awaited sequentially instead of in parallel.**

Both loops in `resolveToOneAccessVisibility` — the `for (const [key, entry] of Object.entries(tree.filters))` loop (each iteration does an independent `model.findMany` against its own list/filter) and the `for (const [key, nestedTree] of Object.entries(tree.nested))` loop below it (each a fully independent recursive resolution) — `await` inside the loop body, so N independent round trips run one after another instead of concurrently.

A read with two filtered to-one relations at the same level (e.g. `include: { author: true, category: true }`, both access-scoped) now pays two sequential DB round trips where they could run via `Promise.all` and complete in parallel — same for sibling nested branches. Worth batching with `Promise.all(Object.entries(...).map(...))` for both loops.

---

borisno2 (commented):
Single-pass review (high effort, Agent tool unavailable — no multi-agent fan-out or subagent verify pass ran; treat findings as unverified by a second pass). Traced the two-phase read pipeline (`buildAccessScopedInclude` → `resolveToOneAccessVisibility` → `filterReadableFields`), the three read entry points (`findUnique`/`findMany`/`get`), and the accompanying test coverage in detail. Found one correctness bug and one efficiency issue, posted inline. No other issues survived re-checking against the diff (the caller-`where`-on-to-one drop, the null-relation `typeof` branch, and the denied/scoped tree-flattening recursion were all checked and are correct as written).

---

packages/core/src/access/access-filter.ts
borisno2: Good catch — fixed in a34cdd8. The id set is now keyed by `String(id)` for de-duplication but the value preserved (and sent to Prisma in the `in` filter) is the raw id in its original type, so an `Int` singleton id is no longer coerced to a string. Added a regression test pinning this.

---

packages/core/src/access/access-filter.ts
borisno2: Fixed in a34cdd8 — both loops now build their per-key work as an array of promises and run it through `Promise.all`, so sibling filters and sibling nested branches resolve concurrently instead of one await at a time. Added a test asserting two independent existence checks interleave rather than serialize.

---

PR #1002, opened 2026-08-22
fix(core,ui,auth): admin UI URL round-trip for non-PascalCase list keys

### LINKED ISSUE #991: auth/ui: better-auth plugin lists get camelCase keys, so their admin UI URLs don't round-trip
> *This was generated by AI during triage.* Found while checking a claim made during #987's triage — the claim was wrong, this is what was actually there.

## Summary

Lists derived from a better-auth plugin's schema take their key from the plugin's own `modelName`, which is camelCase. That violates the repo's PascalCase list-key convention, and it breaks the admin UI's URL round-trip: the route for such a list resolves to a list key that does not exist.

## Current behavior

`convertBetterAuthSchema` derives each list key as:

```
resolveBaseModelKey(tableName, baseModelKeys) || tableSchema.modelName || toPascalCase(tableName)
```

For a base model the first branch wins. For a *plugin* table there is no base-model remap, so `tableSchema.modelName` wins — and better-auth's plugins declare camelCase model names. The `toPascalCase` fallback only applies when a plugin omits `modelName` entirely.

Confirmed against generated schema output for `authPlugin({ betterAuthPlugins: [mcp({ loginPage: '/sign-in' })] })`:

```
MODELS: User, Session, Account, Verification, oauthApplication, oauthAccessToken, oauthConsent
```

Three camelCase Prisma models, against `CLAUDE.md`'s stated conventions — *"List Names in Config: Always use PascalCase"* and *"Prisma Models: PascalCase"*.

## Why it matters — the URL round-trip breaks

The admin UI resolves a list from its URL segment via `getListKeyFromUrl`, then indexes `config.lists` with the result. For a camelCase key that does not round-trip:

```
listKey=oauthApplication  dbKey=oauthApplication  urlKey=oauth-application
getListKeyFromUrl('oauth-application') -> 'OauthApplication'   // ≠ 'oauthApplication'
roundTrips=false
```

So the link the UI generates for the list (`getUrlKey` → `oauth-application`) resolves back to `OauthApplication`, which is not a key in `config.lists`. The lookup misses and the route falls through to the dashboard branch rather than opening the list.

`getDbKey` is unaffected — a camelCase key is already the db key — so `context.db.oauthApplication` works fine. This is specifically a UI routing defect, not a data-access one.

**Severity is bounded** by ADR-0013: plugin-injected lists ship closed, so these lists don't surface in the admin UI at all unless an application grants access. The break is latent until someone does — at which point the list is visible in navigation but its route doesn't open.

**Not verified:** the exact user-visible symptom (silent dashboard fallback vs. an error) and whether navigation links to these lists render at all when access is granted. Worth confirming with a real admin UI run before deciding severity — the routing miss is proven, the presentation of it is not.

## Suggested direction

Options, with a decision needed:

1. **PascalCase the plugin-derived list key** regardless of what `modelName` says, keeping the table name mapped via `@@map` so the physical table still matches better-auth. Fixes the round-trip and honours the stated convention. Changes the generated model names for anyone already using a better-auth plugin — a schema change needing a migration note.
2. **Make the case utilities round-trip-safe** for keys that aren't PascalCase, e.g. by resolving the URL segment against actual `config.lists` keys rather than reconstructing a key from the string. More general, and would harden every list key against this class of bug, but is a core change rather than an auth one.
3. **Both** — (2) as the robust fix, (1) for convention compliance.

Option 2 is the more interesting one: `getListKeyFromUrl` reconstructing a key by string transformation is inherently lossy, and any list key that isn't strict PascalCase hits the same wall. That suggests the bug is really in core's URL resolution, with the auth plugin merely the first caller to expose it.

## Note on existing coverage

`packages/auth/tests/schema-converter.test.ts` asserts `lists` has property `OauthApplication` (PascalCase) — but it reaches that via the `toPascalCase(tableName)` fallback with a snake_case table name and no `modelName`, which is not the path the real MCP plugin takes. So current coverage passes while the real plugin produces `oauthApplication`. Any fix should add a case that exercises the `modelName`-present path.

## Correcting the record

This issue exists because of a claim I made in [#987's triage notes](https://github.com/OpenSaasAU/stack/issues/987#issuecomment-5377376565): that *"the stack currently derives no models for better-auth plugin tables, so an MCP-enabled app has no Prisma models for its OAuth tables."* **That is false** — `authPlugin` has converted plugin schemas into lists all along, and the OAuth models do reach the generated schema, as the output above shows. A correction is posted on #987.

## Related

- #987 — where the incorrect claim was made; plugin-table derivation is an *existing* capability, which strengthens rather than weakens that issue's staging argument
- ADR-0013 — plugin-injected lists ship closed; the bounding condition on severity

borisno2:
## Summary

- `packages/core`: add `resolveListKeyFromUrl(urlSegment, listKeys)` — a config-aware companion to `getListKeyFromUrl` that matches a URL segment against a list of actual list keys via `getUrlKey` (the same helper that builds the URL) instead of reconstructing a key by string rules. Returns `undefined` for an unmatched segment; throws if two list keys would collide on the same URL segment. `getListKeyFromUrl` itself is unchanged, still exported, and untouched by existing tests.
- `packages/ui`: `AdminUI`'s route resolution now uses `resolveListKeyFromUrl(urlSegment, Object.keys(config.lists))` instead of `getListKeyFromUrl(urlSegment)`. A list keyed with anything other than strict PascalCase (e.g. a camelCase key derived from a better-auth plugin) is now reachable from its own navigation link. An unresolved segment falls back to the raw segment string so the existing "List not found" branches (which key off `config.lists`) still render, instead of silently falling through to the dashboard.
- `packages/auth`: `convertBetterAuthSchema` now PascalCases a better-auth plugin's camelCase `modelName` when deriving a list key (`oauthApplication` → `OauthApplication`, `rateLimit` → `RateLimit`), matching the repo's PascalCase list-key convention and fixing the round-trip bug at the source for these lists. The original camelCase name is preserved via `db.map` (`@@map`) so the physical database table is unaffected.

### Why this was broken

`getListKeyFromUrl` reconstructs a list key from a URL segment purely by string transformation (kebab → Pascal). That's lossy for any list key that isn't strict PascalCase. A better-auth plugin's derived list (e.g. the `mcp` plugin's `oauthApplication` OAuth table) is exactly such a key: it appeared correctly in navigation and its link was built correctly (`getUrlKey('oauthApplication')` → `/oauth-application`), but resolving that same URL back (`getListKeyFromUrl('oauth-application')` → `'OauthApplication'`) produced a key that doesn't exist in `config.lists`, so the route rendered "List not found" for a list that does exist. See #991 for the full trace.

## Test plan

- [x] `packages/core`: new tests for `resolveListKeyFromUrl` (`case-utils.test.ts`) — PascalCase and camelCase resolution, unmatched segment, ambiguous-segment throw, and a round-trip check against `getUrlKey` for a set of list keys
- [x] `packages/ui`: new `AdminUIListKeyResolution.test.tsx` — a camelCase list key resolves for its list/create/edit routes, a PascalCase list key still resolves as before, root still renders the dashboard, and an unknown segment still falls through to the list's own "not found" state rather than throwing or redirecting to the dashboard
- [x] `packages/auth`: new tests in `schema-converter.test.ts` — a camelCase plugin `modelName` (including `rateLimit`) is PascalCased and mapped back to the original table name via `db.map`; an already-PascalCase `modelName` gets no `db.map`; a `baseModelKeys` remap is unaffected
- [x] `pnpm lint` passes (only 2 pre-existing, unrelated warnings)
- [x] `pnpm format` and `pnpm manypkg fix` applied, no unexpected diffs
- [x] Full test suites pass: core (1105 tests), auth (278 tests, 2 pre-existing skips), ui (585 tests)
- [x] Changeset added (`@opensaas/stack-core` minor, `@opensaas/stack-ui` patch, `@opensaas/stack-auth` patch) — the auth changeset calls out the schema-affecting model-name change and that the physical table is unaffected

Closes #991

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wz9vP6EhcxhasnQRYxFA7r)_

### DIFF
--- a/.changeset/wise-hounds-listen.md
+++ b/.changeset/wise-hounds-listen.md
@@ -1,0 +1,22 @@
+---
+'@opensaas/stack-core': minor
+'@opensaas/stack-ui': patch
+'@opensaas/stack-auth': patch
+---
+
+Fix admin UI URL round-trip for a list keyed with anything other than strict PascalCase (issue #991). `getListKeyFromUrl` reconstructs a list key by string transformation, which is lossy for a non-PascalCase key — a real example is a better-auth plugin's derived list (e.g. `oauthApplication`, from the `mcp` plugin's OAuth tables). Such a list appeared in navigation but its own link resolved to a key that did not exist in `config.lists`, rendering "List not found".
+
+`@opensaas/stack-core` adds `resolveListKeyFromUrl(urlSegment, listKeys)` alongside the existing `getListKeyFromUrl`, which is unchanged and still exported. The new resolver matches a URL segment against the config's actual list keys via `getUrlKey` — the same helper that builds the URL — instead of reconstructing one, so route lookup and URL generation cannot drift apart. It returns `undefined` for a segment matching no list (so callers keep rendering their existing "not found" state), and throws if two distinct list keys would produce the same URL segment.
+
+```typescript
+import { resolveListKeyFromUrl } from '@opensaas/stack-core'
+
+resolveListKeyFromUrl('oauth-application', Object.keys(config.lists)) // 'oauthApplication'
+resolveListKeyFromUrl('does-not-exist', Object.keys(config.lists)) // undefined
+```
+
+`@opensaas/stack-ui`'s `AdminUI` now uses `resolveListKeyFromUrl` for its route resolution, fixing the broken link for any such list.
+
+`@opensaas/stack-auth`'s `convertBetterAuthSchema` now PascalCases a better-auth plugin's camelCase `modelName` when deriving a list key (`oauthApplication` → `OauthApplication`, `rateLimit` → `RateLimit`), matching the repo's PascalCase list-key convention and fixing the same round-trip bug at the source for these lists.
+
+**Schema-affecting for `@opensaas/stack-auth` users with a better-auth plugin that declares extra tables** (e.g. `mcp`'s OAuth tables, or `rateLimit.storage: 'database'` with no `modelName` remap configured): the generated Prisma **model name** changes to match the new PascalCase list key. The physical **table name** does not change — the previous camelCase name is preserved via `db.map` (`@@map`) — so `prisma db push` / `prisma migrate dev` sees a model rename, not a table rename, and `context.db.oauthApplication` (the camelCase db accessor) keeps working unchanged. Regenerate (`pnpm generate`) and re-run your migration/push step after upgrading.
--- a/packages/auth/src/server/schema-converter.ts
+++ b/packages/auth/src/server/schema-converter.ts
@@ -93,10 +93,15 @@ function convertField(
  * Per ADR-0013, ships closed — no access is set here. See
  * `packages/auth/CLAUDE.md` ("Access control on Auth lists") for how an app
  * grants access to a table this produces.
+ *
+ * @param tableMap - Physical table name for `db.map` (`@@map`), when the list
+ *   key had to change case to satisfy the PascalCase convention and would
+ *   otherwise rename the live table (issue #991).
  */
 export function convertTableToList(
   tableName: string,
   tableSchema: BetterAuthTableSchema,
+  tableMap?: string,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any -- ListConfig must accept any TypeInfo
 ): ListConfig<any> {
   const fields: Record<string, FieldConfig> = {}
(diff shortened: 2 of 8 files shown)


borisno2:
The `Vercel` status failure is unrelated to this PR: it's the `stack-docs` project hitting Vercel's free-tier daily deployment quota (`api-deployments-free-per-day`), not a build or test failure caused by this diff — nothing here touches the docs site. No action to take on our side; it should clear once the quota resets. The changesets-bot bump to `minor` across all `@opensaas/stack-*` packages is also expected — this repo releases them as a fixed group (`.changeset/config.json`), so any package minor bump (here, `@opensaas/stack-core`) promotes the whole group. Waiting on the `test`/`e2e` GitHub Actions checks, which are still running.

---

PR #1003, opened 2026-08-22
feat(core): add context.withSession() alongside sudo()

### LINKED ISSUE #980: Missing seam: `sudo()` has no sibling for substituting a session on an existing context
## The gap

`StackContext` exposes `sudo()` — "bypass access control, keep the session, keep running hooks". There is no counterpart for the other axis: **derive a context that keeps this context's client and config but carries a DIFFERENT session.**

That axis matters because list `validate` hooks read `args.context.session`, and `sudo()` deliberately does not satisfy a hook. So any caller that is legitimately authorised but arrives without the session a hook wants — an unattended dispatcher running a committed intent, a system/service principal, a job runner — has no supported way to present one. `sudo()` is the wrong tool: it erases the check rather than answering it.

## Why it cannot be worked around cleanly today

The obvious workaround is to rebuild the context: `getContext(config, prisma, session)`. Both inputs are out of reach at the point that needs them.

1. **The resolved config is not obtainable from a context.** `config` may be a `Promise` when plugins are present, so the only resolved copy lives in the generated `.opensaas/context.ts` module scope (`resolvedConfig`), which is not exported. A consumer holding a `StackContext` cannot get back to the config that built it.
2. **The client is a process singleton in the generated factory.** `getContext()` there builds and memoises one `PrismaClient`. A caller that must reuse *its own* client — notably a test suite running against an in-process PGlite instance, but equally anything inside `context.transaction()` — cannot use it, and rebuilding a second client is a second connection with none of the first's transaction state.

`StackContext` exposes `prisma`, so half the inputs are reachable; the config half is not, and it is the half that cannot be re-derived.

## What we did instead

We are shipping the substitution as an **injected factory**, threaded through three call layers (cron route → maintenance sweep → outbox sweep → dispatch), typed as:

```ts
export type SystemContextFactory = (args: { session: Session }) => Promise<StackContext>
```

Each entry point supplies it from whatever built its own context — `getContext` in the app, the test harness's own factory in tests. It works and it is explicit, but it is plumbing that exists only because the context cannot answer the question itself, and every future sessionless-but-authorised caller pays it again.

## What would close it

Something like `context.withSession(session)` sitting beside `sudo()`, reusing the already-resolved config and this context's own client (including a transaction client, so a substituted-session write stays inside the caller's transaction). Every call site above collapses to one line at the point of use, and the factory type disappears.

Two properties we would want if you take this on:

- **Orthogonal to `sudo()`.** `context.withSession(s).sudo()` and `context.sudo().withSession(s)` should both be meaningful and equivalent — substitution changes who the hooks see, elevation changes whether access rules run.
- **It must not widen anything by itself.** Substituting a session is not an authorisation; whatever the new session can do is exactly what any context with that session could do. The application decides who may call it.

Happy to send a PR if the shape above is roughly what you would want.

borisno2:
## Summary

- Adds `context.withSession(session)` to `StackContext`, a sibling to `sudo()` on the other axis: `sudo()` keeps the session and bypasses access control; `withSession()` keeps access control (and hooks) running but substitutes the session.
- Reuses the receiver's already-resolved config, client (including a transaction client — a call inside `context.transaction()` stays in that transaction), and storage — no second connection is opened.
- Orthogonal to `sudo()`: `context.withSession(s).sudo()` and `context.sudo().withSession(s)` are equivalent, since `withSession` preserves the receiver's sudo state rather than resetting it.
- Grants no authority of its own — the derived context can do exactly what any context built directly with that session could do; access rules still evaluate against the new session.
- `withSession(null)` is a legitimate way to drop to an anonymous context.
- Documents the new method in `packages/core/CLAUDE.md` and `docs/content/reference/context-api.md`, alongside `sudo()`.

This closes the gap described in #980: an unattended dispatcher, service principal, or job runner that is legitimately authorised but arrives without the session a list `validate` hook expects previously had no supported way to present one — `sudo()` erases the check rather than answering it, and rebuilding a context from scratch requires config/client that aren't reachable from outside.

## Test plan

- [x] New test suite `packages/core/tests/with-session.test.ts` covering: session substitution without mutating the original, hooks/access control seeing the substituted session, parity with a context built directly from that session, `withSession(null)` yielding an anonymous context, `withSession()`/`sudo()` orthogonality (both directions), sudo-state preservation, client/storage reuse, and same-transaction participation (including rollback) inside `context.transaction()`.
- [x] `pnpm build` (core package typechecks)
- [x] `pnpm test` (full core suite, 1111 tests passing)
- [x] `pnpm lint`
- [x] `pnpm manypkg fix`
- [x] `pnpm format`

Closes #980

---
_Generated by [Claude Code](https://claude.ai/code/session_01FuvZVoPwu4apbvhXStZYYM)_

### DIFF
--- a/.changeset/witty-goats-listen.md
+++ b/.changeset/witty-goats-listen.md
@@ -1,0 +1,19 @@
+---
+'@opensaas/stack-core': minor
+---
+
+Add `context.withSession(session)` — a sibling to `sudo()` for the other axis. It derives a `StackContext` that reuses the receiver's already-resolved config, client (including a transaction client — a call inside `context.transaction()` stays in that transaction), and storage, but carries a substituted session, so access control and hooks run against the new session as normal.
+
+This closes a gap for callers that are legitimately authorised but arrive without the session a list `validate` hook expects — an unattended dispatcher, a service principal, or a job runner:
+
+```typescript
+// Runs with the job owner's session so hooks see the right identity, while
+// still going through the normal access control checks for that session.
+const asOwner = context.withSession(job.ownerSession)
+await asOwner.db.task.update({ where: { id: job.taskId }, data: { status: 'done' } })
+
+// Drop to anonymous
+const anonymous = context.withSession(null)
+```
+
+`withSession` grants no authority of its own — the derived context can do exactly what any context built with that session directly could do. It's orthogonal to `sudo()`: `context.withSession(s).sudo()` and `context.sudo().withSession(s)` are equivalent, since `withSession` preserves the receiver's sudo state instead of resetting it.
--- a/.changeset/witty-goats-listen.md
+++ b/.changeset/witty-goats-listen.md
@@ -1,5 +1,6 @@
 ---
 '@opensaas/stack-core': minor
+'@opensaas/stack-cli': minor
 ---
 
 Add `context.withSession(session)` — a sibling to `sudo()` for the other axis. It derives a `StackContext` that reuses the receiver's already-resolved config, client (including a transaction client — a call inside `context.transaction()` stays in that transaction), and storage, but carries a substituted session, so access control and hooks run against the new session as normal.
@@ -17,3 +18,5 @@ const anonymous = context.withSession(null)
 ```
 
 `withSession` grants no authority of its own — the derived context can do exactly what any context built with that session directly could do. It's orthogonal to `sudo()`: `context.withSession(s).sudo()` and `context.sudo().withSession(s)` are equivalent, since `withSession` preserves the receiver's sudo state instead of resetting it.
+
+The generated `Context<TSession>` type (`.opensaas/types.ts`) now includes `withSession: (session: TSession | null) => Context<TSession>` alongside `sudo`, so the method is typed in application code — run `opensaas generate` (or `pnpm generate`) to pick it up.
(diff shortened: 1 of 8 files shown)


borisno2:
## Self-review

Ran `/code-review` against this PR's diff before considering it done. Three findings came back, all addressed in the follow-up commit:

1. **Generated `Context<TSession>` type didn't expose `withSession`** — `packages/cli/src/generator/types.ts` only emitted `sudo()`, so application code calling `context.withSession(...)` against the generated `.opensaas/context.ts` factory would fail to typecheck even though it worked at the core-package level. Fixed by adding `withSession: (session: TSession | null) => Context<TSession>` to the generated interface, with snapshot updates in both the generator and integration test suites.
2. **Docs typed `withSession` as `(session: Session) => Context`**, omitting `| null`, contradicting the very next example in the same doc that calls `context.withSession(null)`. Fixed in `docs/content/reference/context-api.md`.
3. **Plugin runtime services are rebuilt (not reused) on every `withSession()` call** — this matches `sudo()`'s existing behavior and is actually necessary for correctness (a plugin service that closes over the session must bind to the *new* one, not stay stale), but it wasn't documented for `withSession` specifically. Added a caveat to `packages/core/CLAUDE.md` explaining this and noting the perf implication for a hot loop.

All fixes are in commit 6dfa7ff. Full core (`1111`) and CLI (`377`) test suites pass, along with `pnpm lint`, `pnpm manypkg fix`, and `pnpm format`.

---

borisno2:
The `Vercel` status check is failing due to a free-tier deployment rate limit ("Resource is limited - try again in 24 hours (more than 100, code: `api-deployments-free-per-day`)"), unrelated to this PR's diff — no code change here can fix it. The `test` and `e2e` GitHub Actions checks are still running; I'll keep watching those.

---

PR #1004, opened 2026-08-22
fix(core): run target list's hooks for nested writes through synthetic reverse relations

### LINKED ISSUE #978: List-level validate hooks do not run for a nested create through a parent list
Raised from `borisno2/enrolm8`, where the two repositories' shared guidance says to report a stack gap here rather than work around it silently.

## What we observed

A list's own `validate` hook does not appear to run when a row of that list is created **nested inside a write to another list**. Creating the same row directly does run it.

Concretely, with a `ChargeRequest` list carrying a `validate` hook that enforces ten separate rules on create:

```ts
// Direct create — the hook runs, and this is refused.
await context.sudo().db.chargeRequest.create({ data: { /* incoherent row */ } })

// Nested create through the parent — the hook is silent, and this COMMITS.
await context.sudo().db.account.update({
  data: { from_ChargeRequest_account: { create: { /* the same incoherent row */ } } },
})
```

The nested path committed a row that violated every rule the hook checks — a `WITHDRAW` carrying an Account and a full Charge (mutually exclusive by that list's rules), a contract version the schema does not admit, an actor kind that is explicitly not a valid actor, and a back-dated timestamp. All ten assertions were silent. The row was caught later by our own read-time parser, not by the hook.

This was found by an independent review probing for exactly this class of gap, and it was reproduced deliberately — it is not a report from a production incident.

## Why it matters

Access control and hooks are the stack's headline guarantee: operations go through the context wrapper *so that* they cannot bypass the checks a list declares about itself. A list author reading the hooks documentation would reasonably conclude that a `validate` hook is a property of the list — that a row of that type cannot come into existence without satisfying it, whatever route it arrives by.

If nested creates skip it, then any list reachable as a relationship target from another list has a second, unchecked door. The author of the guarded list cannot see that door from their own file, and would have no reason to look for it.

In our case the practical exposure is nil: no production code names that nested key, and a non-`sudo` attempt is refused earlier ("not a field of this list"). We are reporting the behaviour, not an incident.

## What we are unsure of

We have not read the stack's nested-write implementation closely enough to say whether this is:

1. a bug — hooks are intended to run and do not on this path;
2. intended, with the reasoning being that the parent's hooks govern the whole write; or
3. intended but under-documented, in which case the docs are the fix.

If it is (2) or (3), that is a legitimate answer and worth stating explicitly in the hooks documentation, because the natural reading of "list-level `validate`" is that it is unconditional. Our own codebase carries a written claim that `sudo()` "bypasses access control but deliberately still runs hooks" — true for a direct create, and we are correcting that sentence on our side to name the nested exception.

## Environment

- `@opensaas/stack-core` as consumed by `borisno2/enrolm8` at the time of writing
- Postgres via Prisma driver adapters; reproduced in-process on PGlite
- Both the guarded list and the parent are ordinary lists with a standard relationship; the nested key was the generated `from_<List>_<field>` synthetic relation field

Happy to supply a minimal reproduction against one of the examples if that would help — say which example you would prefer it in.

borisno2:
## Summary

- A nested `create`/`update`/`delete` addressed to a list-only ref's **synthetic** reverse relation (`from_<List>_<field>`, the back-relation the generator synthesizes because Prisma requires an opposite field the config never declares) was unrecognised by `processNestedOperations` and handed to Prisma completely untransformed — skipping `resolveInput`, `validate`, built-in field validation, and `beforeOperation`/`afterOperation`. Reachable only under `sudo()`, since the undeclared-key guard already refused it otherwise.
- Added `resolveSyntheticReverseRelation` (`packages/core/src/access/engine.ts`), which resolves such a key back to the declared relationship field that owns it, reusing the exact `from_<List>_<field>` construction the generator emits (`getSyntheticFieldName`, exported from `fields/index.ts`) so the two names can't drift.
- `processNestedOperations` now routes a resolved synthetic key through the identical nested-write pipeline a declared relationship field gets — same hooks, same field-level access, same created-row recovery via `include`.
- Also tightened `filterWritableFields` under `sudo()`: an undeclared key that does **not** resolve to a synthetic reverse relation is now refused, closing the same silent-bypass shape for any undeclared key, not just relations. This restores the documented `sudo()` contract ("bypasses access control but still runs hooks") on this path.
- Nested writes through **declared** relationship fields are unchanged (pinned by the existing suite, which still passes in full).

## Test plan

- [x] New unit tests for `resolveSyntheticReverseRelation` (`packages/core/src/access/engine.test.ts`)
- [x] New end-to-end nested-write tests (`packages/core/tests/nested-write-synthetic-relation.test.ts`): sudo nested create runs `validate` and refuses an invalid row; `resolveInput`/`beforeOperation`/`afterOperation` all fire; created-row recovery works for a to-many synthetic relation; nested update/delete fire with correct `originalItem`/`item`; a throwing hook rolls back the whole write; a non-sudo write via the synthetic key is still refused (unchanged); a genuinely unknown key is refused even under sudo
- [x] New unit tests pinning the tightened `filterWritableFields` sudo guard (`packages/core/src/access/field-access.test.ts`)
- [x] Verified all new tests fail without the fix (reverted the fix locally, confirmed 12 failures, restored it)
- [x] Full core test suite passes (1116 tests)
- [x] `pnpm lint` passes
- [x] `pnpm format` / `pnpm manypkg fix` applied

Closes #978

---
_Generated by [Claude Code](https://claude.ai/code/session_01QaoWmNty8ZUwKkWSdHA25Z)_

### DIFF
--- a/.changeset/tame-otters-comply.md
+++ b/.changeset/tame-otters-comply.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+Fix a nested create/update/delete through a list-only ref's synthetic reverse relation (`from_<List>_<field>`) silently bypassing the target list's hooks and validation. It now runs the same pipeline a declared relationship field's nested write gets. Under `sudo()`, an undeclared key that isn't a synthetic reverse relation is now refused rather than passed through unchecked.
--- a/packages/core/src/access/engine.test.ts
+++ b/packages/core/src/access/engine.test.ts
@@ -1,0 +1,67 @@
+import { describe, it, expect } from 'vitest'
+import { resolveSyntheticReverseRelation } from './engine.js'
+import type { OpenSaasConfig, FieldConfig } from '../config/types.js'
+
+// A relationship field pointing at another list.
+function rel(ref: string, many = false): FieldConfig {
+  return { type: 'relationship', ref, many } as unknown as FieldConfig
+}
+
+/**
+ * #978: a list-only `ref` (`ref: 'ListName'`, no target field) generates a
+ * synthetic back-relation on its target — `from_<SourceList>_<sourceField>`
+ * (see `getSyntheticFieldName` in fields/index.ts). `resolveSyntheticReverseRelation`
+ * is the runtime counterpart that recognises the same name and resolves it
+ * back to the declared field that owns it, so a nested write addressed to it
+ * gets the same hook pipeline a declared relationship field gets.
+ */
+describe('resolveSyntheticReverseRelation', () => {
+  const config: OpenSaasConfig = {
+    db: { provider: 'sqlite', url: 'file:./dev.db' },
+    lists: {
+      Account: { fields: { name: { type: 'text' } as unknown as FieldConfig } },
+      ChargeRequest: {
+        fields: {
+          kind: { type: 'text' } as unknown as FieldConfig,
+          // List-only ref — no field named on Account, so a back-relation is
+          // synthesized on it as `from_ChargeRequest_account`.
+          account: rel('Account'),
+        },
+      },
+      // A bidirectional relationship — Post.author names User.posts directly,
+      // so no back-relation is synthesized anywhere for it.
+      User: { fields: { posts: rel('Post.author', true) } },
+      Post: { fields: { author: rel('User.posts') } },
+    },
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  } as any
+
+  it('resolves the synthetic back-relation name to its owning field', () => {
+    const result = resolveSyntheticReverseRelation('from_ChargeRequest_account', 'Account', config)
+    expect(result).not.toBeNull()
+    expect(result?.sourceListName).toBe('ChargeRequest')
+    expect(result?.sourceFieldName).toBe('account')
+    expect(result?.sourceFieldConfig.ref).toBe('Account')
+  })
+
+  it('returns null for a name that does not match the synthetic construction', () => {
+    expect(
+      resolveSyntheticReverseRelation('from_ChargeRequest_amount', 'Account', config),
+    ).toBeNull()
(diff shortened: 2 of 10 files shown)


borisno2:
## Self-review

Ran a high-effort self-review of this diff before handing it off.

**Finding**: `resolveSyntheticReverseRelation` was being called twice per synthetic key — once from `filterWritableFields` (to validate under `sudo()`) and again from `processNestedOperations` (to dispatch the nested write) — a redundant `lists × fields` config scan.

**Attempted fix**: moved the resolution into `processNestedOperations` only, and reverted `filterWritableFields`'s sudo branch to unconditional passthrough. Running the full test suite caught a real regression this introduced: raw per-part columns from a multi-column field (`m_url`/`m_size`, #789) are *also* undeclared from `processNestedOperations`'s field-config perspective, and reach it on every write — sudo or not. `filterWritableFields` already disambiguates those from a genuinely-unrecognised key via its own `splitColumnOwners` map, before it ever reaches its undeclared-key branch; `processNestedOperations` has no such map, so treating "unresolved" as "throw" there wrongly rejected legitimate multi-column writes (caught by the existing `multi-column-validation-ordering.test.ts` suite).

**Resolution**: reverted to resolving in both places. The double scan is a real, minor cost (bounded by schema size, on the rare sudo + undeclared-key path only) — correctness won out over the micro-optimization. Left an explanatory comment in `processNestedOperations` recording why this can't move, so a future pass doesn't rediscover the same regression the hard way.

Full suite (1116 tests), `pnpm build`, `pnpm lint`, and `pnpm format` all pass on the current head.


---

borisno2:
The "Vercel" status is red for an unrelated reason: `stack-docs` preview deployment is rate-limited on Vercel's free tier ("Resource is limited - try again in 24 hours"), not a build/code failure — nothing in this diff touches the docs site. Not actionable from this PR; GitGuardian passed, and the real `test`/`e2e` GitHub Actions checks are still running on the latest commit. Will follow up once those land.


---

Issue #994, opened 2026-08-22
auth: better-auth plugin tables get no foreign keys, so deleting a user orphans OAuth tokens and consents


borisno2:
> *This was generated by AI during triage.* Split out of #992 so the data-integrity fix can land without waiting on the converter consolidation.

## Agent Brief

**Category:** bug
**Summary:** Reference fields on better-auth plugin tables are emitted as bare nullable columns with no relation, so there is no foreign-key constraint and no cascade — and user deletion leaves orphaned OAuth access tokens and consents.

**Current behavior:**

The auth package's better-auth schema converter deliberately drops reference fields to plain columns. Its own comment explains why:

> Better Auth uses one-way references; OpenSaaS requires bidirectional relationships, so reference fields are kept as plain text (foreign-key) columns here. An app can add a bidirectional relationship manually by extending the corresponding list.

**That premise is false.** OpenSaaS supports list-only refs (`ref: 'ListName'`, naming no field on the target), which generate a real foreign key plus a synthetic reverse collection. Verified:

```prisma
model User {
  from_OauthAccessToken_user OauthAccessToken[] @relation("OauthAccessToken_user")
}
model OauthAccessToken {
  userId String @map("user")
  user   User   @relation("OauthAccessToken_user", fields: [userId], references: [id])
  @@index([userId])
}
```

Because references are dropped instead, the generated models look like this:

```prisma
model oauthAccessToken {
  clientId String?
  userId   String?
}
```

against better-auth declaring both as required, cascading references (`clientId → oauthApplication.clientId`, `userId → user.id`, both `onDelete: cascade`). Three defects in one:

- **no `@relation` and no foreign-key constraint** — no referential integrity
- **no `onDelete: Cascade`**
- **nullable columns despite `required: true`** — the converter sets `validation: { isRequired: … }`, an application-layer check that does not affect the Prisma column (the footgun `CLAUDE.md` warns about; same class as #863)

**The consequence is confirmed, not theoretical.** better-auth's internal `deleteUser` removes exactly two things before deleting the user:

```js
deleteUser: async (userId) => {
  if (…) await deleteMany([{ field: "userId", value: userId }], "session")
  await deleteMany([{ field: "userId", value: userId }], "account")
  await delete([{ field: "id", value: userId }], "user")
}
```

`session` and `account` only. Plugin tables are not enumerated — better-auth relies on the database cascade its own schema declares. Since the stack emits no constraint, nothing cleans them up:

[...]

PR #1005, opened 2026-08-22
fix(auth): consolidate better-auth plugin-table derivation into deriveAuthLists

### LINKED ISSUE #992: auth: consolidate the two better-auth→list converters, and give plugin tables real foreign keys
> *This was generated by AI during triage.* Referenced as follow-up in #987 and #991; filing it properly.

## Summary

There are two independent code paths turning better-auth schema into stack lists, each implementing roughly half the job, and they disagree with each other:

| | `deriveAuthLists` | `convertBetterAuthSchema` |
| --- | --- | --- |
| Covers | the 4 core models + rate-limit | better-auth plugin tables |
| How | hand-written transcription | reads better-auth metadata |
| `references` → relationship | ✅ hand-written | ❌ **dropped** — kept as a bare column |
| Column maps (`fieldName`) | ✅ | ❌ not read |
| `index: true` | ❌ (#937) | ❌ not read |
| `unique` | ✅ | ✅ |
| `defaultValue` | ✅ | ✅ |
| `bigint: true` | ✅ | ✅ (#917) |
| List-key casing | PascalCase | upstream `modelName`, i.e. camelCase (#991) |

Neither is a superset of the other, so bugs get fixed twice or once. #917 (`bigint: true` → 32-bit `Int`) was fixed in the plugin converter; the same concern in the core lists was handled separately by hand.

## The concrete cost: plugin tables have no foreign keys

`convertTableToList` deliberately drops reference fields to plain columns:

> Better Auth uses one-way references; OpenSaaS requires bidirectional relationships, so reference fields are kept as plain text (foreign-key) columns here. An app can add a bidirectional relationship manually by extending the corresponding list.

Generated output for an MCP-enabled config:

```prisma
model oauthAccessToken {
  id        String   @id @default(cuid())
  accessToken  String? @unique
  refreshToken String? @unique
  …
  clientId     String?
  userId       String?
}
```

Against better-auth's declarations for that table:

```
clientId -> { model: "oauthApplication", field: "clientId", onDelete: "cascade" }
userId   -> { model: "user",             field: "id",       onDelete: "cascade" }
```

So the generated model has:

- **no `@relation` and no foreign-key constraint** — no referential integrity at the database level
- **no `onDelete: Cascade`** — deleting a user leaves orphaned access tokens and consents, where better-auth's own schema cascades them
- **nullable columns despite `required: true`** — the converter sets `validation: { isRequired: … }`, which is an application-layer check and does not make the Prisma column non-null (the footgun `CLAUDE.md` warns about, and the same class as #863)

All three of the MCP plugin's tables (`oauthApplication`, `oauthAccessToken`, `oauthConsent`) declare references and are affected.

## The stated reason for the limitation no longer holds

The comment justifies dropping references because "Better Auth uses one-way references; OpenSaaS requires bidirectional relationships." That was a real obstacle — but the [#987 FK spike](https://github.com/OpenSaasAU/stack/issues/987#issuecomment-5377507366) demonstrated the translation is mechanical: emit a `relationship()` on the child from `references`, graft a `many: true` reverse onto the parent, and synthesize the reverse field name by pluralizing the child model. That prototype reproduced the hand-written core-model FK shape exactly, cascade and all.

The one non-derivable piece — the reverse field's name — is the same problem in both converters and wants one answer, not two.

## Suggested direction

One converter, reading better-auth metadata, used for core models and plugin tables alike:

- FK translation from #987's spike, applied to both — which fixes the plugin tables' missing constraints and cascades as a side effect
- `fieldName` column maps, `index: true`, `unique`, `defaultValue`, `bigint` read once, in one place
- `required` → `db.isNullable: false`, uniformly
- list keys PascalCased independently of table names, fixing #991 for plugin tables and preventing rate-limit from regressing into it
- the reverse-naming rule defined once

**Sequencing:** #987 builds the metadata-driven derivation for the core models and already encodes most of these rules. The cheapest path is probably to land #987 first, then extend its converter to absorb the plugin-table path, rather than doing a separate consolidation pass. Worth confirming during triage — if so, this issue is #987's direct successor and should not start before it.

**Possible split:** the nullability defect on plugin-table reference columns (`required: true` producing a nullable column) is a plain bug and is independently fixable without any consolidation. A triager may want it as its own issue if consolidation is deferred.

## Related

- #987 — metadata-driven derivation for the core models; the FK spike that removes this issue's stated blocker
- #991 — list-key casing, one of the disagreements between the two converters
- #988 — the drift test; should cover whichever converters survive
- #917 — `bigint: true` fixed in the plugin converter only, an example of the duplicated-fix cost
- #863 — the same nullability class of defect, previously fixed in the core lists
- ADR-0013 — plugin-injected lists ship closed, which bounds the exposure of these tables today

borisno2:
## Summary

- Consolidates the two better-auth → stack-list converters (`deriveAuthLists` for the four base Auth models, and the now-deleted `convertBetterAuthSchema`/`convertTableToList` for better-auth plugin tables) into one derivation.
- Better-auth plugin tables (e.g. the MCP plugin's `oauthApplication`/`oauthAccessToken`/`oauthConsent`) now go through the same model registry and field-derivation pass as the base models: a reference to a target's `id` becomes a real `relationship()` with the correct `onDelete` cascade, index, uniqueness and nullability. This closes a confirmed data-integrity defect — deleting a user previously orphaned their OAuth rows, since neither the database (no FK constraint existed) nor better-auth's own `deleteUser` (which only cleans up `session`/`account`) removed them.
- A reference whose target field isn't the target's `id` (better-auth's own oidc-provider schema does this for `oauthAccessToken.clientId → oauthApplication.clientId`) is deliberately left as a plain, indexed scalar column, since `relationship()` only supports `id`-based foreign keys.
- Plugin-table scalar fields now honour `fieldName` column maps and `index: true`. List keys are PascalCased from better-auth's resolved `modelName`, with `db.map` restoring the physical table name when the case changed.
- `authPlugin()`'s `init` is simplified to a single pass over one combined `lists` record instead of two separate add-vs-extend loops — reverse relations onto a base model now land correctly regardless of processing order, since the whole registry is built before any list config is assembled.
- Adds `docs/adr/0034-...md` recording the decision and updates ADR-0033's "out of scope" note to point at it.

## Test plan

- [x] `packages/auth/tests/plugin-table-derivation.test.ts` (new) — unit coverage for the consolidated derivation: PascalCase keys + `db.map`, real FK for an `id`-targeting reference, non-PK reference stays scalar, referential actions beyond cascade, reverse relations with no collisions, `fieldName`/`index: true` on plugin scalars, closed access, unresolvable-reference error, base-model schema extension, and remapped base-model keys.
- [x] `packages/auth/tests/auth-lists-drift.test.ts` (extended) — field-by-field comparison of the derived MCP OAuth tables against `getAuthTables`' own output, catching drift the same way the existing base-model drift test does. (Fixed a latent gap in the shared comparison helpers: they assumed `fieldName` is always set, true for the four base models but not for plugin fields that don't declare one.)
- [x] `packages/auth/tests/generated-fk-shape.test.ts` (extended) — generated-schema-text assertions: cascade/`@map`/`@@index` on the OAuth tables' `user` relation, `@@map` to the physical table name, the non-PK `clientId` column staying a plain indexed column, and that Session/Account/Verification's generated blocks are byte-identical with and without the MCP plugin enabled (only `User` gains reverse collections).
- [x] `packages/auth/tests/mcp-oauth-cascade-e2e.test.ts` (new, opt-in via `RUN_MCP_OAUTH_CASCADE_E2E=1`, same pattern as the existing rate-limiter e2e guard) — live proof against a real generated schema + SQLite database: signs up two users, attaches OAuth rows to each via the raw Prisma client, deletes one user through better-auth's own `/delete-user` endpoint, and asserts the database cascade removed exactly that user's rows while the other user's rows survive. Ran locally with the env flag set — passes.
- [x] `pnpm --filter @opensaas/stack-auth test` — 268 passed, 3 skipped (the env-gated e2e guards)
- [x] `pnpm --filter @opensaas/stack-core test` / `pnpm --filter @opensaas/stack-cli test` — unaffected, all green
- [x] `pnpm lint`, `pnpm exec manypkg check`, `pnpm format` — clean
- [x] Changeset added (`@opensaas/stack-auth`: minor)

Closes #992

Supersedes #994 (see ADR-0034's Consequences section for why).

---
_Generated by [Claude Code](https://claude.ai/code/session_01BEt3NnYnzQVCVUJcWX5az4)_


borisno2:
## Self-review (code-review skill, medium effort)

Ran a review pass against this branch before handing it off. Two findings, both confirmed by reproducing against the built package and fixed in 47d2448:

1. **Correctness — unsupported field type crashed instead of degrading.** `buildScalarField`'s switch now runs against arbitrary better-auth plugin schema fields, which can legitimately be `json`/array/enum types it doesn't model. It threw, which would crash `config()` resolution for any app whose plugin declares such a field — a regression from the deleted `convertField`'s warn-and-fall-back-to-`text()` behavior. Restored the fallback (with the same console warning) instead of throwing.
2. **Correctness — silent reverse-relation collision.** `reverseRelationName` is keyed by model, not field, so two `id`-referencing FKs on the same plugin table targeting the same model would silently collapse onto one reverse relation name and produce an ambiguous/incorrect generated schema with no error. This is a pre-existing limitation of the naming rule (out of scope to change per the issue brief), not something this PR introduces, but consolidation newly exposes it to arbitrary plugin schemas. Rather than changing the naming rule, added a check that fails loudly with a clear error pointing at the existing `REVERSE_RELATION_NAME_OVERRIDES` escape hatch, instead of silently corrupting the schema.

Neither case is triggered by any built-in better-auth plugin (MCP, admin, organization, two-factor, siwe) checked. Added regression tests for both in `plugin-table-derivation.test.ts`. Full `@opensaas/stack-auth` suite (270 passed, 3 skipped e2e guards), `@opensaas/stack-core`, and `@opensaas/stack-cli` all green; lint/manypkg/format clean.

---

borisno2:
The `Vercel` status failure is the docs-site preview deployment hitting Vercel's free-tier daily deployment cap (`api-deployments-free-per-day`, retry in 24h) — an account-wide quota, unrelated to this diff (which touches `@opensaas/stack-auth` and its tests, not the docs site). Not treating it as a blocker. The real CI (`test`/`e2e` jobs) is still running on the latest commit — will follow up once it completes.

---

PR #1006, opened 2026-08-22
core: allow db.foreignKey.map on list-only relationship refs

### LINKED ISSUE #993: core: `db.foreignKey` is rejected on list-only relationships, though the generator already supports it
> *This was generated by AI during triage.* Found while checking a claim in #992 about whether one-sided relationships are usable.

## Summary

`relationship({ ref: 'ListName' })` — a list-only ref — throws if given `db.foreignKey`:

```
db.foreignKey can only be used on bidirectional relationships (ref: "ListName.fieldName").
List-only refs (ref: "ListName") always create foreign keys automatically.
```

The reasoning in the message is sound as far as it goes: a list-only ref *does* always create the foreign key, so `foreignKey: true` (the "which side owns it" sense) is meaningless. But `db.foreignKey` also carries `map` — the **column name** — and that is not meaningless on a list-only ref. The validation rejects both senses because they share one option.

## The generator already handles it

The FK emission path does not branch on bidirectional vs list-only. It reads the map if present and falls back to the field name otherwise:

```ts
const mapModifier =
  typeof field.db?.foreignKey === 'object' && field.db.foreignKey.map
    ? ` @map("${field.db.foreignKey.map}")`
    : ` @map("${fieldName}")`
```

So a list-only ref carrying `foreignKey: { map: 'userId' }` would generate correctly today. The only thing preventing it is the validation throw. This is an arbitrary restriction, not a structural one.

## Why it matters

Without it, a list-only ref's FK column is always named after the **relationship field**. Verified:

```prisma
model OauthAccessToken {
  userId String @map("user")           // ← column is "user", not "userId"
  user   User   @relation("OauthAccessToken_user", fields: [userId], references: [id])
  @@index([userId])
}
```

That is the same Keystone-parity default behind #935, but on the list-only path there is **no way to override it**. Consequences:

- **Adoption is impossible on this path.** Any list-only relationship pointing at a table whose live FK column doesn't match the relationship field name will diff dirty forever, with no escape hatch short of switching to a bidirectional ref purely to gain the naming knob.
- **It distorts the fix for #992.** The plugin-table converter could emit list-only refs to gain foreign keys and cascades — which is otherwise exactly what is needed — but would be locked into a `@map("user")` column that no better-auth database has. Without this restriction lifted, that fix has to either accept a wrong column name or synthesise bidirectional refs (and therefore reverse field names) for no reason other than the naming knob.
- **The workaround costs schema surface.** Switching to a bidirectional ref just to name a column adds a reverse relation field to the target list that nobody asked for.

## Suggested direction

Narrow the validation rather than removing it. On a list-only ref:

- `foreignKey: true` / `false` (the ownership sense) stays rejected — it genuinely has no meaning there, and the current error text explains why well.
- `foreignKey: { map: '…' }` is accepted and used.

The `many: true` rejection above it is unaffected and correct — the many side never stores the key.

Worth confirming during triage whether any other `db.foreignKey` member exists or is planned that would be ownership-flavoured, since that decides whether narrowing by shape (`object` vs `boolean`) is sufficient or whether the option wants splitting into two.

## Reproduction

```ts
relationship({
  ref: 'User',                                  // list-only
  db: { foreignKey: { map: 'userId' } },        // throws
})
```

Removing the `db.foreignKey` generates `userId String @map("user")` — a working relation with the wrong column name.

## Related

- #992 — the plugin-table FK gap, where this restriction shapes the available fix
- #935 — the same column-naming default on the bidirectional path, where an override does exist
- #987 — derivation; if this is lifted, list-only refs become viable for derived tables and the reverse-relation naming problem shrinks

borisno2:
## Summary
- `relationship({ ref: 'ListName' })` — a list-only ref — previously threw if given any `db.foreignKey`, even though the FK-emission path already handled the `{ map }` form correctly for that path.
- Narrowed the validation in `relationship()` (`packages/core/src/fields/index.ts`) so a list-only ref only rejects the *boolean* form of `db.foreignKey` (`true`/`false`, the "which side owns the FK" sense, which is meaningless when the ref always owns the FK). The object form (`{ map: '...' }`, the column-naming sense) is now accepted and used to rename the generated foreign key column.
- The `many: true` rejection and the bidirectional-ref behavior are unchanged.
- Updated the `db.foreignKey` TSDoc in `packages/core/src/config/types.ts` to document the list-only case.

## Test plan
- [x] Added `packages/core/src/fields/relationship-foreign-key.test.ts` covering: boolean rejected on list-only refs, object accepted on list-only refs, boolean still accepted on bidirectional refs, and `many: true` still rejected regardless of ref shape.
- [x] Added a CLI generator test in `packages/cli/src/generator/prisma.test.ts` asserting `db.foreignKey: { map: 'category_id' }` on a list-only ref generates `categoryId String? @map("category_id")`.
- [x] `pnpm test` (core: 1136 passed, cli: 378 passed)

Closes #993

---
_Generated by [Claude Code](https://claude.ai/code/session_01S24jAnCVafU9hk6ZqJvfhp)_

### DIFF
--- a/.changeset/tame-jokes-relax.md
+++ b/.changeset/tame-jokes-relax.md
@@ -1,0 +1,5 @@
+---
+'@opensaas/stack-core': patch
+---
+
+`relationship({ ref: 'ListName' })` list-only refs now accept `db.foreignKey: { map: '...' }` to rename the foreign key column. The boolean form (`true`/`false`) is still rejected there since ownership is implicit on a list-only ref.
--- a/packages/cli/src/generator/prisma.test.ts
+++ b/packages/cli/src/generator/prisma.test.ts
@@ -773,6 +773,34 @@ describe('Prisma Schema Generator', () => {
       expect(schema).toContain('categoryId   String? @map("category")')
     })
 
+    it('should honour db.foreignKey.map to rename a list-only relationship foreign key column', () => {
+      const config: OpenSaasConfig = {
+        db: {
+          provider: 'sqlite',
+        },
+        lists: {
+          Category: {
+            fields: {
+              name: text(),
+            },
+          },
+          Post: {
+            fields: {
+              title: text(),
+              category: relationship({
+                ref: 'Category',
+                db: { foreignKey: { map: 'category_id' } },
+              }),
+            },
+          },
+        },
+      }
+
+      const schema = generatePrismaSchema(config)
+
+      expect(schema).toContain('categoryId   String? @map("category_id")')
+    })
+
     it('should generate a required FK column and relation field with db.isNullable: false', () => {
       const config: OpenSaasConfig = {
         db: {
--- a/packages/core/src/config/types.ts
+++ b/packages/core/src/config/types.ts
@@ -1136,13 +1136,20 @@ export type RelationshipField<TTypeInfo extends TypeInfo = TypeInfo> =
        */
       isNullable?: boolean
       /**
-       * Controls foreign key placement and column name for bidirectional relationships
-       * Can be a boolean or an object with a map property
-       * Only valid on single (non-many) relationships
-       * Cannot be true on both sides of a one-to-one relationship
+       * Controls foreign key placement and column name.
+       * Can be a boolean or an object with a map property.
+       * Only valid on single (non-many) relationships.
+       * Cannot be true on both sides of a one-to-one relationship.
        *
-       * When a boolean, defaults the foreign key column name to the field name
-       * When an object with map, uses the provided column name
+       * The boolean form (the "which side owns the foreign key" sense) is only
+       * meaningful on a bidirectional ref (`ref: 'ListName.fieldName'`) — a
+       * list-only ref (`ref: 'ListName'`) always owns the foreign key, so a
+       * boolean here is rejected. The `{ map }` form (the column-name sense)
+       * works on both: it renames the foreign key column without changing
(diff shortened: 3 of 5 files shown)


borisno2:
## Automated review

Ran a full self-review of this diff (correctness, reuse/simplification, efficiency, altitude, and CLAUDE.md conventions). No findings.

**Checked:**
- The validation change only narrows a prior blanket rejection (boolean `db.foreignKey` still rejected on list-only refs; the `{ map }` form is now allowed) — a strict loosening, so no previously-working config can break.
- The FK-emission path (`mapModifier` in `getPrismaRelation`) already handled the `{ map }` case correctly for list-only refs; it was only ever gated by the validation fixed here. `shouldHaveForeignKey` unconditionally owns the FK for list-only refs, so the one-to-one ownership logic is never reached on this path — consistent with the new error message.
- Grepped all other `db.foreignKey` call sites (`packages/auth/src/config/derive-auth-lists.ts`, `packages/core/src/access/field-access.ts`) — both unrelated/unaffected.
- New tests cover all four relevant cases (boolean rejected on list-only, object allowed on list-only, boolean still allowed on bidirectional, `many: true` still rejected) plus the generated-schema output.
- Full suites green: core (1136 passed), cli (378 passed). `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all clean.

---

PR #1007, opened 2026-08-22
MCP query tool accepts a fields projection argument

### LINKED ISSUE #851: MCP query tools should let the assistant ask for relations

Follow-up from #848 / ADR-0024 (#849). Filed as agreed during that triage; not yet specified.

After #848, the derived MCP `query` tool issues a bare read and so returns scalars plus foreign keys. That is the right default for a tool result — a five-level relation tree lands byte-for-byte in a context window — and relations stay reachable, because an assistant can follow a foreign key with a second tool call.

If assistants turn out to round-trip often enough for that to be the wrong trade, the fix is to let them ask: an `include` argument on the tool's input schema.

The part that needs thought before it's built: a caller-supplied include that names a relation deeper than the Access Filter can scope throws `AccessScopeDepthExceededError`. Coming from an assistant, that is a thrown error where a tool result was expected, so it needs to surface as something the model can recover from — an error result naming what it asked for and why it was refused, not a 500. Worth deciding at the same time whether the include should be depth-limited in the schema itself, so the common case is unreachable rather than caught.

Not urgent. Worth revisiting once there's real usage data on how assistants navigate related data through the tools.

borisno2:
## Summary

Implements #851 per the design recorded in ADR-0033 (`docs/adr/0033-mcp-tools-advertise-a-bounded-projection.md`), unblocked now that #995 (isError tool results) has landed.

- The derived `list_{dbKey}_query` MCP tool accepts an optional `fields` projection — the wire form of the runtime's existing fragment field selection (`{ scalarField: true, relation: { fields: {...} } }`) — instead of the `include` originally proposed in #851, per the maintainer's triage decision.
- Omitting `fields` is unchanged: a bare read, exactly as before (ADR-0024).
- A to-one relation takes a nested `fields` only; a to-many additionally accepts `where`/`orderBy`/`take`/`skip` and a `count` (in place of, or alongside, its rows). A nested `take` defaults to 5 and is clamped to a hard cap of 50.
- The generated schema enumerates exactly **two levels** of each list's own vocabulary, per session — a relation whose target list denies this session's operation-level `query` access, or has `mcp.enabled: false`, is omitted from the vocabulary entirely (and terminates the schema against a self-referential relationship like `Category.parent`/`children`).
- A `fields` request naming anything the schema doesn't advertise (unknown field, relation nested a third level deep, wrong shape) is refused as an `isError` tool result naming what was asked for and what's available — never served on a best-effort basis, never a JSON-RPC protocol error.
- `fields` is translated into a plain `context.db` `include`, not the fragment `query:` argument, so the *existing* read pipeline does the work — `buildAccessScopedInclude`'s nested-`where` AND-fold, the to-one existence check, `needs` folding, and the depth cap all apply exactly as they do for any other caller `include`. No parallel read path (see `packages/core/CLAUDE.md`'s new "The MCP `query` Tool's `fields` Projection" section for the detail, including the one documented trade-off this choice makes).
- **Behaviour change:** `tools/list` is now evaluated per session. A list whose operation-level `query` access denies the session outright no longer appears in the listing at all — none of its four CRUD tools, and no relation entry elsewhere pointing at it.
- Small additive extension to `buildAccessScopedInclude`/`asEntryObject` (`access-filter.ts`) to carry a caller's nested `orderBy`/`skip` through on a to-many include, mirroring the existing `take` precedent — needed for the to-many pagination criteria above.
- `fieldToJsonSchema`/`generateFieldSchemas` moved from `mcp/handler.ts` to a new `mcp/field-schema.ts` so both the `create`/`update` schema generator and the new `fields` projection schema generator can share it without a circular import.

## Test plan

- [x] `pnpm test` — 1153 tests pass (`packages/core`), including 43 tests in the extended `mcp-handler.test.ts` (schema shape, per-session omission, request→`include` translation, count handling, refusals) and 4 tests in a new `mcp-fields-projection-access.test.ts` that exercise the *real* `getContext` pipeline (nested to-many `where` AND-combined with the related list's access filter, to-one nulling via the real existence check, field-level access still dropping values, a selected computed field folding its declared `needs`).
- [x] `pnpm build` — typechecks clean.
- [x] `pnpm lint` — no new warnings/errors.
- [x] `pnpm manypkg fix` / `pnpm format` — run, no changes beyond formatting.
- [x] Changeset added (minor, `@opensaas/stack-core`) calling out the `tools/list` behaviour change.
- [x] MCP how-to docs (`docs/content/how-to/mcp.md`) updated with a "Selecting Related Data with `fields`" section and a note on per-session tool listing.

Closes #851

---
_Generated by [Claude Code](https://claude.ai/code/session_01CthNCkBfwh5G3HGmG1AVLK)_

### DIFF
--- a/.changeset/violet-otters-shine.md
+++ b/.changeset/violet-otters-shine.md
@@ -1,0 +1,27 @@
+---
+'@opensaas/stack-core': minor
+---
+
+The derived MCP `query` tool now accepts an optional `fields` projection — the wire form of the runtime's existing fragment field selection — so an assistant can select scalars and nested relation fields (with `where`/`orderBy`/`take`/`skip`, and a to-many's row count) in a single call instead of following a foreign key with a second one. Omitting `fields` is unchanged, a bare read exactly as before.
+
+```json
+{
+  "name": "list_post_query",
+  "arguments": {
+    "fields": {
+      "title": true,
+      "author": { "fields": { "name": true } },
+      "comments": {
+        "fields": { "text": true },
+        "where": { "approved": { "equals": true } },
+        "take": 5,
+        "count": true
+      }
+    }
+  }
+}
+```
+
+The generated tool schema enumerates two levels of each list's own fields and relations, per session, and refuses (as an `isError` tool result, never a protocol error) anything it doesn't advertise — an unknown field, or a relation named a third level deep. See the ADR (`docs/adr/0033-mcp-tools-advertise-a-bounded-projection.md`) for the full design.
+
+**Behaviour change:** `tools/list` is now evaluated per session. A list whose operation-level `query` access denies the session outright no longer appears in the tool listing at all — none of its four CRUD tools, and no relation entry elsewhere pointing at it. Previously every list's tools were listed regardless of session.
--- a/docs/content/how-to/mcp.md
+++ b/docs/content/how-to/mcp.md
@@ -279,6 +279,7 @@ The MCP handler creates CRUD tools for each list. `{dbKey}` is the camelCase for
   orderBy?: Record<string, 'asc' | 'desc'>,
   take?: number,
   skip?: number,
+  fields?: Record<string, any>, // Projection — see "Selecting Related Data" below
 }
 ```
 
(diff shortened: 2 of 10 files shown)


borisno2:
## Self-review (`/code-review medium`)

Ran a code review over the diff before CI. Three findings came back, all fixed in ebd59c6:

1. **Correctness — `id` never selectable/returned.** `listConfig.fields` doesn't include the automatic `id`/`createdAt`/`updatedAt` system fields, so a `fields` projection could never name `id` and a narrowed result had no way to identify the record it came from — breaking any follow-up `update`/`delete` call. Fixed: `id` is now always projected regardless of selection (at every level, including nested relations), and `id`/`createdAt`/`updatedAt` are explicitly advertised and acceptable as selectable system fields.
2. **Simplification — `MCP_FIELDS_SCHEMA_DEPTH` unused.** The constant documented the projection schema's depth but nothing actually read it — the generator/validator are hand-written for exactly two levels. Changing the constant would have been a silent no-op. Fixed by rewriting the doc comment to say so honestly (the depth is fixed by design per ADR-0033/ADR-0026, not meant to be a live parameter — making it genuinely configurable would be scope creep against the ADR's own rejection of a depth knob).
3. **Correctness — missing type validation.** A relation selector's `where`/`orderBy`/`take`/`skip`/`count` reached Prisma unchecked, so a malformed value (e.g. `where: "not-an-object"`) surfaced as an opaque database error instead of the clear `McpProjectionRefusedError` every other malformed `fields` input gets. Fixed with explicit type checks mirroring `access-filter.ts`'s `asEntryObject`.

Added tests for all three (always-returned `id`, explicit system-field selection, and the new malformed-input refusals). Full suite: 1156/1156 passing, clean build/lint/format.

---

borisno2:
Thanks @chatgpt-codex-connector — all three findings were real gaps, fixed in c50f10d:

1. **P1, nested `where`/`orderBy` field-read bypass**: a nested predicate on a to-many relation reached Prisma unchecked, so a session could infer a field-level-denied related field's value from which rows came back, their order, or (combined with `count`) how many there were. Now validated against the related list's config the same way the root read path validates its own `where`/`orderBy` (reusing `validateQueryKeys`/`validateQueryFieldReadAccess`, #912/#915), before the include is ever built.
2. **P1, relationship-field-level `count` bypass**: `_count` isn't a declared field, so `filterReadableFields` never applied the relationship field's own field-level `read` access to it the way it already does for the relation's rows. Now checked explicitly per row before a count is emitted — a relation whose rows a denied field never reveals can't reveal its count either.
3. **P2, negative nested `take`**: `Math.min` alone doesn't cap a negative value, so `take: -100000` could bypass the advertised 50-row ceiling via Prisma's reverse-pagination semantics. Refused outright now (nested reverse pagination isn't a feature this surface supports).

Added targeted tests for all three (negative take refusal, denied-field-in-nested-predicate refusal, count suppressed when the relationship field itself denies read). Full suite still green (1160/1160), clean build/lint/format.

(Separately: the Vercel docs-preview deployment failed on a free-tier daily rate limit — `api-deployments-free-per-day` — unrelated to this PR's code and not something a push can fix; it should recover on its own.)

---

borisno2:
Good catch on the row-dependent issue with my previous fix — re-checking `checkFieldAccess` against the already-filtered `rawItem` could disagree with the canonical decision if the access rule depends on another field's raw value that field-visibility had already stripped.

Fixed in 6b4cb9c with the approach you described: a counted relation is now always named in the include (a count-only one with `take: 0`, fetching no rows), so the ordinary read pipeline evaluates the relationship field's own `read` access exactly once, against the true raw row, in the one place that already does this correctly (`filterReadableFields`). `projectMcpResult` now just reads that decision off whether the key survived into the filtered result — no redundant access evaluation, and it's synchronous again.

Added a wire-level test (`mcp-handler.test.ts`) confirming the `take: 0` include shape, plus an end-to-end test through the real `getContext` pipeline (`mcp-fields-projection-access.test.ts`) proving the count is actually suppressed when the relationship field denies read access. Full suite green (1162/1162).

Note: the applied-locally commit `7594810` mentioned in the previous comment never made it into this PR (no git remote in that environment) — 6b4cb9c is my own independent implementation of the same fix, not a cherry-pick of that one.

---

Issue #966, opened 2026-08-19
access: pre-account / anonymous flows have no access-scoped path, forcing sudo() + hand-rolled ownership checks


borisno2:
> Reported from a downstream app review (the app's own guidance invites stack-ergonomics feedback). No app-specific data included.

## The gap

Row-level access filters scope by an identity that resolves to an existing row — e.g. `{ account: { user: { id: { equals: session.userId } } } }`. That works once the owning row exists.

In **pre-account / anonymous** flows it doesn't. A signed-in-but-not-yet-onboarded user (e.g. an anonymous Better Auth session mid-signup) has a real `session.userId` but **no `Account` row yet**, so any filter that keys off `session.data.accountId` can't resolve — the thing it scopes to doesn't exist. The list's normal access filter can't express "the rows this half-formed session may touch."

## The consequence

The practical workaround is to drop to `sudo()` (access-bypassing context) for these flows and re-implement ownership by hand, e.g.:

```ts
const ctx = getSudoContext(session)
const student = await ctx.db.student.findUnique({ where: { id: studentId } })
if (student?.account?.id !== derivedAccountId) throw forbidden() // manual, not the access layer
```

This inverts the stack's usual safety posture. Filter-based access **fails closed** — a session that can't resolve its scope simply sees/writes nothing. A `sudo()` + manual-check path **fails open**: it's only safe as long as every hand-rolled ownership check is present and correct, and a future edit that drops one silently becomes an IDOR (the manual check is the *only* thing standing between a client-supplied id and a privileged write). In the reviewed app several such tool actions run entirely under sudo with a single manual `x.account.id === accountId` guard each — correct today, but fragile by construction, and the root cause is that there's no access-scoped option for these sessions.

## What would help

Some first-class way to keep pre-account / anonymous flows on the access-scoped context instead of sudo, e.g.:

- an access-filter mechanism that can key off the session directly (userId) with an explicit **deny-when-derivation-absent** semantics, so a half-formed session fails closed rather than needing sudo, or
- a documented, safe pattern/helper for "this session owns rows via derivation X; deny if X can't be resolved" that doesn't require bypassing access control.

The goal is that pre-signup flows keep the same fail-closed guarantee as the rest of the access layer, rather than each app re-deriving ownership by hand under sudo.

## Not a vulnerability report

The reviewed app's current sudo paths are correct; this is about removing the pressure to use sudo for a whole class of flows, so the fail-open footgun stops being the only option.

[...]

PR #1011, opened 2026-08-22
Throw when create access control returns a non-boolean

### LINKED ISSUE #1009: access: a filter returned from create access is silently treated as allow
> *This was generated by AI during triage.* Surfaced while verifying #966; filed separately so the fix isn't blocked on that issue's larger API question.

## Summary

`OperationAccess.create` is typed as `AccessControl`, which returns `boolean | PrismaFilter`. On the create path the engine tests the result only for `=== false`. So a `create` rule that returns a filter — the natural thing to write, and exactly what an author reaching for row-scoped create ownership will write — is **silently treated as a full allow**.

```ts
// Looks like it scopes the create. Does nothing. Allows everything.
create: ({ session }) => ({ ownerId: { equals: session.userId } })
```

Both create paths have it: the write pipeline's create step, and the nested-create path used by nested writes. Both call `checkAccess` with **no `item` and no input data**, then branch only on `false`.

## Why it happens

`AccessControl`'s argument shape is `{ session, item?, context }`, and `item` is documented as "Present for update/delete operations". On create there is no row and no input data in scope, so there is nothing for a filter to be applied to. Update and delete handle a returned filter properly — they re-check the target via `findFirst(mergeFilters(where, filter))`. Create has no equivalent, and rather than rejecting the filter it ignores it.

The type is what makes this reachable: it advertises a return value the create path cannot honour, and then accepts it.

## Why it matters

This is the same overloaded-signal defect ADR-0022 fixed on the read path — a value meaning "cannot scope this" being treated as "nothing to scope, pass through" — still live on the create path. ADR-0022 settled that an engine which cannot compute a scope must **deny**, never pass through.

It also lands precisely where users are most likely to hit it. #966 documents applications dropping to `sudo()` for create-time ownership because they cannot express it in access control; an author trying to avoid that by returning a filter from `create` gets a rule that reads as restrictive and enforces nothing.

## Verification performed

- Read both create call sites; confirmed each calls `checkAccess` without `item` or input data and branches only on `accessResult === false`.
- Confirmed `AccessControl` is `(args) => boolean | PrismaFilter<T> | Promise<...>` and that `OperationAccess.create` uses it unchanged.
- Confirmed the update/delete path does re-check a returned filter, so this is specific to create rather than general.
- No test covers a filter returned from `create` access.

## Related

- #966 — the pre-account/anonymous flows issue this surfaced under; that issue's broader API question is deliberately not bundled here
- ADR-0022 — access control fails closed when it cannot scope; this is the same defect shape on a path it didn't cover
- ADR-0030 — field-level access fails closed on a non-boolean result; the field-level analogue of this rule

borisno2:
## Summary

- `OperationAccess.create` is typed as `AccessControl` (`boolean | PrismaFilter`), but both create paths — the write pipeline's top-level create step and the nested-create path (which also backs `connectOrCreate`'s create branch) — only ever tested the result against `=== false`. A `create` rule returning a filter, the natural thing to write for row-scoped create ownership, silently fell through as a full allow.
- Added `checkCreateAccess`, a single shared evaluator in `access/engine.ts` that both create paths now call, so they cannot drift again. It throws a new `InvalidCreateAccessResultError` for any result that isn't a strict boolean — naming the list and explaining that create has no row to scope a filter against, and pointing at the alternative (a `resolveInput`/`validate` hook).
- `create: () => false` still denies via Silent failure exactly as before — only the non-boolean case is a new throw.
- This applies the decision already on record in ADR-0030 (field-level access fails closed on a non-boolean result) and ADR-0022 (an engine that cannot compute a scope must deny, never pass through) to this path. No new ADR needed.
- Updated `OperationAccess.create`'s TSDoc and the `config-api.md` reference doc to state that `create` accepts a boolean only, and why.
- Deliberately out of scope (per the issue): narrowing `OperationAccess['create']`'s type itself, changing `AccessControl`'s shared signature, and giving `create` access the input data (that's #966).

## Test plan

- [x] Unit tests for `checkCreateAccess` (filter/undefined/string throw; `true`/`false`/absent behave as before), naming the list in the error
- [x] Write-pipeline test: top-level `create` access returning a filter throws `InvalidCreateAccessResultError`, before DB and before `beforeOperation`
- [x] Nested-write integration test: nested `create` access returning a filter throws the same error (covers the nested path shared with `connectOrCreate`'s create branch)
- [x] `pnpm build`, `pnpm test` (1172 tests), `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all pass
- [x] Changeset added (minor, `@opensaas/stack-core`)

Closes #1009

---
_Generated by [Claude Code](https://claude.ai/code/session_011Li31AoECc3LcX7CusKTsV)_

### DIFF
--- a/.changeset/loud-otters-jump.md
+++ b/.changeset/loud-otters-jump.md
@@ -1,0 +1,29 @@
+---
+'@opensaas/stack-core': minor
+---
+
+`OperationAccess.create` now throws `InvalidCreateAccessResultError` when the rule returns anything other than `true`/`false` — most notably a Prisma filter, which previously fell through the `create` access check unrecognised and was silently treated as a full allow (both the top-level write pipeline and nested-create paths were affected).
+
+Create has no existing row to scope a filter against, so a filter can no longer be honoured here:
+
+```typescript
+// Before: type-checked, read as row-scoped, actually allowed everyone
+create: ({ session }) => ({ ownerId: { equals: session.userId } })
+
+// Now throws InvalidCreateAccessResultError. Scope ownership in a hook instead:
+hooks: {
+  resolveInput: async ({ resolvedData, context, operation }) => {
+    if (operation === 'create') {
+      return { ...resolvedData, ownerId: context.session?.userId }
+    }
+    return resolvedData
+  },
+},
+access: {
+  operation: {
+    create: ({ session }) => !!session, // boolean only
+  },
+},
+```
+
+`create: () => false` still denies via Silent failure as before; only a non-boolean result now throws.
--- a/docs/content/reference/config-api.md
+++ b/docs/content/reference/config-api.md
@@ -441,11 +441,32 @@ access: {
 
 #### Properties
 
-Each operation accepts an `AccessControl` function that returns:
+`query`, `update`, and `delete` accept an `AccessControl` function that returns:
 
 - `true` - Allow access
 - `false` - Deny access
 - `PrismaFilter` - Prisma where clause to filter accessible records
+
+**`create` accepts a `boolean` result only** — `true` or `false`. It shares
+`AccessControl`'s type (so a filter still type-checks), but there is no
+existing row for `create` to scope with a filter, and — unlike `update`/
+`delete`, which re-check a returned filter against the target row — no
+equivalent re-check is possible for a row that doesn't exist yet. A `create`
+rule that returns a filter (or any other non-boolean) throws
+`InvalidCreateAccessResultError` rather than being silently treated as an
+allow. To scope who may create a row by ownership, evaluate the condition in
+a `resolveInput` or `validate` hook, where the input data is in scope:
+
+```typescript
+hooks: {
+  resolveInput: async ({ resolvedData, context, operation }) => {
+    if (operation === 'create') {
+      return { ...resolvedData, ownerId: context.session?.userId }
+    }
+    return resolvedData
+  },
+},
+```
 
 **Type:** `AccessControl<T>`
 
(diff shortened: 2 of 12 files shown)


borisno2:
## Code review (medium effort)

Single-pass review of the diff against `main`.

**Checked:**
- Traced all `checkAccess`/`checkCreateAccess` call sites across the codebase — confirmed only the two `create` gates (write pipeline's top-level create, and the nested-create path shared by `connectOrCreate`'s create branch) needed the new evaluator, and both now route through the same `checkCreateAccess` in `engine.ts`.
- Verified the MCP `create` tool goes through `context.db[dbKey].create()`, so it inherits the fix automatically — no bypass path.
- Confirmed the throw happens in the pre-transaction access gate, before any hook or DB call fires, matching the documented behavior and the tests for it.
- Confirmed the `describeFieldAccessResult` → `describeAccessResult` rename left no other callers on the old name.
- Ran `pnpm build` (tsc), `pnpm test` (1172 tests / 57 files, all pass), and `pnpm lint` (clean, only pre-existing unrelated warnings) on the full repo.
- Reviewed test coverage: filter/`undefined`/string all throw `InvalidCreateAccessResultError`; `true`/`false`/absent access are unaffected; both create paths (top-level and nested) are covered; assertions confirm the DB call and `beforeOperation`/`resolveInput` never fire on the throwing path.
- Checked new comments/TSDoc against CLAUDE.md's comments policy — the `OperationAccess.create` docblock is a public-API export (the documented exception) and inline comments carry external rationale (ADR references) rather than restating code.

**Result:** no correctness, reuse/simplification, efficiency, or altitude issues found.


---

PR #1016, opened 2026-08-22
ui: identify password columns by field type, add password Cell

### LINKED ISSUE #982: ui: password columns are curated by field name, no password Cell is registered, and getFieldDisplayValue is dead code
> *This was generated by AI during triage.* Surfaced while verifying #965; filed separately rather than widening that issue.

## Summary

Two related loose ends around `password()` fields, neither of which is a live credential leak once #965 lands, but both of which shape what users copy from the examples.

## 1. No example models a field-level `read` gate on a password field

`examples/blog`, `examples/composable-dashboard`, `examples/custom-field`, and `examples/starter` each declare `password: password({ validation: { isRequired: true } })` with no field-level `access.read` rule. Field-level access allows when no rule is configured for the operation, so the value is readable through `context.db` in all four.

Once #965 redacts `HashedPassword.toJSON()`, this stops being a serialization leak — the hash cannot reach output through `JSON.stringify` regardless. The residual concern is that `examples/starter` is the template `create-opensaas-app` scaffolds from, so whatever it does is what every new project starts with, and none of the four demonstrates that gating the field is an option the user has.

Worth deciding explicitly rather than by default: **is a field-level `read` deny the shape we want examples to model, or is serialization-layer redaction considered sufficient?** #965's triage rejected making deny-read the *framework default* — the field's documented authentication pattern reads the value through `context.db` and calls `.compare()` on it, so a global default would break the field's advertised use. That reasoning does not settle whether an example should show the gate for a list that never needs `compare()`.

## 2. The admin UI curates password columns by field *name*, not field *type*

Two places drop password columns from default views using a hardcoded name list:

- the standalone list table's default column derivation, filtering `['password', 'createdAt', 'updatedAt']`
- the item-view derivation's `DEFAULT_EXCLUDED_COLUMNS`, the same three names

Meanwhile the shared value formatter masks by **type** (`case 'password': return '••••••••'`).

So a field declared as `secret: password()` — a `password()` field not named `password` — is not excluded from default columns, while a field merely *named* `password` is excluded whatever its type. The curation and the masking disagree about what identifies a password field, and the exclusion is the one using the weaker signal.

**Limit of verification:** I confirmed the name-based filtering and the type-based masking by reading both, but did not verify end-to-end whether the cell renderer routes a differently-named `password()` field through the type-based formatter. If it does, such a field renders as a masked column rather than being hidden — untidy but not a leak. If it does not, it renders the raw hash into the admin table, which would be a live UI leak and would make this a `bug` rather than an `enhancement`. **That check should be the first thing done when triaging this**, since it decides the category.

This may warrant splitting into its own issue depending on how that check lands.

## Suggested direction

- Make the UI's password-column curation type-based, matching the formatter, so it holds for any `password()` field regardless of name.
- Decide whether examples should model a field-level `read` gate, and if so apply it consistently across the four — prioritising `examples/starter`, since it is the scaffolding template.

## Related

- #965 — `HashedPassword.toJSON()` leaking the hash (`ready-for-agent`). Its fix covers the serialization path for all four examples; this issue is what remains after it.
- #981 — derived auth lists' credential fields, a separate and sharper gap.

borisno2:
## Summary

- Default column derivation excludes password columns by field **type** rather than by matching the field name `"password"`, in all three places that curated default columns: `ListViewClient` (main list view), the standalone `ListTable`, and `deriveItemView`'s Relationship-table column derivation. A field declared `secret: password()` is now excluded from default columns; a field merely named `password` that isn't `password()`-typed (e.g. `password: text()`) is no longer excluded.
- Registered a `password` Cell in the cell registry (`cellComponentRegistry`) that renders a fixed `••••••••` mask, ignoring the value entirely. A password-typed column shown deliberately via an explicit `columns` prop now renders the mask instead of falling through to the default text Cell (previously `[object Object]`, saved only incidentally by `HashedPassword.toJSON()` redaction).
- Removed the dead `getFieldDisplayValue` export from `@opensaas/stack-ui` — it had zero callers in the actual rendering path (Cells render each field type directly) and disagreed with what was actually rendered.
- Added a changeset (minor, per pre-1.0 convention) documenting the removed export as a breaking change even though it ships as minor.

## Test plan

- [x] Added/updated unit tests covering: password-typed columns excluded by default regardless of name (`ListViewClient`, `ListTable`, `deriveItemView`), a field merely *named* `password` is **not** excluded when it isn't password-typed, and the new `PasswordCell` (direct render + via `CellRenderer` type resolution) renders a fixed mask ignoring the raw value.
- [x] `pnpm test` in `packages/ui` — 594/594 passing
- [x] `pnpm build` in `packages/ui` (typecheck via `tsc`) — passes
- [x] `pnpm manypkg fix` / `pnpm format` — clean

Closes #982

---
_Generated by [Claude Code](https://claude.ai/code/session_014dMr934DP8TbqiLHrizQ2m)_

### DIFF
--- a/.changeset/silent-jars-refuse.md
+++ b/.changeset/silent-jars-refuse.md
@@ -1,0 +1,9 @@
+---
+'@opensaas/stack-ui': minor
+---
+
+Password columns are now identified by field type, not field name, across the list view, standalone `ListTable`, and item-view Relationship tables. A field declared `secret: password()` is now excluded from default columns even though it isn't named `password`; a field merely named `password` (e.g. `password: text()`) is no longer excluded unless it is actually a `password()` field.
+
+A `password` Cell is now registered in the cell registry, so a password-typed column shown via an explicit `columns` prop renders a fixed `••••••••` mask instead of the raw value.
+
+BREAKING (shipped as minor — pre-1.0 packages ship breaking changes as minor): the unused `getFieldDisplayValue` export has been removed from `@opensaas/stack-ui`. It had no callers in the rendering path — Cells render each field type directly — so nothing in this package depended on it; a consumer importing it directly should port to a project-local formatter.
--- a/packages/ui/src/components/ListViewClient.tsx
+++ b/packages/ui/src/components/ListViewClient.tsx
@@ -185,7 +185,9 @@ export function ListViewClient({
 
   const displayColumns =
     columns ||
-    Object.keys(fieldTypes).filter((key) => !['password', 'createdAt', 'updatedAt'].includes(key))
+    Object.keys(fieldTypes).filter(
+      (key) => fieldTypes[key] !== 'password' && !['createdAt', 'updatedAt'].includes(key),
+    )
 
   // Items are already sorted by the server via orderBy; no in-memory sort needed.
 
--- a/packages/ui/src/components/cells/PasswordCell.tsx
+++ b/packages/ui/src/components/cells/PasswordCell.tsx
@@ -1,0 +1,13 @@
+'use client'
+
+import * as React from 'react'
+import type { CellComponentProps } from './registry.js'
+
+/**
+ * Renders a fixed mask for a `password()` column regardless of `value` — a
+ * password column must never render the underlying value, so this Cell does
+ * not read it at all.
+ */
+export function PasswordCell(_props: CellComponentProps) {
+  return <span data-slot="cell-password">••••••••</span>
+}
--- a/packages/ui/src/components/cells/index.ts
+++ b/packages/ui/src/components/cells/index.ts
@@ -5,6 +5,7 @@ export { CheckboxCell } from './CheckboxCell.js'
 export { SelectCell } from './SelectCell.js'
 export { TimestampCell } from './TimestampCell.js'
 export { RelationshipCell } from './RelationshipCell.js'
+export { PasswordCell } from './PasswordCell.js'
 export { AvatarLabelCell } from './AvatarLabelCell.js'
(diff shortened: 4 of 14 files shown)


borisno2:
## Code review

Ran a full review pass (line-by-line diff, removed-behavior audit, cross-file tracing, reuse/simplification/efficiency/altitude, CLAUDE.md conventions).

**Findings: none.**

- All three default-column call sites (`ListViewClient`, standalone `ListTable`, `deriveItemView`'s Relationship-table columns) consistently switched from name-based to type-based password exclusion.
- Confirmed `getFieldDisplayValue` has zero remaining callers repo-wide before removing it.
- Confirmed no other stray name-based `'password'` filters exist outside the three sites fixed here.
- The new `PasswordCell` respects the existing `ui.cell` → `ui.fieldType` → `field.type` override priority in `CellRenderer`.
- Tests cover both directions of the type-vs-name distinction (a `secret: password()` field is excluded; a `password: text()` field is not) across all three sites, plus the Cell rendering a fixed mask independent of the raw value.

`pnpm test` (594/594), `pnpm build`, `pnpm lint`, `pnpm manypkg fix`, and `pnpm format` all pass clean.

---

borisno2:
The `Vercel` status check is failing with "Deployment rate limited — retry in 24 hours" (`api-deployments-free-per-day` quota on the `stack-docs` project) — an external free-tier limit unrelated to this diff, not something a code change here can fix. The real CI checks (`test`, `e2e`) are still running; will follow up once they complete.

---

PR #1017, opened 2026-08-22
auth: let apps declare model-level indexes on derived auth lists

### LINKED ISSUE #921: auth: derived auth lists can't adopt a live constraint name, so `User.email` / `Session.token` can diff dirty

### LINKED ISSUE #985: auth: no seam for an app to declare its own indexes on a derived auth list
> *This was generated by AI during triage.* Split out of #937 so that issue can stay scoped to emitting the three indexes better-auth declares.

## Summary

An application cannot add an index to a plugin-derived auth list. The derived lists (`User`, `Session`, `Account`, `Verification`, `RateLimit`) declare their fields and their `db` block inside the auth plugin, and there is no passthrough for an application to contribute a `db.indexes` entry to them.

This is a different problem from #937. That issue is about indexes **better-auth itself declares** that the stack fails to emit — a parity bug with a fixed, knowable answer. This issue is about indexes **the application needs** that neither better-auth nor the stack can know about.

## The reported case

From a downstream app review ([#937 comment](https://github.com/OpenSaasAU/stack/issues/937#issuecomment-5340155165)): a per-email OTP resend-cooldown check does

```ts
findFirst({ where: { identifier }, orderBy: { createdAt: 'desc' } })
```

on the verification table, which wants a composite `(identifier, createdAt)` index. The app had added it by migration as `verification_identifier_createdAt_idx`. After adopting the stack's derived `Verification` list, `migrate diff` produced a `DROP INDEX` for it — the same failure mode as #937, but for an index the app added rather than one better-auth ships, and with no way to re-declare it.

Note that `db.indexes` already exists and supports exactly this shape — named, composite, with sort direction. The gap is purely that a derived list cannot receive one from the application side.

## better-auth 1.7.x shipped the API this should mirror

Worth knowing before designing this. `@better-auth/core` gained a **table-level `indexes` option** in the 1.7 line — it does not exist in 1.6.25 (zero occurrences in that version's table definitions). In 1.7.1, each model accepts an `indexes` array and merges it with the defaults:

```js
account: {
  indexes: mergeTableIndexes([{ fields: ["issuer", "accountId"], unique: true }], account?.indexes),
  …
}
```

Entries are `{ name?, fields, unique? }`, deduplicated by that triple. `user`, `session`, and `verification` take the same `?.indexes` passthrough, and plugin schemas merge their own in.

Two consequences:

1. **The shape is settled upstream.** Whatever seam the stack exposes should mirror this rather than invent a parallel vocabulary, so an app migrating from a better-auth-native config finds the same concept in the same place. The stack's own `db.indexes` entry shape (`{ fields, unique?, name?, sort }`) is already close.
2. **It bounds the "app-supplied" category.** With the upstream option available, some indexes an app needs can be declared through better-auth's config instead. The stack still needs its own seam — the derived lists are the stack's schema surface, and the generator reads the stack's config, not better-auth's — but the two should not disagree about what the index set is.

## Design questions for triage

- **Where does the passthrough live?** The natural home is alongside the existing per-model access passthrough in the auth plugin config (`authPlugin({ access: { session: … } })`), keyed the same way by better-auth model name. That is consistent with ADR-0013's principle — the application authors the config, the plugin is the conduit.
- **Merge or replace?** If #937 lands and the derived lists emit better-auth's three indexes themselves, an app-supplied array must merge with those rather than replace them. better-auth's own `mergeTableIndexes` dedupes by `(name, fields, unique)`; the stack would want equivalent behaviour, plus a sensible error when an app-supplied entry collides with a field-level `isIndexed` on the same column (core already errors on that case for app-declared lists).
- **How wide is the seam?** This is the second concrete instance of "an application cannot configure a derived auth list" — see #981, where the same structural gap blocks field-level access on credential columns. It may be worth deciding the general shape of app-authored config passthrough for derived lists once, rather than adding one keyhole per need.

## Related

- #937 — emit the three indexes better-auth declares; this issue was split out of its second comment
- #981 — the same missing-seam structure, for field-level access on derived credential fields
- #921 — constraint-name adoption on derived lists, another instance of the derived lists not being configurable enough to adopt a live database
- #935 — the FK column these indexes sit on
- ADR-0013 — the application authors config for plugin-injected lists; the plugin is the conduit

borisno2:
## Summary

- `authPlugin()`'s per-model config (`user`/`session`/`account`/`verification`/`rateLimit`) now accepts an `indexes` option, in the same shape as a list's own `db.indexes` (core's `ListIndex`). Entries are threaded through to the derived list's `db.indexes` at generate time.
- When an app-supplied entry covers a column the stack already derives an index for (e.g. `User.email`'s derived `@unique`), the derived index is **suppressed for that column only** and the app's entry is emitted instead — the application's declaration wins over a derived default (ADR-0035). This makes two previously-unreachable cases expressible:
  - Adopting a live constraint's real name (e.g. `{ fields: ['email'], unique: true, name: 'user_email_key' }`) — closes #921.
  - Extending a derived column into a composite index (e.g. a per-`identifier` resend-cooldown index on `Verification`, spanning `identifier` + `createdAt`).
- Fixes a related generator gap: `db.indexes` can now reference `createdAt`/`updatedAt` even when the list has no explicitly declared field for them and relies on `db.timestamps` for the auto-injected columns — required for the `Verification` composite-index case above, and generically useful for any list using `db.timestamps: true`.
- Unknown-field errors reuse core's existing generator error path unchanged, and correctly name the list under its remapped `modelName` (e.g. `AuthVerification`).
- Plugin-derived tables (from `betterAuthPlugins`) and adopting better-auth's own 1.7 table-level `indexes` option are explicitly out of scope (tracked separately as #986).

## Test plan

- [x] New tests in `packages/auth/tests/generated-fk-shape.test.ts` cover: composite index + suppression, named-unique adoption on `User.email`, per-column suppression (other derived indexes untouched), field-key resolution through a model's column map, the #921 round-trip (`User.email` + `Session.token`), unknown-field errors (plain and under `modelName` remapping), and unchanged output when no `indexes` are configured.
- [x] New tests in `packages/cli/src/generator/prisma.test.ts` cover the generic `createdAt`/`updatedAt` auto-timestamp resolution in `db.indexes`, independent of auth.
- [x] Updated two pre-existing `packages/auth/tests/config.test.ts` assertions for the new `indexes: []` field on normalized model config.
- [x] `pnpm lint`, `pnpm manypkg fix`, `pnpm format` all pass.
- [x] Full `core`/`cli`/`auth` package test suites pass (1175 + 381 + 279 tests).
- [x] Changeset added (`@opensaas/stack-auth` minor, `@opensaas/stack-cli` minor, `@opensaas/stack-core` patch for the TSDoc clarification).
- [x] Reference docs (`docs/content/reference/auth.md`) and how-to guide (`docs/content/how-to/authentication.md`) updated, cross-linked to core's `db.indexes` reference.

Closes #985

https://claude.ai/code/session_01PLqppkKWzoMZvy3aCDK6DY

---
_Generated by [Claude Code](https://claude.ai/code/session_01PLqppkKWzoMZvy3aCDK6DY)_

### DIFF
--- a/docs/content/how-to/authentication.md
+++ b/docs/content/how-to/authentication.md
@@ -1125,6 +1125,46 @@ existed: the table name follows `modelName` whenever it differs from the
 better-auth default, so a renamed-table install (the `adoptBetterAuthTables()`
 default from the previous section) keeps working without any changes.
 
+### Adopting a live constraint name or adding your own index (`indexes`)
+
+Each per-model block (`user`/`session`/`account`/`verification`/`rateLimit`)
+also accepts `indexes`, in the same shape as a list's own `db.indexes` (see
+the [`db.indexes` reference](/docs/reference/config-api#dbindexes)) — entries
+name the model's own field keys, not raw column names.
+
+The stack already derives some indexes from better-auth's own table
+definitions — `User.email` and `Session.token` are unique, for example. If
+your live database's constraint has a different name than the one Prisma
+would derive, adopting it under its real name is a generate-clean diff away:
+
+```typescript
+authPlugin({
+  user: { indexes: [{ fields: ['email'], unique: true, name: 'user_email_key' }] },
+  session: { indexes: [{ fields: ['token'], unique: true, name: 'session_token_key' }] },
+})
+```
+
+An entry covering a column the stack already derives an index for **replaces**
+that derived index rather than erroring — your declaration wins. This is the
+opposite of `db.indexes` on a list you declare yourself, where a collision
+with a field's own `isIndexed` is a config-time error: here, one of the two
+declarations is derived, so there's nothing of yours to remove.
+
+The same seam lets you extend a derived column into a composite index — e.g.
+a per-identifier resend-cooldown check on the verification table:
+
+```typescript
+authPlugin({
+  verification: {
+    indexes: [{ fields: ['identifier', { field: 'createdAt', sort: 'desc' }] }],
+  },
+})
+```
+
+This suppresses the derived single-column index on `identifier` in favor of
+the composite, which serves the same lookups. Suppression is per-column: every
+other index the stack derives for that model is unaffected.
+
 ### Linking your app User to the Auth identity
 
 Because the Auth identity (`AuthUser`) and your domain `User` are separate
--- a/.changeset/quiet-otters-declare.md
+++ b/.changeset/quiet-otters-declare.md
@@ -1,0 +1,25 @@
+---
+'@opensaas/stack-auth': minor
+'@opensaas/stack-cli': minor
+'@opensaas/stack-core': patch
+---
+
+Let an application declare model-level indexes (`db.indexes`) on the derived auth lists (`User`/`Session`/`Account`/`Verification`/`RateLimit`).
+
+Each per-model block in `authPlugin()` now accepts `indexes`, in the same shape as a list's own `db.indexes`:
+
+```typescript
+authPlugin({
+  // Adopt a live constraint's real name instead of Prisma's derived one.
+  user: { indexes: [{ fields: ['email'], unique: true, name: 'user_email_key' }] },
+  session: { indexes: [{ fields: ['token'], unique: true, name: 'session_token_key' }] },
+  // Extend a derived column into a composite index.
+  verification: {
+    indexes: [{ fields: ['identifier', { field: 'createdAt', sort: 'desc' }] }],
+  },
+})
+```
+
+An entry covering a column the stack already derives an index for (e.g. `User.email`) suppresses that derived index for that column and emits only the app's entry, rather than erroring — the application's declaration wins (ADR-0035). Suppression is per-column: every other derived index on the model is unaffected.
+
+This also fixes a related generator gap: a list's `db.indexes` can now reference `createdAt`/`updatedAt` even when the list has no explicit field for them and relies on `db.timestamps` for the auto-injected columns (previously only a list with an explicitly declared `createdAt`/`updatedAt` field could be indexed on it).
--- a/docs/content/reference/auth.md
+++ b/docs/content/reference/auth.md
@@ -360,7 +360,7 @@ authPlugin({
 
 Derivation keys off `storage` alone, not `enabled` — `{ enabled: false, storage: 'database' }` still produces the `RateLimit` list, since better-auth still expects the table regardless of whether the limiter is currently active (`enabled` is routinely environment-driven, and tying the generated schema to it would make dev and prod schemas differ).
 
-`rateLimit` also carries the same adoption knobs as the other four models — `modelName`, `fields`, `tableName`, `schema` — so an app with an existing database-backed limiter table can adopt it rather than being forced into a new one:
+`rateLimit` also carries the same adoption knobs as the other four models — `modelName`, `fields`, `tableName`, `schema`, `indexes` — so an app with an existing database-backed limiter table can adopt it rather than being forced into a new one:
 
 ```typescript
 authPlugin({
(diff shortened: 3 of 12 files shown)


docs/content/how-to/authentication.md
```diff
@@ -1125,6 +1125,46 @@ existed: the table name follows `modelName` whenever it differs from the
...
+a per-identifier resend-cooldown check on the verification table:
+
+```typescript
+authPlugin({
+  verification: {
+    indexes: [{ fields: ['identifier', { field: 'createdAt', sort: 'desc' }] }],
+  },
+})
+```
+
```
borisno2, line 1163: This example relies on `Verification`'s auto-timestamps (`db.timestamps`) to resolve `createdAt` for the index even though no field declares it — the new capability this PR adds to `resolveIndexFieldColumn` in `packages/cli/src/generator/prisma.ts` (and covers with tests in `prisma.test.ts`: "createdAt/updatedAt via auto-timestamps, with no declared field (#985)").

`docs/content/reference/config-api.md` documents `db.indexes` in general (not just for auth lists) but wasn't updated to match: its `AuthVerification` example (around line 401) still declares `createdAt: timestamp()` as an explicit field, and its error list (line 417: "An entry naming a field the list doesn't have") still reads as unconditional — no mention of the new `db.timestamps`-backed exception. A reader of that page alone would believe declaring the field is still required, or that indexing an auto-timestamp column without declaring it still errors.

Failure scenario: a developer following only the general `db.indexes` reference (not this auth how-to) always adds a redundant explicit `createdAt`/`updatedAt` field to index it, or avoids the feature entirely believing it still throws.

---

borisno2 (commented):
Single-pass review at high effort (Agent tool unavailable in this session, so no multi-agent fan-out / subagent verify pass ran — this is one careful pass through the diff, cross-file callers, tests, ADR-0035, and CLAUDE.md conventions).

Verified: types (`packages/core/src/config/types.ts`, `packages/auth/src/config/types.ts`), the `deriveAuthLists` suppression logic (`packages/auth/src/config/derive-auth-lists.ts`), the generator's `createdAt`/`updatedAt` auto-timestamp resolution (`packages/cli/src/generator/prisma.ts`), and all associated tests all typecheck, lint clean, and pass (`pnpm vitest run` in both `packages/auth` and `packages/cli`, plus `tsc --noEmit` in both packages). The implementation matches ADR-0035 exactly, and the suppress-per-column / plugin-tables-out-of-scope claims in the new docblocks and code comments all check out against the code.

No correctness bugs found in the code itself. One finding: the general `db.indexes` reference doc (`docs/content/reference/config-api.md`) wasn't updated to reflect the new `createdAt`/`updatedAt`-via-auto-timestamps behavior this PR ships, so it now reads as stale/incomplete relative to the shipped generator behavior — flagged inline on `docs/content/how-to/authentication.md` since `config-api.md` itself has no diff lines to anchor a comment to.

---

